PR #4, opened 2022-05-24
Compilation

leviska:
Basic compilation and compiler caching


verification/src/http_server/handlers/verification/solidity/types.rs
```diff
@@ -17,20 +22,7 @@ pub struct VerificationResponse {
...
-    SpuriousDragon,
-    Byzantium,
-    Constantinople,
-    Petersburg,
-    Istanbul,
-    Berlin,
-    London,
-    Default,
-}
+type EvmVersion = ethers_solc::EvmVersion;
```
sevenzing, line 25: `ethers_solc::EvmVersion` contains items starting with capital letters. we have to rename all of them to camel case. before it was done by using `#[serde(rename_all = "camelCase")]` but I don't know how to do it without access to struct

verification/src/solidity/compiler_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    async fn fetch(cache: &CompilerCache, ver: &Version) {
+        let now = std::time::Instant::now();
+        cache.get(&ver).await.unwrap();
+        println!("{} {:?}", ver, now.elapsed());
```
sevenzing, line 57: what is the reason for println here? by default tests will not print stdout, you should add `--nocapture` to workflow command. If this print for local debug only, then it should be in `log::info`

verification/src/http_server/handlers/verification/solidity/types.rs
leviska: yes, I know, I'll fix it

verification/src/solidity/compiler_cache.rs
```diff
@@ -0,0 +1,85 @@
+use ethers_solc::Solc;
+use semver::Version;
+use std::{collections::HashMap, sync::Arc};
+
+#[derive(Default)]
+pub struct CompilerCache {
```
rimrakhimov, line 6: May be use `SolcManager` as the name of the structure with `fetch` public method which checks whether the path is in the cache or download it, otherwise? 

The problem with `CompilerCache`, in my opinion, is that when you get something from cache in most cases you expect it to return quickly and definitely not to run a long request downloading compilers from somewhere

rimrakhimov, line 6: Also, some explicit initialization would be useful to allow reuse locally downloaded compilers in case the service will be rerun

verification/src/solidity/compiler_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+    async fn fetch(&self, ver: &Version) -> anyhow::Result<Solc> {
+        let lock = {
+            let mut cache = self.cache.lock().unwrap();
+            Arc::clone(cache.entry(ver.clone()).or_default())
+        };
+        let mut entry = lock.write().await;
+        match entry.as_ref() {
+            Some(solc) => Ok(solc.clone()),
+            None => {
+                log::info!("installing solc version {}", ver);
```
rimrakhimov, line 34:
```suggestion
                log::info!(target: "compiler_cache", "installing solc version {}", ver);
```

Consider using `target` key describing the module which generated a log

verification/src/solidity/compiler_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    async fn fetch(cache: &CompilerCache, ver: &Version) {
+        let now = std::time::Instant::now();
+        cache.get(&ver).await.unwrap();
+        log::debug!("{} {:?}", ver, now.elapsed());
+    }
```
rimrakhimov, line 58: What is the reason to use logs in tests?

verification/src/solidity/compiler_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 3)]
+    async fn it_works() {
```
rimrakhimov, line 61: I believe the name should be more meaningful

verification/src/solidity/compiler_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+            let mut tasks = vec![];
+            for j in 0..i + 1 {
+                let c = Arc::clone(&cache);
+                let v = versions[j].clone();
+                tasks.push(tokio::spawn(async move {
+                    fetch(&c, &v).await;
+                }));
+            }
+            let r: Result<Vec<_>, _> = futures::future::join_all(tasks).await.into_iter().collect();
+            r.unwrap();
```
rimrakhimov, line 82: It is not clear what does that test check?

verification/Cargo.toml
```diff
@@ -9,7 +9,11 @@ repository = "https://github.com/blockscout/blockscout-rs/verification"
...
 serde = { version = "1", features = ["derive"] }
-tokio = { version = "0.2", features = ["macros"]}
+tokio = { version = "1.18", features = ["full"]}
+ethers-solc = { git = "https://github.com/gakonst/ethers-rs", version = "0.3", features = ["svm-solc"] }
 serde_json = "1"
 log = "0.4"
 env_logger = "0.9"
+semver = "1.0"
+anyhow = "1.0"
+futures = "0.3"
```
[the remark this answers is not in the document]
leviska: we might need to use them later not in tests, this package has a lot of useful functions

verification/src/solidity/compiler_cache.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    async fn fetch(cache: &CompilerCache, ver: &Version) {
+        let now = std::time::Instant::now();
+        cache.get(&ver).await.unwrap();
+        println!("{} {:?}", ver, now.elapsed());
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 3)]
+    async fn it_works() {
+        env_logger::init();
```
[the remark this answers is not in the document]
leviska: why not? if something crashes in tests, it is very useful to have logs to debug the problem

verification/Cargo.toml
```diff
@@ -9,7 +9,11 @@ repository = "https://github.com/blockscout/blockscout-rs/verification"
 actix-web = "4"
 clap = { version = "3", features = ["derive"] }
 serde = { version = "1", features = ["derive"] }
-tokio = { version = "0.2", features = ["macros"]}
+tokio = { version = "1.18", features = ["full"]}
```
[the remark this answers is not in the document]
leviska: 1. we need macros, I'll replace it

verification/src/http_server/handlers/verification/solidity/types.rs
```diff
@@ -13,23 +18,41 @@ pub struct VerificationRequest<T> {
...
+    fn from(v: ethers_solc::EvmVersion) -> Self {
+        Self(v)
+    }
+}
+
+impl From<EvmVersion> for ethers_solc::EvmVersion {
+    fn from(v: EvmVersion) -> Self {
+        v.0
+    }
 }
```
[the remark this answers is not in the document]
leviska: I've implemented from traits for future use, like serialization

If we implement our own enum, what happens when the new evm version comes? Right now we only need to update the library, our code will be intact, but if we'll implement our own enum, we'll need to update it too

The most flexible variant is just to accept the string, and pass it to the compiler, because we basically don't care if it's valid (also some newer versions aren't supported on old compilers), but this isn't compatible with ethers_solc library as of now

verification/src/solidity/compiler_cache.rs
```diff
@@ -0,0 +1,85 @@
+use ethers_solc::Solc;
+use semver::Version;
+use std::{collections::HashMap, sync::Arc};
+
+#[derive(Default)]
+pub struct CompilerCache {
+    cache: std::sync::Mutex<HashMap<Version, Arc<tokio::sync::RwLock<Option<Solc>>>>>,
```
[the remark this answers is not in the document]
leviska: less boilerplate?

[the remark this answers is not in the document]
leviska: just the opposite 

> The exception to this idiom is if we’re bringing two items with the same name into scope with use statements

with most types it is clear where they come from, but here we have tokio::sync::Mutex and std::sync::Mutex, and similar with rwlock, and I think it is important to know which lock is used where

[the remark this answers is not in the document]
leviska: we might just wanna wait for rust 1.62, where std mutexes become much better https://twitter.com/m_ou_se/status/1526220580361490438

verification/src/solidity/compiler_cache.rs
leviska: > is that when you get something from cache in most cases you expect it to return quickly

well that's just not true :) 
it is very common to cache static data for any use, including long term (for example caching popular videos in youtube, which will be watched for a long time)

`ThingManager` imo is not descriptive, how does this struct manages solc? Configures, downloads, stores, loads in the memory, compiles?

verification/src/http_server/handlers/verification/solidity/types.rs
```diff
@@ -13,23 +18,41 @@ pub struct VerificationRequest<T> {
...
 
 #[derive(Debug, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
```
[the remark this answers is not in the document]
leviska: my bad

verification/Cargo.toml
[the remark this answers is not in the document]
rimrakhimov: If we need them, I believe it will not be a problem to move it out of `[dev-dependencies]` later.

verification/src/download_cache.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    async fn try_get(&self, ver: &Version) -> Option<PathBuf> {
+        let entry = {
+            let cache = self.cache.lock();
+            cache.get(ver).cloned()
+        };
+        match entry {
+            Some(lock) => {
+                let solc = lock.read().await;
+                solc.as_ref().cloned()
```
sevenzing, line 32: since DownloadCache doesnt know anything about solc, I think it should not be `solc` variable

sevenzing (approved):

verification/src/download_cache.rs
```diff
@@ -0,0 +1,189 @@
+use async_trait::async_trait;
+use semver::Version;
+use std::{collections::HashMap, path::PathBuf, sync::Arc};
+
+#[async_trait]
+pub trait Fetcher {
+    type Error: std::error::Error;
```
rimrakhimov, line 7:
```suggestion
    type Error;
```
Why should it implement `std::error::Error` and cannot be more general?

rimrakhimov (approved):
Accidentally left the comments from yesterday pending, sorry. 

In general I believe we may merge the PR, however, it would be great to address last comments, at least the ones which do not require great changes (e.g. removing trait bounds from `Fetcher::Error` if possible, changing test function names)

PR #6, opened 2022-05-30
Update config and add sourcify

sevenzing:
+ Add configuration parsing from `verification/configuration.toml` file
+ Add initial sourcify handler, that sends request to sourcify api
+ Add integration test for sourcify handler, mainly as example of how can we test our application


verification/tests/endpoints/contracts/storage/metadata.json
```diff
@@ -0,0 +1,88 @@
+{
```
leviska, line 1: suggest new folder name: `storage` -> `test_data`?

verification/src/configuration.rs
```diff
@@ -0,0 +1,31 @@
+use config::{Config, File, FileFormat};
+use serde::Deserialize;
+use std::net::SocketAddr;
+
+#[derive(Deserialize, Clone)]
+pub struct Configuration {
```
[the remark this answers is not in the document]
rimrakhimov: @sevenzing You may use `config::Config` in the places there it is required, or import it with `use config::Config as LibConfig`

verification/src/http_server/handlers/verification/solidity/sourcify.rs
```diff
@@ -0,0 +1,70 @@
+use actix_web::{error, error::Error, web::Json};
```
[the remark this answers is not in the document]
rimrakhimov: Not quite sure that I get you. In general all those `Sourcify*` structures (except `SourcifyRequest`) are not exposed outside the module. So, we can name them as would be clearer for the `sourcify` mod. But I didn't get what structures do you mean

verification/tests/endpoints/contracts/storage/metadata.json
rimrakhimov: I suggest moving everything out of `endpoints` directory. So that we will have `verification/tests/contracts/contract_name/..` with all contracts and `verification/tests/sourcify.rs` with actual test file.

Also, in integration tests we do not need `mod.rs` to combine them, as each `.rs` file is a separate crate by itself which runs all tests.

verification/tests/endpoints/sourcify.rs
```diff
@@ -0,0 +1,45 @@
+#[cfg(test)]
+mod tests {
```
rimrakhimov, line 2: It is not required to put tests under a separate `#[cfg(test)]` module, as for integration tests `#[cfg(test)]` is implicitly implied.

verification/tests/integration.rs
```diff
@@ -0,0 +1 @@
+mod endpoints;
```
rimrakhimov, line 1: That file is not required

verification/src/http_server/handlers/verification/sourcify/types.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub address: String,
+    pub chain: String,
+    pub files: HashMap<String, String>,
+}
+
+// Definition of sourcify.dev API response
+// https://docs.sourcify.dev/docs/api/server/v1/verify/
+#[derive(Deserialize)]
+#[serde(untagged)]
+pub enum SourcifyVerifyResponse {
```
rimrakhimov, line 17: We may may expose those types to `sourcify/mod.rs` without exposing them to others
```suggestion
pub(super) enum SourcifyVerifyResponse {
```

rimrakhimov (approved):

verification/src/http_server/handlers/verification/solidity/sourcify.rs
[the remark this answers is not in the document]
leviska: I'm talking about `SourcifyRequest`, `SourcifyVerifyResponse`, `sourcify_verification_request`

> So, we can name them as would be clearer for the sourcify mod

Isn't it will be clearer, if we'd have just `Request`, `VerifyResponse` and `verification_request` inside sourcify? This is a sourcify module and these are local to sourcify module types and functions, so we don't need to add the prefix to the names

This suggestion comes from [go](https://go.dev/doc/effective_go#package-names), where they suggest use package names as prefixes to the names:

> The importer of a package will use the name to refer to its contents, so exported names in the package can use that fact to avoid repetition. <...> For instance, the buffered reader type in the bufio package is called `Reader`, not `BufReader`, because users see it as `bufio.Reader`, which is a clear, concise name. Moreover, because imported entities are always addressed with their package name, `bufio.Reader` does not conflict with `io.Reader`. <...> Use the package structure to help you choose good names.

In Rust we don't necessary need to use package names, but we can, and I think in this example it can be easier, if we move prefix to the module name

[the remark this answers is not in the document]
sevenzing: agree with lev, think its better to remove those sourcify prefixes

verification/tests/sourcify.rs
```diff
@@ -0,0 +1,42 @@
...
+        // https://kovan.etherscan.io/address/0xfbe36e5cad207d5fdee40e6568bb276a351f6713
+        "address": "0xFBe36e5cAD207d5fDee40E6568bb276a351f6713",
+        "chain": "42",
+        "files": {
+            "source.sol": source,
+            "metadata.json": metadata,
+        }
+    });
+
+    let resp = TestRequest::get()
```
leviska, line 29: what's the difference between `TestRequest` and `reqwest`? 

Also, what if the tests are runned on the machine with no internet connection (it can be some CI/CD runner in the datacenter), or sourcify is down? 

Are we testing, that current sourcify api will return the correct response (i.e. it didn't change the api), or are we testing, that we're calling the sourcify api correctly? If it's the second, may be we want to mock the sourcify api then? And I don't think, that we're really need the first kind of tests, but if we are, I think it'll be better to do as a separate test

leviska: But if we decide to change the test, I suggest doing it in another PR

sevenzing: + we didn't start server on some port, TestRequest just calls service of app.

+ We are testing whole endpoint, checking that it works correctly. I decided not to write mock for soucify because I don't think we need it right now. When tests will fail due to internet connection, then we will fix it.

+ Yes, we can change the test in another PR

rimrakhimov: > And I don't think, that we're really need the first kind of tests, but if we are, I think it'll be better to do as a separate test

Why?

leviska: > Why?

I can trust sourcify not to change their api without backward compability, and I think we don't need these tests _at least right now_

leviska: > we didn't start server on some port, TestRequest just calls service of app.

Ah, thanks, that's convinient

> I decided not to write mock for soucify because I don't think we need it right now

Fair enough

leviska (approved):

PR #8, opened 2022-06-02
Solidity verifier initialization

rimrakhimov:
Continues #5 

This PR implements verifier initialization. Mostly initialization consists of two main procedures: parsing deployed bytecode and creation transaction input from strings into more structured formats. 

The parsing consists of the following steps:
1. Deployed bytecode is parsed from hex to bytes. Returns `InitializationError::InvalidDeployedBytecode` if the string is invalid hex.
2. Metadata hash is extracted from deployed bytecode and parsed trying to retrieve `solc` key value. If any error occurs at that step, `InitializationError::MetadataHashParseError` error is returned (that includes case if byte string was empty).
3. Creation transaction input is parsed from hex to bytes. Returns `InitializationError::InvalidCreationTxInput` if the string is invalid hex.
4. Encoded metadata hash obtained from deployed bytecode is searched in creation transaction input. If not found `InitializationError::MetadataHashMismatch` is returned. Otherwise, two parts are extracted from creation transaction input: bytecode located before metadata hash and data located after metadata hash.
5. Initialization succeeds if no errors occurred on previous steps.

Currently, the structure responsible for parsing of creation transaction input is named `BytecodeWithConstructorArgs`. The same structure was supposed to be used when parsing `LocalBytecode` - a bytecode obtained as a result of local compilation. However, for now constructor args cannot be extracted from the structure. In the next PRs I'll add the constructor arguments into the structure, and add a division indicating whether the structure was obtained from the requester data or from the local compilation result.


verification/src/solidity/verifier.rs
```diff
@@ -62,7 +292,7 @@ pub(crate) struct Verifier {
     file_path: Option<String>,
     /// Bytecode used on the contract creation transaction
     bc_creation_tx_input: BytecodeWithConstructorArgs,
-    /// Bytecode stored in the chain and being used by EVM
+    /// Bytecode stored in the chain and being used by EVMrap_err()
```
rimrakhimov, line 295:
```suggestion
    /// Bytecode stored in the chain and being used by EVM
```

verification/src/solidity/verifier.rs
```diff
@@ -97,3 +327,329 @@ impl Verifier {
...
+
+    #[test]
+    fn test_initialization_with_valid_data() {
+        let verifier = Verifier::new(
+            DEFAULT_CONTRACT_NAME.to_string(),
+            None,
+            DEFAULT_CREATION_TX_INPUT,
+            DEFAULT_DEPLOYED_BYTECODE,
+        );
+        assert!(verifier.is_ok(), "Initialization with \"0x\" prefix failed");
```
sevenzing, line 363: why "with 0x prefix"? probably you mixed it up with next test

sevenzing, line 363: why "with 0x prefix"? probably you mixed it up with next test

sevenzing (commented):
what do you think about splitting `verifier.rs` into several files? 
but maybe we don't need it right now, I am not sure

verification/src/solidity/verifier.rs
rimrakhimov: Fixed, thanks!

rimrakhimov:
@sevenzing Personally, I prefer to keep it in one file for now. For me, creating a new folder will make it more difficult to navigate between them

verification/Cargo.toml
```diff
@@ -7,23 +7,28 @@ repository = "https://github.com/blockscout/blockscout-rs/verification"
...
 log = "0.4"
 env_logger = "0.9"
 semver = "1.0"
-config = "0.13"
-reqwest = { version = "0.11", features = ["json"] }
-actix-rt = "2.7.0"
 anyhow = "1.0"
 async-trait = "0.1"
-thiserror = "1.0"
+thiserror = { version = "1.0" }
```
leviska, line 22: = "1.0"?

verification/src/solidity/verifier.rs
```diff
@@ -1,51 +1,281 @@
...
-#[derive(Clone, Debug, Error)]
-pub(crate) enum InitializationError {}
+#[derive(Clone, Debug, PartialEq, Error)]
+pub(crate) enum InitializationError {
+    #[error("creation transaction input is not a valid hex string")]
+    InvalidCreationTxInput,
+    #[error("deployed bytecode is not a valid hex string")]
+    InvalidDeployedBytecode,
+    #[error("cannot parse metadata hash from deployed bytecode: {0}")]
+    MetadataHashParseError(String),
```
leviska, line 20: I suggest remove "error" suffixes from error enum variants, because it will be mostly used with the enum itself, which already has error in the name, i.e:

```rust
enum FooError {
  BarError,
  Baz,
}

FooError::BarError // two errors, too much
FooError::Baz // one error, nice
```

So I suggest changing `MetadataHashParseError` to `MetadataHashParse`

verification/src/solidity/verifier.rs
```diff
@@ -1,51 +1,281 @@
...
+    }
+}
 
 impl FromStr for DeployedBytecode {
     type Err = InitializationError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        todo!()
+        let bytes = DisplayBytes::from_str(s)
+            .map_err(|_| InitializationError::InvalidDeployedBytecode)?
```
leviska, line 150: pass original error too?

verification/src/solidity/verifier.rs
```diff
@@ -1,51 +1,281 @@
...
+        }
+
+        // Further we will cut bytes from encoded representation, but original raw
+        // bytes are still required for `DeployedBytecode.bytes`
+        // (cloning takes O(1) due to internal `bytes::Bytes` implementation)
+        let mut b = encoded.clone();
+
+        // Decode length of metadata hash representation
+        let metadata_hash_length = {
+            let b_len = b.len();
```
leviska, line 175: ?????

leviska: does borrow checker errors out here?

verification/src/solidity/verifier.rs
```diff
@@ -1,51 +1,281 @@
...
+            let mut length_bytes = b.split_off(b_len - 2);
+            length_bytes.get_u16() as usize
+        };
+
+        if b.len() < metadata_hash_length {
+            return Err(InitializationError::MetadataHashParseError(
+                "not enough bytes".to_string(),
```
leviska, line 182: more detailed error
"provied bytecode is less, than metadata length"

verification/src/solidity/verifier.rs
```diff
@@ -1,51 +1,281 @@
...
+            .0;
+
+        DeployedBytecode::try_from(bytes)
+    }
+}
+
+impl TryFrom<bytes::Bytes> for DeployedBytecode {
+    type Error = InitializationError;
+
+    fn try_from(encoded: bytes::Bytes) -> Result<Self, Self::Error> {
```
leviska, line 160: do we really need `bytes::Bytes` here? Isn't just `&[u8]` enough?

verification/src/solidity/verifier.rs
```diff
@@ -1,51 +1,281 @@
...
+
+        // Now decode the metadata hash itself
+        let metadata_hash = {
+            let b_len = b.len();
+            let encoded_metadata_hash = b.split_off(b_len - metadata_hash_length);
+            MetadataHash::from_cbor(encoded_metadata_hash)
+        };
+
+        if let Err(err) = metadata_hash {
+            let message = if err.is_custom() {
```
leviska, line 194: I'm too lazy to go to the library sources, but I assume you can't match on the error type or make this somehow statically checked and without .expect

verification/src/solidity/verifier.rs
```diff
@@ -1,51 +1,281 @@
...
+        DeployedBytecode::try_from(bytes)
+    }
+}
+
+impl TryFrom<bytes::Bytes> for DeployedBytecode {
+    type Error = InitializationError;
+
+    fn try_from(encoded: bytes::Bytes) -> Result<Self, Self::Error> {
+        // If metadata is present, last two bytes encode its length in a two-byte big-endian encoding
+        if encoded.len() < 2 {
```
leviska, line 162: I suggest adding a struct, which will declaratively describe the bytecode format, i.e.
```rust
struct Bytecode {
  // bytecode starts with this
  compiler: Compiler,
  // 4 zeroes specifying the next value
  guard: Guard,
  // something something size of 16 bytes
  foo: Foo,
  // last 2 bytes are the length
  length: u16,
}
```

so you could just look at it and understand what format bytecode is, instead of looking at the `parse` function and trying to guess something from it (like reading here, that last two bytes are the length etc)
and then implement parse on it (may be `DeployedBytecode` already is this struct, but then I'd like to see all the values)

_In C, one of the important usages of struct was memory mapping, so you could just convert a pointer to the raw memory (like &[u8]) to this struct and get all the correct values. Time had passed, there is a lot of nuance nowadays (like different endiannesses), but I think the idea in general to "declare" the format for the part of the memory (or bytes) is kinda nice._

verification/src/solidity/verifier.rs
```diff
@@ -97,3 +327,329 @@ impl Verifier {
...
+        assert_eq!(
+            verifier.unwrap_err(),
+            InitializationError::MetadataHashMismatch(Mismatch::expected(
+                DisplayBytes::from_str(DEFAULT_ENCODED_METADATA_HASH).unwrap()
+            ))
+        )
+    }
+
+    #[test]
+    fn test_initialization_with_creation_tx_input_as_invalid_hex_should_fail() {
```
leviska, line 395: https://go.dev/doc/effective_go#names
> Another short example is once.Do; once.Do(setup) reads well and would not be improved by writing once.DoOrWaitUntilDone(setup). Long names don't automatically make things more readable. A helpful doc comment can often be more valuable than an extra long name.

1) `test` - it's already a test, there is a `#[test]` thing on top of it and when you filter it with `cargo test` you too write `test` word. I think it is pretty clear, that this function is a test
2) `should_fail` - I think it is actually misleading. Your **test** will not fail, the code will work fine. But **one** of the functions inside will return error, yes. Usually I saw `should_fail` on the compilation tests, where compilation is actually failing, or on usual tests which are failing for some reason, and `should_fail` is a note to the testing software that this is correct. 
3) just rearrange some words to make it smaller without losing the information: 
`fn initialize_with_invalid_creation_tx_input()` 
(also I think it's pretty clear, that function with invalid input should, indeed, fail)

verification/src/solidity/verifier.rs
```diff
@@ -97,3 +327,329 @@ impl Verifier {
...
+    #[test]
+    fn test_initialization_with_deployed_bytecode_as_invalid_hex_should_fail() {
+        let invalid_input = "0xabcdefghij";
+        let verifier = Verifier::new(
+            DEFAULT_CONTRACT_NAME.to_string(),
+            None,
+            DEFAULT_CREATION_TX_INPUT,
+            invalid_input,
+        );
+        assert!(verifier.is_err(), "Verifier initialization should fail");
```
leviska, line 434: `expect_err`?

verification/src/solidity/verifier.rs
```diff
@@ -97,3 +327,329 @@ impl Verifier {
...
+    #[test]
+    fn test_deserialization_metadata_hash_without_solc_tag() {
+        // given
+        // { "bzzr0": b"d4fba422541feba2d648f6657d9354ec14ea9f5919b520abe0feb60981d7b17c" }
+        let hex =
+            "a165627a7a72305820d4fba422541feba2d648f6657d9354ec14ea9f5919b520abe0feb60981d7b17c";
+        let encoded = DisplayBytes::from_str(hex).unwrap().0;
+        let expected = MetadataHash { solc: None };
+
+        // when
```
leviska, line 496: imo the comments `given, when, then` are useless

verification/src/types/unexpected.rs
```diff
@@ -0,0 +1,72 @@
+//! Error utils
+
+use std::fmt;
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+/// Error indicating an expected value was not found.
+pub struct Mismatch<T> {
```
leviska, line 7: why not just two separate errors? 
`Mismatch` and `MismatchFound`

verification/src/types/unexpected.rs
```diff
@@ -0,0 +1,72 @@
+//! Error utils
```
leviska, line 1: `Useful error types`

leviska, line 1: may be leave here only the mismatch type and rename the file into `mismatch`?

leviska:
> @sevenzing Personally, I prefer to keep it in one file for now. For me, creating a new folder will make it more difficult to navigate between them

may be create two files without folder?

verification/src/solidity/verifier.rs
rimrakhimov: `bytes::Bytes` allows us to clone the structure for O(1) and easily share different parts which is used in current implementation. Besides that, it provides convenient methods to process underlying bytes

verification/src/solidity/verifier.rs
rimrakhimov: Didn't get your point, could you please clarify?

verification/src/solidity/verifier.rs
leviska: you can copy `&[u8]` easily too and there is a split function in slices too

verification/src/solidity/verifier.rs
leviska: it's just a question, and I saw an answer later :)

verification/src/types/unexpected.rs
rimrakhimov: Why to create different types while using just one would be enough to successfully do what it is supposed to and being clear about internal functionality and usage?

verification/src/solidity/verifier.rs
rimrakhimov: I prefer using an assertion here as verification failure is the part of what we are actually testing. `expect_err` mostly indicates that we are sure that the result is error due to some preconditions and we just want to get the internal error. Here it is not the case

verification/src/solidity/verifier.rs
rimrakhimov: I believe it is an unnecessary work and information right now. as It is an internal structure which is not exposed to outside the module. Current layout allows to effectively and conveniently operate with the structure internally and provide some public api for the users of that structure.

verification/src/solidity/verifier.rs
rimrakhimov: But you have to solve ownership and lifetime problems which do not occur when you use `bytes::Bytes`

verification/src/solidity/verifier.rs
rimrakhimov: In my experience, I've seen the tests which used that notation and were clearer for me. I believe it does not worsen the code, but still may be useful for someone

verification/src/types/unexpected.rs
rimrakhimov: Changed description

verification/src/solidity/verifier.rs
rimrakhimov: Agreed with (1). Removed the prefix from test functions.
(3) Rearranged words for specified test function.

About the (2) still believe `should_fail` is a useful suffix. It may be obvious that function should fail when _invalid_ word is used in the name of the function, but it becomes unclear and requires to look inside of the test when the word _invalid_ is absent (e.g. `deserialization_of_non_cbor_hex` - is not obvious that it should fail). Adding the suffix only to some of the tests which should fail (without _invalid_) and not to add it to others (with _invalid_) is inconsistent and would confuse users. Furthermore with `test` prefix removed (as is now) it becomes clear that it is `initialization_with_creation_tx_input_as_invalid_hex` which should fail.

verification/src/solidity/verifier.rs
rimrakhimov: Agree

verification/Cargo.toml
rimrakhimov: Fixed

verification/src/solidity/verifier.rs
rimrakhimov: It did, but seems not to be a problem anymore. Fixed that, thanks

rimrakhimov:
> may be create two files without folder?

What two files you suggest to split it into? Still believe that it would be better to leave it as it is now, and think about splitting after the whole functionality would be finished

verification/src/solidity/verifier.rs
leviska: that's not for public interface, but for other developers. With json, for example, you can just look at the json itself to understand the format. With binary formats you can't do that, so it is very helpful if you can see the format "specification" somewhere. Think about yourself after 3 months or about another developer, when you/he'll need to add/fix something connected to the binary format

leviska: it's like a comment explaining the format, but more strict

leviska: we can do this in another PR tho

verification/src/solidity/verifier.rs
```diff
@@ -97,3 +327,329 @@ impl Verifier {
...
+        let decoded =
+            MetadataHash::from_cbor(encoded).expect("Error when decoding valid metadata hash");
+
+        // then
+        assert_eq!(expected, decoded, "Incorrectly decoded");
+    }
+
+    #[test]
+    fn test_deserialization_metadata_hash_with_solc_as_version() {
```
leviska: is github lagging or you forgot to push some commits or didn't you removed the `test` from test names? https://github.com/blockscout/blockscout-rs/pull/8#discussion_r895070780

verification/src/solidity/verifier.rs
rimrakhimov: I'll create an issue to consider that question after PR will be merged

PR #9, opened 2022-06-02
add verification response

### LINKED ISSUE #11: Add verification response format
Currently we come up with the following verification response structure:

### Success
```
{
  "message": "OK",
  "result": {
    "ABI": "[{\n\"type\":\"event\",\n\"inputs\": [{\"name\":\"a\",\"type\":\"uint256\",\"indexed\":true},{\"name\":\"b\",\"type\":\"bytes32\",\"indexed\":false}],\n\"name\":\"Event\"\n}, {\n\"type\":\"event\",\n\"inputs\": [{\"name\":\"a\",\"type\":\"uint256\",\"indexed\":true},{\"name\":\"b\",\"type\":\"bytes32\",\"indexed\":false}],\n\"name\":\"Event2\"\n}, {\n\"type\":\"function\",\n\"inputs\": [{\"name\":\"a\",\"type\":\"uint256\"}],\n\"name\":\"foo\",\n\"outputs\": []\n}]\n",
    "CompilerVersion": "v0.2.1-2016-01-30-91a6b35",
    "ContractName": "Test",
    "ImplementationAddress": "0x000000000000000000000000000000000000000e",
    "IsProxy": "true",
    "OptimizationUsed": "1",
    "SourceCode": "pragma solidity >0.4.24;\n\ncontract Test {\nconstructor() public { b = hex\"12345678901234567890123456789012\"; }\nevent Event(uint indexed a, bytes32 b);\nevent Event2(uint indexed a, bytes32 b);\nfunction foo(uint a) public { emit Event(a, b); }\nbytes32 b;\n}\n"
  },
  "status": "0"
}
```
In case of success, status of "0" is returned with all contract data as a `result` value. The data returned partially copies the data received with a request. That is done purposefully, as such redundancy simplifies the requester allowing it not to keep that data internally.

### Verification Error
```
{
  "message": "Verification error: bytecode does not match compilation output",
  "result": nil,
  "status": "1"
}
```
If verification fails, non-zero status is returned with reason provided as a `message` value. `result` in that case if `nil`. 

Notice, that both success and verification error responses have `200` HTTP status code, and `status` field is used as a discriminant. The field is a string in order to ease introduction of new verification error statuses.

### Other errors
The server may also return `400` (`Bad Request`) in case if `deployedBytecode` or `creationTxInput` are invalid (e.g. are invalid hexes, or have different metadata hash values). As those data are extracted from the chain by the requester, they should be valid and it is the requester's responsibility. Thus, these errors are treated differently from the verification errors.

sevenzing:
Add struct `VerificationRespose` - response for all verification handlers

Closes #11

### DIFF
--- a/verification/src/http_server/handlers/verification/mod.rs
+++ b/verification/src/http_server/handlers/verification/mod.rs
@@ -1,6 +1,6 @@
 #![allow(dead_code)]
 
-use std::collections::HashMap;
+use std::{collections::HashMap, fmt::Display};
 
 use serde::{Deserialize, Serialize};
 
@@ -39,7 +39,7 @@ pub enum VerificationStatus {
     #[serde(rename = "1")]
     BytecodeDismatch,
     #[serde(rename = "99")]
-    UnknowError,
+    UnknownError,
 }
 
 impl VerificationStatus {
@@ -47,7 +47,7 @@ impl VerificationStatus {
         match self {
             Self::Ok => "OK",
             Self::BytecodeDismatch => "Bytecode doesn't match, please try again",
-            Self::UnknowError => "Unknow error",
+            Self::UnknownError => "Unknow error",
         }
     }
 }
@@ -61,9 +61,14 @@ impl VerificationResponse {
         }
     }
 
-    pub fn err(status: VerificationStatus, message: Option<String>) -> Self {
+    pub fn err(status: VerificationStatus) -> Self {
+        let message = status.default_message();
+        Self::err_with_message(status, message)
+    }
+
+    pub fn err_with_message(status: VerificationStatus, message: impl Display) -> Self {
         Self {
-            message: message.unwrap_or_else(|| status.default_message().into()),
+            message: format!("{}", message),
             result: None,
             status,
         }
--- a/verification/src/http_server/handlers/verification/mod.rs
+++ b/verification/src/http_server/handlers/verification/mod.rs
@@ -51,7 +51,7 @@ impl VerificationResponse {
 
     pub fn err(message: impl Display) -> Self {
         Self {
-            message: format!("{}", message),
+            message: message.to_string(),
             result: None,
             status: VerificationStatus::Failed,
         }
--- a/verification/src/http_server/handlers/verification/mod.rs
+++ b/verification/src/http_server/handlers/verification/mod.rs
@@ -37,40 +37,23 @@ pub enum VerificationStatus {
     #[serde(rename = "0")]
     Ok,
     #[serde(rename = "1")]
-    BytecodeDismatch,
-    #[serde(rename = "99")]
-    UnknownError,
-}
-
-impl VerificationStatus {
-    fn default_message(&self) -> &'static str {
-        match self {
-            Self::Ok => "OK",
-            Self::BytecodeDismatch => "Bytecode doesn't match, please try again",
-            Self::UnknownError => "Unknow error",
-        }
-    }
+    Failed,
 }
 
 impl VerificationResponse {
     pub fn ok(result: VerificationResult) -> Self {
         Self {
-            message: VerificationStatus::Ok.default_message().into(),
+            message: "OK".to_string(),
             result: Some(result),
             status: VerificationStatus::Ok,
         }
     }
 
-    pub fn err(status: VerificationStatus) -> Self {
-        let message = status.default_message();
-        Self::err_with_message(status, message)
-    }
-
-    pub fn err_with_message(status: VerificationStatus, message: impl Display) -> Self {
+    pub fn err(message: impl Display) -> Self {
         Self {
             message: format!("{}", message),
             result: None,
-            status,
+            status: VerificationStatus::Failed,
         }
     }
 }
--- a/verification/src/http_server/handlers/verification/mod.rs
+++ b/verification/src/http_server/handlers/verification/mod.rs
@@ -1,10 +1,71 @@
-use serde::Serialize;
+#![allow(dead_code)]
+
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
 
 pub mod routes;
 pub mod solidity;
 pub mod sourcify;
 
-#[derive(Debug, Serialize, PartialEq)]
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub struct ContractLibrary {
+    pub lib_name: String,
+    pub lib_address: String,
+}
+
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
 pub struct VerificationResponse {
-    pub verified: bool,
+    pub message: String,
+    pub result: Option<VerificationResult>,
+    pub status: VerificationStatus,
+}
+
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub struct VerificationResult {
+    pub contract_name: String,
+    pub compiler_version: String,
+    pub evm_version: String,
+    pub constructor_arguments: Option<String>,
+    pub contract_libraries: Option<Vec<ContractLibrary>>,
+    pub abi: String,
+    pub sources: HashMap<String, String>,
+}
+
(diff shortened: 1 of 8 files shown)


verification/src/http_server/handlers/verification/mod.rs
```diff
@@ -1,10 +1,71 @@
...
+}
+
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub enum VerificationStatus {
+    #[serde(rename = "0")]
+    Ok,
+    #[serde(rename = "1")]
+    BytecodeDismatch,
+    #[serde(rename = "99")]
+    UnknowError,
```
leviska, line 42: let's just make this 1 error
also unknow_n_

verification/src/http_server/handlers/verification/mod.rs
```diff
@@ -1,10 +1,71 @@
...
+impl VerificationResponse {
+    pub fn ok(result: VerificationResult) -> Self {
+        Self {
+            message: VerificationStatus::Ok.default_message().into(),
+            result: Some(result),
+            status: VerificationStatus::Ok,
+        }
+    }
+
+    pub fn err(status: VerificationStatus, message: Option<String>) -> Self {
```
leviska, line 64: may be make this as two separate functions
`fn err(status)` and `fn err_with_message(status, message)`?

leviska: also may be accept `impl Display`, not string, so we could send actual errors here too, like `parse_something().map_err(|e| err(UnknownError, e)?;`

rimrakhimov: I don't really like that we can send an error with `VerificationStatus::Ok`

If we decide to leave only `VerificationStatus::Ok` and `VerificationStatus::Error` options for now, it would be simpler as we will not need to pass the status in `err()` method.

rimrakhimov: Also do we really need a default messages? What if `message` field would be empty in case if no messages are there at all (e.g. status is Ok)?

verification/src/http_server/handlers/verification/mod.rs
```diff
@@ -1,10 +1,71 @@
...
+            Self::BytecodeDismatch => "Bytecode doesn't match, please try again",
+            Self::UnknowError => "Unknow error",
+        }
+    }
+}
+
+impl VerificationResponse {
+    pub fn ok(result: VerificationResult) -> Self {
+        Self {
+            message: VerificationStatus::Ok.default_message().into(),
```
rimrakhimov, line 58:
```suggestion
            message: VerificationStatus::Ok.to_string(),
```
I suggest to implement `Display` trait for `VerificationStatus` and use `to_string()` method on it

verification/src/http_server/handlers/verification/mod.rs
```diff
@@ -1,10 +1,71 @@
...
+
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
+pub enum VerificationStatus {
+    #[serde(rename = "0")]
+    Ok,
+    #[serde(rename = "1")]
+    BytecodeDismatch,
+    #[serde(rename = "99")]
+    UnknowError,
+}
```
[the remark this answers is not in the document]
sevenzing: perhaps this approach is a little more clear, however it is much larger and it seems that it is not really needed, moreover we have to repeat error code two times, it may cause strange bugs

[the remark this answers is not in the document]
rimrakhimov: It seems to be beneficial from two points of view:
1. If new errors are added, we are enforced by the compiler to process new errors during serialization process, while by `serde(rename)` it is not required (and we are allowed not to specify `serde(rename)`)
2. If we decide to change the type of the status (e.g. to be an uint), it will be impossible to do with current approach

I agree, that error codes duplication may cause some bugs, but in general, seems that `Deserialization` is required only for testing purposes and is not going to be used in production purposes. Thus, the greatest problem it may cause is not to let tests pass. But if that happens, we may easily understand that we forgot to update `TryFrom<String> for VerificationStatus`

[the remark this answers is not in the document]
rimrakhimov: @leviska What do you think?

verification/src/http_server/handlers/verification/mod.rs
```diff
@@ -1,10 +1,59 @@
...
+        Self {
+            message: "OK".to_string(),
+            result: Some(result),
+            status: VerificationStatus::Ok,
+        }
+    }
+
+    pub fn err(message: impl Display) -> Self {
+        Self {
+            message: format!("{}", message),
```
rimrakhimov, line 54:
```suggestion
            message: message.to_string(),
```

verification/src/http_server/handlers/verification/mod.rs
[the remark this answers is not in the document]
rimrakhimov: Sorry, forgot two directives:
```
#[serde(into = "String")]
#[serde(try_from = "String")]
```
Otherwise, it works
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2fcce260c1e89a12e456828e749878fa

verification/src/http_server/handlers/verification/mod.rs
```diff
@@ -1,10 +1,59 @@
...
+use std::{collections::HashMap, fmt::Display};
+
+use serde::{Deserialize, Serialize};
 
 pub mod routes;
 pub mod solidity;
 pub mod sourcify;
 
-#[derive(Debug, Serialize, PartialEq)]
+#[derive(Debug, Deserialize, Serialize, PartialEq)]
```
leviska, line 11: а мы точно хотим иметь какой-то код в mod.rs, а не тупо импорты подмодулей? 
мб вынести это в отдельный файл?
@rimrakhimov

leviska (approved):

verification/src/http_server/handlers/verification/mod.rs
rimrakhimov: I suggest to leave as it is now, as most probably we will change that layout further. But, in general, I'm not against extracting them

rimrakhimov (commented):
@sevenzing Could you please also add some tests for the serialization and deserialization of VerificationResponse?

verification/src/tests/parse.rs
```diff
@@ -0,0 +1,26 @@
...
+{
+    for (s, value) in tests {
+        let v: T = serde_json::from_str(s).unwrap();
+        assert_eq!(v, value);
+    }
+}
+
+pub fn test_serialize_json_ok<T>(tests: Vec<(T, Value)>)
+where
+    T: Debug + PartialEq + Serialize,
```
rimrakhimov, line 18:
```suggestion
    T: Serialize,
```
Seems that the code compiles without `Debug` and `PartialEq` traits

verification/tests/sourcify.rs
```diff
@@ -34,9 +34,16 @@ async fn should_return_200() {
...
+        "failed to verify contract, status is {}",
         resp.status()
     );
 
+    let body: VerificationResponse = test::read_body_json(resp).await;
+    assert_eq!(
+        body.status,
+        VerificationStatus::Ok,
+        "invalid verification status",
+    )
```
rimrakhimov, line 46: Is there any special reason we do not check `VerificationResult` of the response?

rimrakhimov (approved):

verification/tests/sourcify.rs
sevenzing: because I return empty response for now:

```VerificationResult {
                contract_name: "".into(),
                compiler_version: "".into(),
                evm_version: "".into(),
                constructor_arguments: None,
                contract_libraries: None,
                abi: "".into(),
                sources: HashMap::new(),
            };```

i will change this test in next pr

PR #13, opened 2022-06-03
Download compilers from github

### LINKED ISSUE #12: Download compilers from github
We need an ability to list all compilers from https://github.com/blockscout/solc-bin/releases and download any release from it

leviska:
Closes #12


verification/src/compiler/version.rs
```diff
@@ -0,0 +1,115 @@
...
+impl Display for ReleaseVersion {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "solc-v{}+commit.{}", self.version, self.commit)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct NightlyVersion {
+    pub version: Version,
+    pub date: String,
```
rimrakhimov, line 42: Maybe use `chrono::NaiveDate` and `chrono::NaiveDate::parse_from_str` to parse the string

rimrakhimov, line 42: Consider using `chrono::NaiveDate` and `chrono::NaiveDate::parse_from_str` to parse the string, so to ensure that the date string is also valid

verification/src/compiler/version.rs
```diff
@@ -0,0 +1,115 @@
...
+
+impl ReleaseVersion {
+    pub fn new(version: Version, commit: String) -> Self {
+        Self { version, commit }
+    }
+}
+
+/// Parses release version from string formated as
+/// `solc-v*VERSION*+commit.*COMMITHASH*`, example
+/// `solc-v0.8.9+commit.e5eed63a`
```
rimrakhimov, line 18: The comments are not shown as a doc. Most probably, you would like to write them over the `from_str` function

rimrakhimov, line 18: The comments are not shown as a doc. Most probably, you would like to write them over the `from_str` function

verification/src/solidity/github_fetcher.rs
```diff
@@ -0,0 +1,33 @@
+use crate::compiler::{fetcher::Fetcher, version::CompilerVersion};
+use async_trait::async_trait;
+use std::{collections::HashMap, path::PathBuf};
+use thiserror::Error;
+
+#[derive(Default)]
+pub struct GithubFetcher {
+    releases: HashMap<CompilerVersion, String>,
```
rimrakhimov, line 8: Could you please clarify, what the map contains?

rimrakhimov, line 8: What is the value of the map?

rimrakhimov (commented):
I've leaved some suggestions. Could you also please write unit tests for parsing of compiler versions, so that it would be possible to ensure any changes over the version's structure are not braking anything

verification/src/solidity/github_fetcher.rs
```diff
@@ -0,0 +1,191 @@
...
+                        CompilerVersion::from_str(&release.tag_name).map_err(ListError::TagName)?;
+                    releases.insert(version, solc.browser_download_url);
+                }
+            }
+
+            page_index += 1;
+        }
+
+        Ok(releases)
+    }
```
rimrakhimov, line 72: Cannot we do the same way it is in svm - maintain _list.json_ (https://solc-bin.ethereum.org/linux-amd64/list.json) in the repository which is updated every time new release is added, and download and parse only that?

verification/src/compiler/version.rs
leviska: we're really don't care much about date, while we could match by commit hash, date is mostly metainformation imo. I store it mostly so display will output the same string

leviska: (also it probably will make some trouble because chrono accepts "2000-01-30" and we have "2000.1.30"

verification/src/solidity/github_fetcher.rs
leviska: why?
you see, if we scan the releases, we get 100% guarantee, that they are valid (yes, you could scan and then delete the release but I'm talking about the other stuff)

while we need to constantly update the file: we need some way to add new releases to it in cicd, if for some reason we delete a release (may be it's broken or something), we need to manually change the file, while doing it we may forget to add/remove the comma in json or something and now the file is invalid

imo that's just a lot of extra work for nothing (actually for a lot of places where we can make an error)

why they do it? if you check the file carefully, you'll see that it actually have another links to the external file storage, which we don't and probably won't have (if we will, let's talk about it then) 

also svm doesn't use this file, it uses some data that actually comes with the svm cargo crate + constructs the url magically (because it downloads from repo, not the releases), which I think is actually wrong way to do it

sevenzing (approved):

verification/src/compiler/version.rs
rimrakhimov: > we're really don't care much about date, while we could match by commit hash, date is mostly metainformation imo. I store it mostly so display will output the same string

Get it. Then yes, probably we may omit checking the date here, though I personally still believe that it would be nice to have feature.

> (also it probably will make some trouble because chrono accepts "2000-01-30" and we have "2000.1.30")
Seems that we may parse with `chrono::NaiveDate::parse_from_str(s,  "%Y.%m.%d")`

verification/src/solidity/github_fetcher.rs
rimrakhimov: In my opinion, that may have 3 main advantages:
1. Iterating all releases from the repository may be quite a long process. As we have to do that on each startup that may considerable increase the cost of running a new instance. Additionally, if _github_ will return non-success response on some of our requests, it may create an inconsistency between versions our client considers as available and versions actually available (and, thus, with other instances).At the same time, downloading only one file on startup is cheap and we may just retry or not to run an instance at all in case of any problems.
2. We are binding ourselves to _github_ and its releases mechanism in that case. When specifying urls to download compilers in a separate file, we are free to keep them wherever we want (e.g. we may freely move to an S3 instance or another SVC provider). We may even keep part of the compilers at a different place from others (yet not sure whether it may be useful).
3. It is not quite clear how we would update the list of available compilers when the node is running. While when using a separate file we most probably will just have a periodic job which redownloads and parses a `list.json` file.

rimrakhimov: I see what you are talking about, but it seems that we should decide between full consistency (when we forget to update `list.json`) and swiftness (when we have to iterate through all releases and github pages on each startup), and, in my opinion, it is better to sacrifice consistency here, as it is something we may ensure by other means.

rimrakhimov: Though probably we may discuss that in a different issue

rimrakhimov (approved):

PR #14, opened 2022-06-07
Add response for sourcify verification

sevenzing:
> NOTE: I noted that we had wrong structure for `libraries` field, I asked customer about it, so it is better to do it as JSON map, not list of JSON maps. So, I  replaced `ContractLibrary` with `BTreeMap`

Sourcify verification takes 3 steps (see `verify_using_sourcify_client`):

1. Make sourcify request for verification
2. If needed, make sourcify request to get actual source files
3. Construct response from actual source files

`SourcifyApiClient` struct is responsible for API calls

`Files` struct  is wrapper for BTreeMap. I implemented type conversion functions for it. These functions are used for constructing response from metadata and `.sol` sources.

Note that using trait `SourcifyApi` allow to use [mock](https://docs.rs/mockall/latest/mockall/) in tests.

### DIFF
--- a/verification/src/http_server/handlers/verification/mod.rs
+++ b/verification/src/http_server/handlers/verification/mod.rs
@@ -1,18 +1,12 @@
 #![allow(dead_code)]
 
-use std::{collections::HashMap, fmt::Display};
+use std::{collections::BTreeMap, fmt::Display};
 
 use serde::{Deserialize, Serialize};
 
 pub mod routes;
 pub mod solidity;
 pub mod sourcify;
-
-#[derive(Debug, Deserialize, Serialize, PartialEq)]
-pub struct ContractLibrary {
-    pub lib_name: String,
-    pub lib_address: String,
-}
 
 #[derive(Debug, Deserialize, Serialize, PartialEq)]
 pub struct VerificationResponse {
@@ -27,9 +21,10 @@ pub struct VerificationResult {
     pub compiler_version: String,
     pub evm_version: String,
     pub constructor_arguments: Option<String>,
-    pub contract_libraries: Option<Vec<ContractLibrary>>,
+    pub optimization_runs: Option<usize>,
+    pub contract_libraries: BTreeMap<String, String>,
     pub abi: String,
-    pub sources: HashMap<String, String>,
+    pub sources: BTreeMap<String, String>,
 }
 
 #[derive(Debug, Deserialize, Serialize, PartialEq)]
@@ -73,10 +68,11 @@ mod tests {
                     compiler_version: "compiler_version".to_string(),
                     evm_version: "evm_version".to_string(),
                     constructor_arguments: Some("constructor_arguments".to_string()),
-                    contract_libraries: Some(vec![ContractLibrary {
-                        lib_name: "lib_name".to_string(),
-                        lib_address: "lib_address".to_string(),
-                    }]),
+                    optimization_runs: Some(200),
+                    contract_libraries: BTreeMap::from([(
+                        "some_library".into(),
+                        "some_address".into(),
+                    )]),
                     abi: "abi".to_string(),
                     sources: serde_json::from_str(
                         r#"{
@@ -93,10 +89,10 @@ mod tests {
                         "compiler_version": "compiler_version",
                         "evm_version": "evm_version",
                         "constructor_arguments": "constructor_arguments",
-                        "contract_libraries": [{
-                            "lib_name": "lib_name",
-                            "lib_address": "lib_address",
-                        }],
+                        "contract_libraries": {
+                            "some_library": "some_address",
+                        },
+                        "optimization_runs": 200,
                         "abi": "abi",
                         "sources": {
                             "source.sol": "content",
--- a/verification/src/http_server/handlers/verification/sourcify/api.rs
+++ b/verification/src/http_server/handlers/verification/sourcify/api.rs
@@ -1,17 +1,33 @@
-use super::types::{ApiFilesResponse, ApiRequest, ApiVerificationResponse};
+use crate::{VerificationResponse, VerificationResult};
+use actix_web::{error, error::Error};
+
+use super::types::{ApiFilesResponse, ApiRequest, ApiVerificationResponse, Files};
+
+#[async_trait::async_trait]
+pub(super) trait SourcifyApi {
+    async fn verification(
+        &self,
+        params: &ApiRequest,
+    ) -> Result<ApiVerificationResponse, reqwest::Error>;
+
+    async fn source_files(&self, params: &ApiRequest) -> Result<ApiFilesResponse, reqwest::Error>;
+}
 
 pub(super) struct SoucifyApiClient {
     host: String,
 }
 
 impl SoucifyApiClient {
-    pub(super) fn new(host: &str) -> Self {
+    pub fn new(host: &str) -> Self {
         Self {
             host: host.to_string(),
         }
     }
+}
 
-    pub(super) async fn verification(
+#[async_trait::async_trait]
+impl SourcifyApi for SoucifyApiClient {
+    async fn verification(
         &self,
         params: &ApiRequest,
     ) -> Result<ApiVerificationResponse, reqwest::Error> {
@@ -24,10 +40,7 @@ impl SoucifyApiClient {
         resp.json().await
     }
 
-    pub(super) async fn source_files(
-        &self,
-        params: &ApiRequest,
-    ) -> Result<ApiFilesResponse, reqwest::Error> {
+    async fn source_files(&self, params: &ApiRequest) -> Result<ApiFilesResponse, reqwest::Error> {
         let url = format!(
             "{host}/files/any/{chain}/{address}",
             host = self.host,
@@ -38,4 +51,47 @@ impl SoucifyApiClient {
 
         resp.json().await
     }
+}
+
+pub(super) async fn verify_using_sourcify_client(
+    sourcify_client: impl SourcifyApi,
+    params: ApiRequest,
+) -> Result<VerificationResponse, Error> {
+    let response = sourcify_client
+        .verification(&params)
+        .await
+        .map_err(error::ErrorInternalServerError)?;
+
+    match response {
+        ApiVerificationResponse::Verified { result: api_result } => {
+            let files = {
+                let contract_was_already_verified = api_result
+                    .first()
+                    .ok_or_else(|| error::ErrorInternalServerError("sourcify empty response"))?
+                    .storage_timestamp
+                    .is_some();
+                if contract_was_already_verified {
+                    let api_files_response = sourcify_client
+                        .source_files(&params)
+                        .await
+                        .map_err(error::ErrorInternalServerError)?;
+                    Files::try_from(api_files_response).map_err(error::ErrorInternalServerError)?
+                } else {
+                    params.files
+                }
+            };
+            let result = VerificationResult::try_from(files).map_err(error::ErrorBadRequest)?;
+            Ok(VerificationResponse::ok(result))
+        }
+        ApiVerificationResponse::Error { error } => Ok(VerificationResponse::err(error)),
+        ApiVerificationResponse::ValidationErrors { message, errors } => {
+            let error_message = format!("{}: {:?}", message, errors);
+            Err(error::ErrorBadRequest(error_message))
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    // TODO: add tests in this PR, using mocked SourcifyApi
 }
--- a/verification/src/http_server/handlers/verification/sourcify/api.rs
+++ b/verification/src/http_server/handlers/verification/sourcify/api.rs
@@ -1,0 +1,27 @@
+use crate::http_server::handlers::verification::sourcify;
+
+pub(super) async fn verification_request(
+    params: &sourcify::types::ApiRequest,
+    sourcify_api_url: &str,
+) -> Result<sourcify::types::ApiVerificationResponse, reqwest::Error> {
+    let resp = reqwest::Client::new()
+        .post(sourcify_api_url)
+        .json(&params)
+        .send()
+        .await?;
+
+    resp.json().await
+}
+
+pub(super) async fn verification_files(
+    params: &sourcify::types::ApiRequest,
+    sourcify_api_url: &str,
+) -> Result<sourcify::types::ApiFilesResponse, reqwest::Error> {
+    let url = format!(
+        "{}/files/any/{}/{}",
+        sourcify_api_url, params.chain, params.address
+    );
+    let resp = reqwest::Client::new().get(url).send().await?;
+
+    resp.json().await
+}
--- a/verification/src/http_server/handlers/verification/sourcify/api.rs
+++ b/verification/src/http_server/handlers/verification/sourcify/api.rs
@@ -1,27 +1,41 @@
-use crate::http_server::handlers::verification::sourcify;
+use super::types::{ApiFilesResponse, ApiRequest, ApiVerificationResponse};
 
-pub(super) async fn verification_request(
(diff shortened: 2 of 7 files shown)


verification/src/http_server/handlers/verification/sourcify/mod.rs
```diff
@@ -1,53 +1,20 @@
...
-        }
-    }
-}
-
 pub async fn verify(
     config: web::Data<Config>,
     params: Json<ApiRequest>,
 ) -> Result<Json<VerificationResponse>, Error> {
-    verification_request(config.as_ref(), &params.into_inner()).await
+    let sourcify_client = SoucifyApiClient::new(&config.sourcify.api_url);
```
rimrakhimov, line 17: I suggest to initialize `SourciftApiClient` only once (e.g. while configuring the corresponding route) and pass it as a parameter instead of config

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,97 @@
...
+            chain = params.chain,
+            address = params.address,
+        );
+        let resp = reqwest::get(url).await?;
+
+        resp.json().await
+    }
+}
+
+pub(super) async fn verify_using_sourcify_client(
```
rimrakhimov, line 56: What if move that as a `verify` method for `SoucifyApiClient` structure?

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,97 @@
...
+                    .is_some();
+                if contract_was_already_verified {
+                    let api_files_response = sourcify_client
+                        .source_files(&params)
+                        .await
+                        .map_err(error::ErrorInternalServerError)?;
+                    Files::try_from(api_files_response).map_err(error::ErrorInternalServerError)?
+                } else {
+                    params.files
+                }
```
rimrakhimov, line 81: It would be glad to leave comments describing why this logic is required

rimrakhimov (commented):
I'll left

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[async_trait::async_trait]
+pub(super) trait SourcifyApi {
+    async fn verification(
+        &self,
+        params: &ApiRequest,
+    ) -> Result<ApiVerificationResponse, reqwest::Error>;
+
+    async fn source_files(&self, params: &ApiRequest) -> Result<ApiFilesResponse, reqwest::Error>;
+}
```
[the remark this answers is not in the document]
sevenzing: As I said in description, I need this trait to mock sourcify api for tests

verification/src/http_server/handlers/verification/sourcify/metadata.rs
```diff
@@ -0,0 +1,187 @@
...
+                None
+            }
+        });
+        let contract_libraries = metadata.settings.libraries;
+        let abi = serde_json::to_string(&metadata.output.abi)?;
+        let sources = files
+            .0
+            .into_iter()
+            .filter(|(name, _)| !name.ends_with(METADATA_FILE_NAME))
+            .collect();
```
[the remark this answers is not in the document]
sevenzing: yes, good idea

verification/src/http_server/handlers/verification/mod.rs
```diff
@@ -27,9 +21,10 @@ pub struct VerificationResult {
     pub compiler_version: String,
     pub evm_version: String,
     pub constructor_arguments: Option<String>,
-    pub contract_libraries: Option<Vec<ContractLibrary>>,
+    pub optimization_runs: Option<usize>,
+    pub contract_libraries: BTreeMap<String, String>,
     pub abi: String,
-    pub sources: HashMap<String, String>,
+    pub sources: BTreeMap<String, String>,
```
leviska, line 27: to sort the sources in alphabetical order? is it important?

verification/src/http_server/handlers/verification/sourcify/api.rs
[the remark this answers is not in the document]
leviska: _as I suggested from the start :)_

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,97 @@
...
+
+pub(super) struct SoucifyApiClient {
+    host: String,
```
leviska, line 17: url::Url? and in config too

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub fn new(host: &str) -> Self {
+        Self {
+            host: host.to_string(),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl SourcifyApi for SoucifyApiClient {
+    async fn verification(
```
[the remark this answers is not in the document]
leviska: or rename to verify

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,97 @@
...
+            .post(&self.host)
+            .json(&params)
+            .send()
+            .await?;
+
+        resp.json().await
+    }
+
+    async fn source_files(&self, params: &ApiRequest) -> Result<ApiFilesResponse, reqwest::Error> {
+        let url = format!(
```
leviska, line 44: if host will be url, then you can do `host.join("files", "any", params.chain, params.address)`

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,97 @@
...
+    params: ApiRequest,
+) -> Result<VerificationResponse, Error> {
+    let response = sourcify_client
+        .verification(&params)
+        .await
+        .map_err(error::ErrorInternalServerError)?;
+
+    match response {
+        ApiVerificationResponse::Verified { result: api_result } => {
+            let files = {
```
leviska, line 67: I suggest may be moving each match arm into separate function

verification/src/http_server/handlers/verification/sourcify/api.rs
leviska: or at least this arm

verification/src/http_server/handlers/verification/sourcify/metadata.rs
```diff
@@ -0,0 +1,187 @@
...
+use serde::Deserialize;
+
+use crate::VerificationResult;
+
+use super::types::Files;
+
+const METADATA_FILE_NAME: &str = "metadata.json";
+
+#[derive(Debug, PartialEq, Deserialize)]
+
```
leviska, line 13: remove new line

verification/src/http_server/handlers/verification/sourcify/metadata.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub compilation_target: BTreeMap<String, String>,
+    pub optimizer: Optimizer,
+    pub libraries: BTreeMap<String, String>,
+    #[serde(rename = "camelCase")]
+    pub evm_version: Option<EvmVersion>,
+}
+
+#[derive(Debug, PartialEq, Deserialize)]
+pub struct Optimizer {
+    pub enabled: Option<bool>,
```
leviska, line 32: I'm not sure why you're declaring this, but you might wanna look into ether_solc::artifacts (but I think you've already looked)

verification/src/http_server/handlers/verification/sourcify/metadata.rs
```diff
@@ -0,0 +1,187 @@
...
+            .ok_or_else(|| anyhow::Error::msg("compilation target not found"))?
+            .1
+            .to_string();
+        let compiler_version = metadata.compiler.version;
+        let evm_version = metadata
+            .settings
+            .evm_version
+            .unwrap_or_default()
+            .to_string();
+        let optimization_runs = metadata.settings.optimizer.enabled.and_then(|enabled| {
```
leviska, line 78: fyi it is possible to have "enabled: false" and it will still do some optimizations (leaving enabled to nil and enabled:false is two different options, sadly)
I'll try to make a PR with this idea in mind

verification/src/http_server/handlers/verification/sourcify/api.rs
sevenzing: I propose this way:
```
let url = self
            .host
            .join(format!("files/any/{}/{}", &params.chain, &params.address).as_str())
            .expect("should be valid url");
```

verification/src/http_server/handlers/verification/sourcify/api.rs
sevenzing: I decided to remove this logic

the user could add unnecessary files to the request, which would be ignored by sourcify, but we would return them.
So, it it better to request files from sourcify in any case

verification/src/http_server/handlers/verification/sourcify/api.rs
sevenzing: I removed unclear logic from it, so i think there is no need to separate into functions

verification/src/http_server/handlers/verification/mod.rs
sevenzing: I saw that `ethers_solc` uses BTreeMap for `sources`, and you used it for `libraries` field. For me it's just convenient to use same structures for same fields
изображение

leviska: then why not just use ether_solc::Sources?

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl SourcifyApiClient {
+    pub fn new(host: Url, request_timeout: u64, verification_attempts: u64) -> Self {
+        Self {
+            host,
+            request_timeout,
+            verification_attempts,
+        }
+    }
+
```
rimrakhimov, line 35: Probably forget to remove underscore

verification/src/http_server/handlers/verification/mod.rs
sevenzing: In my opinion, direct mention of BTreeMap is better than type alias. Also `Sources` have `PathBuf` as key, and it will create unnecessary type conversions with error handling

verification/src/http_server/handlers/verification/sourcify/api.rs
sevenzing: no, it is like private inner function for `verification_request`, that will be called several times (see below)

sevenzing: I agree that naming may be wrong, maybe call it `verification_request_once` ?

verification/src/http_server/handlers/verification/sourcify/metadata.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub output: Output,
+}
+
+#[derive(Debug, PartialEq, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MetadataSettings {
+    pub compilation_target: BTreeMap<String, String>,
+    pub optimizer: Optimizer,
+    pub libraries: BTreeMap<String, String>,
+    #[serde(rename = "camelCase")]
```
rimrakhimov, line 28: Is there any reason to put macros once again while you have `rename_all` above the structure?

verification/src/http_server/handlers/verification/sourcify/metadata.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub version: String,
+}
+
+#[derive(Debug, PartialEq, Deserialize)]
+pub struct Output {
+    pub abi: serde_json::Value,
+}
+
+impl Files {
+    fn try_extract_metadata(self) -> Result<(Metadata, BTreeMap<String, String>), anyhow::Error> {
```
rimrakhimov, line 49: I suggest to name it `extract_metadata_and_source_files`

rimrakhimov (approved):

rimrakhimov (approved):

verification/src/http_server/handlers/verification/sourcify/metadata.rs
rimrakhimov: Get you. Do you have any idea why it is the case?

sevenzing: rename_all renames all fields in struct `MetadataSettings` (evm_version -> evmVersion)
but it doesn't (and shouldn't) apply this rule to every type of fields (e.g. for `EvmVersion`)

I think this reasonable, maybe I don't want this kind of behavior

verification/src/http_server/handlers/verification/mod.rs
leviska: not sure about pathbuf being a bad thing, but I missed that it's just a type alias

> direct mention of BTreeMap is better than type alias

yes and no. It is easier to read, yes, but imagine the library will change the type, for example from BTreeMap to HashMap, then you'll need to fix it in your code, while if you're using `Sources` the code might work fine 

but it's just food for thought, not asking to replace it (only if you want :) )

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,119 @@
...
+    async fn make_retrying_request<F, Fut, Response>(
```
leviska, line 37: let's make this just a usual function, not the method of the api client and accept `attempts` as a parameter

sevenzing: Also thought about it, but where to put this function?
Is `src/network.rs` ok ?

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,119 @@
...
+        &self,
+        request: F,
+    ) -> Result<Response, reqwest::Error>
+    where
+        F: Fn() -> Fut,
+        Fut: Future<Output = Result<Response, reqwest::Error>>,
```
leviska, line 43: can we make the output be generic parameter too? like with `tokio::spawn`

leviska: ah, we can't, because we need to check for is_err. let's make the error type generic then

verification/src/http_server/handlers/verification/sourcify/api.rs
leviska: `src/network.rs` is fine for me, but also you could just leave it here for the time being.

sevenzing: hm, ok. will leave it here

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,117 @@
...
+        Self {
+            host,
+            request_timeout,
+            verification_attempts,
+        }
+    }
+}
+
+pub async fn make_retrying_request<F, Fut, Response, Error>(
+    verification_attempts: u64,
```
leviska, line 39: just attempts, function doesn't know if it's verification or downloading cat photo from google

sevenzing: oops

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -0,0 +1,117 @@
...
+    request: F,
```
leviska, line 40: try to remove `F` and make request of type `Fut`, then you can call this with `make_retrying_request(move { /* code */ });`, see https://docs.rs/tokio/latest/tokio/fn.spawn.html

leviska (approved):

PR #15, opened 2022-06-10
Fix optimizer cast

leviska:
Fixes optimizer cast from flattened input to compiler input


rimrakhimov (approved):

rimrakhimov:
Before version 0.8.6 omitting the 'enabled' key was not equivalent to setting it to false and would actually disable all the optimizations. When we compile the source code with Remix with optimization disabled, it still automatically uses `false` as a value for `enabled` key.

This PR sets `enabled = false` in case if optimization runs are omitted. That interface makes the verifier behavior consistent with Remix.

https://docs.soliditylang.org/en/latest/using-the-compiler.html#compiler-input-and-output-json-description

PR #23, opened 2022-06-13
Implemented routes as methods on structs

leviska:
We need to init the values separately from the config function, plus we'd like to have an async initialization, so I propose this struct-styled initialization to both have a one entry for the config and to declare all the data and methods locally

### DIFF
--- a/verification/src/http_server/routes.rs
+++ b/verification/src/http_server/routes.rs
@@ -10,10 +10,10 @@ pub struct AppConfig {
 }
 
 impl AppConfig {
-    pub fn new(config: Config) -> anyhow::Result<Self> {
+    pub async fn new(config: Config) -> anyhow::Result<Self> {
         Ok(Self {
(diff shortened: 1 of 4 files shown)


verification/src/http_server/routes.rs
```diff
@@ -4,12 +4,28 @@ use crate::Config;
...
 
-pub fn config(service_config: &mut web::ServiceConfig, app_config: Config) {
-    service_config
-        .app_data(web::Data::new(app_config))
-        .route("/health", web::get().to(status))
-        .service(
-            web::scope("/api/v1")
-                .service(web::scope("/verification").configure(verification::routes::config)),
-        );
+pub struct AppConfig {
```
sevenzing, line 7: what do you think about struct `AppRouter` with method `register_routes` ?

leviska:
Overriden by #24

PR #24, opened 2022-06-14
Update routes configuration

rimrakhimov:
Updates how different routes are configured. Changes allow to disable `sourcify` or `verificaion` endpoints from the configuration file. Updates the structure of http server module and extract routers configuration out of the handlers into a separate module.

Changes continues #23. It fixes repeating structure initialization problem, which made structures to be reinitialized (and, thus, reestablish all connections and re-download all data) for every thread HTTP server runs. 

Changes continues #21. It fixes parsing config arguments from CLI for integration sourcify test, and makes use of default values instead.


sevenzing (approved):

verification/src/http_server/routers/app.rs
```diff
@@ -0,0 +1,45 @@
...
+use actix_web::web;
+
+pub struct AppRouter {
+    verification: Option<SolidityRouter>,
+    sourcify: Option<SourcifyRouter>,
+}
+
+impl AppRouter {
+    pub async fn new(config: Config) -> anyhow::Result<Self> {
+        let verification = match config.verifier.disabled {
```
leviska, line 12: https://doc.rust-lang.org/std/primitive.bool.html#method.then

verification/src/http_server/routers/solidity.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub async fn new() -> anyhow::Result<Self> {
+        let fetcher = GithubFetcher::new("blockscout", "solc-bin", "compilers/".into())
+            .await
+            .map_err(anyhow::Error::msg)?;
+        Ok(Self {
+            cache: web::Data::new(DownloadCache::new(fetcher)),
+        })
+    }
+
+    pub fn register_routes(&self, service_config: &mut web::ServiceConfig) {
```
leviska, line 20: let's implement this as a trait `RouteRegiter` or something like that

verification/src/http_server/routers/app.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    pub fn register_routes(&self, service_config: &mut web::ServiceConfig) {
+        service_config
+            .route("/health", web::get().to(status::status))
+            .service(
+                web::scope("/api/v1").service(
+                    web::scope("/verification")
+                        .configure(|service_config| {
+                            if let Some(client) = self.verification.as_ref() {
```
leviska, line 33: if you implement trait `RouteRegister`, then you can make a function which will take an impl trait and return a lambda for `configure`, so you'll have less copypasta

verification/src/config.rs
```diff
@@ -25,16 +26,23 @@ impl Default for ServerConfiguration {
     }
 }
 
+#[derive(Default, Deserialize, Clone)]
+pub struct VerifierConfiguration {
+    pub disabled: bool,
```
leviska, line 31: I'm not sure about this. I don't see much a reason to disable some functionality from the service in real world usage
We mostly want to disable some services for tests, where we can do it by code, and moving it into config? I'm not sure

leviska: Actually, your way is worse, because we'll still need parameters for disabled services. Like, Sourcify requieres `SourcifyConfiguration` for it, but if we want to start only `verifier`, then we don't need this config
I thought about making a `new` function for all the services, which initializes all the children (and will be used in main), and may be an `empty` function, which doesn't initialize any children, with functions like `with_service_foo(service)`, which will init the `foo` child service
So it'll look like
```rust
let service = AppRoute::empty(config).with_verification(Verification::new().await.unwrap());
```
notice no sourcify config needed here

verification/src/http_server/routers/app.rs
rimrakhimov: It's nice idea, but seems that it may be a problem to `await` inside an `Option`. Not sure that it is applicable in our case, though try to research

verification/src/config.rs
rimrakhimov: > I'm not sure about this. I don't see much a reason to disable some functionality from the service in real world usage
We mostly want to disable some services for tests, where we can do it by code, and moving it into config? I'm not sure

I believe it may be a useful feature actually. Some may wish to disable sourcify proxy and use only internal verifier, for example which seems quite possible.

> Actually, your way is worse, because we'll still need parameters for disabled services.

verification/src/http_server/routers/app.rs
leviska: yeah, await probably won't work

verification/src/config.rs
leviska: > It does, but is it a problem actually?
Right now it isn't, but if a service will accept something long to load. For example, imagine if `VerificationRoute` would accept `GithubFetcher` as an argument, then we'll need to load it

And actually, speaking of which, I think there may be another way to make optional services

verification/src/http_server/routers/app.rs
rimrakhimov: Applied that for the sourcify initialization (without `.await` call)

verification/src/config.rs
rimrakhimov: Do you have any idea how what we can do with that while leaving services disabling feature?

In general, we can overcome the problem by making `VerifierConfiguration` (`SourcifyConfiguration`) fields optional, and adding another structure which is obtained from the former and which sets absent fields with default values. But it requires another level of indirection and probably is better to implement only if current solution becomes too inefficient. If you have better idea we may try to implement it instead.

leviska, line 31: enabled

verification/src/config.rs
```diff
@@ -25,16 +26,23 @@ impl Default for ServerConfiguration {
...
+}
+
 #[derive(Deserialize, Clone)]
 pub struct SourcifyConfiguration {
-    pub api_url: url::Url,
+    pub disabled: bool,
```
leviska, line 36: enabled

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -44,6 +44,7 @@ where
     Fut: Future<Output = Result<Response, Error>>,
 {
     for _ in 1..attempts - 1 {
```
leviska, line 46: can you change it to `for _ in 0..attempts-1` please?

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -44,6 +44,7 @@ where
...
+        // TODO: what if attempts == 0?
```
leviska, line 47: we can accept [std::num::NonZeroUsize](https://doc.rust-lang.org/stable/std/num/struct.NonZeroUsize.html) (or u64)

verification/src/http_server/mod.rs
```diff
@@ -1,3 +1,24 @@
...
+pub async fn run(config: Config) -> std::io::Result<()> {
+    let socket_addr = config.server.addr;
+    log::info!("Verification server is starting at {}", socket_addr);
+    let app_config = Arc::new(
+        AppRouter::new(config)
+            .await
+            .expect("couldn't initialize the app"),
+    );
+    HttpServer::new(move || {
+        App::new().configure(|service_config| app_config.register_routes(service_config))
```
leviska, line 19: `configure_router`

verification/src/http_server/routers.rs
```diff
@@ -0,0 +1,26 @@
+mod app;
```
leviska, line 1: let's have one style everywhere: `module/mod.rs` or `module.rs`.

rimrakhimov: Agree. Personally I prefer the second style as it seems to be more descriptive. But that requires to change all other modules. Changed `routers.rs` to `routers/mod.rs` for now, and suggest to discuss style changing at another issue

verification/src/http_server/handlers/verification/sourcify/api.rs
```diff
@@ -36,14 +37,14 @@ impl SourcifyApiClient {
...
-    attempts: u64,
+    attempts: NonZeroUsize,
     request: F,
 ) -> Result<Response, Error>
 where
     F: Fn() -> Fut,
     Fut: Future<Output = Result<Response, Error>>,
 {
-    for _ in 1..attempts - 1 {
+    for _ in 0..(attempts.get() - 1) {
```
leviska, line 47: fyi brackets not needed

leviska (approved):

PR #27, opened 2022-06-15
Solidity verifier verificaiton

rimrakhimov:
Continues #8

This PR implements the actual verification.


verification/src/solidity/verifier.rs
```diff
@@ -311,10 +416,135 @@ impl Verifier {
...
+
+        let constructor_args = self.extract_constructor_args(abi.constructor(), &bytecode)?;
+
+        // We do not know file path and contract name yet. We may create private structure
+        // to be returned from `compare` function which do not include those fields,
+        // but it would make the code less clear. At the same time the overhead of setting
+        // empty strings here is really small (it is done only once per verification).
+        Ok(VerificationSuccess {
+            file_path: "".to_string(),
+            contract_name: "".to_string(),
```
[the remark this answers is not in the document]
sevenzing: oh, saw your comment above. 
I suggest create new struct `ComparisonSuccess` that only has `abi` and `constructor_args` nad return this from .compare() func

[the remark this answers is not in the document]
rimrakhimov: > Maybe pass file_path and contract_name as function arguments?

For me, not passing those values as arguments makes a clear distinction between `compare` and `verify` methods, where the only `compare` responsibility is to check that the contract may be verified while `verify` takes care of iterating through all compiled contracts.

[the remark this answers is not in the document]
rimrakhimov: I agree that current solution may be confusing as well. Updated the `compare` method to return just a tuple of abi and constructor args.

sevenzing (approved):

verification/src/solidity/verifier.rs
```diff
@@ -311,10 +420,137 @@ impl Verifier {
...
+        }
+        Ok(())
+    }
+
+    /// Extracts constructor arguments from the creation transaction input specified on
+    /// [`Verifier`] initialization.
+    ///
+    /// Returns `Err` if constructor arguments cannot be extracted (should not be the case
+    /// if `Bytecode.verify_bytecode_with_extra_data` was called before).
+    fn extract_constructor_args(
```
leviska, line 510: why do you have `constructor_args` as `impl Bytecode`, and `extract_constructor_args` as `impl Verifier`?

verification/src/solidity/verifier.rs
```diff
@@ -274,18 +329,73 @@ impl BytecodeWithConstructorArgs {
...
+            )))
+        }
+    }
+
+    /// Verifies that bytecode and extra data obtained from creation transaction input
+    /// corresponds to the bytecode and extra data obtained from local compilation result.
+    pub fn verify_bytecode_with_extra_data(
+        &self,
+        compiled_bytecode: &Bytecode<CompilationResult>,
+    ) -> Result<(), VerificationError> {
```
leviska, line 369: It seems like an external function, and not a method of the class, may be make this as 
```rust
pub fn verify_bytecode_with_extra_data(
    transaction: &Bytecode<CreationTxInput>,
    compiled: &Bytecode<CompilationResult>,
)
```

leviska: or may be I wrong?

verification/src/solidity/verifier.rs
```diff
@@ -210,21 +238,48 @@ impl TryFrom<bytes::Bytes> for DeployedBytecode {
...
+/// Marker type under [`Bytecode`] indicating that the struct was obtained from the result of local compilation.
+struct CompilationResult;
+
 /// Wrapper under `evm.bytecode.object` from the standard output JSON
 /// (https://docs.soliditylang.org/en/latest/using-the-compiler.html#output-description)
 /// excluding metadata hash and optionally including constructor arguments used on a contract creation.
-#[derive(Clone, Debug, PartialEq)]
-struct BytecodeWithConstructorArgs {
+#[derive(PartialEq)]
+struct Bytecode<Source> {
```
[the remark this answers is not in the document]
rimrakhimov: If argument type is `Bytecode<CreationTxInput> then passing `Bytecode<CompilationResult>` would be invalid. So basically it allows us to differentiate between Bytecode(s) obtained from different sources and define methods that use this differentiation to ensure correctness.

verification/src/solidity/verifier.rs
rimrakhimov: They are a little bit different. `constructor_args` is mostly just a sequence of bytes that happens to occur in `Bytecode` after all metadata and aux data information. Thus, we obtain that sequence of bytes from `Bytecode` structure. At the same time `extract_constructor_args` is more related to `Verifier` as it gets that sequence obtained from `Bytecode` and ensure that it is valid constructor args according to abi of the contract.

verification/src/solidity/verifier.rs
rimrakhimov: It may be added as external function as well. But I preferred to make it as a method, just because the code seems to be more structured that way, imo

rimrakhimov: I see what you are talking about. Actually, that can be a method on `Verifier` structure and it may even make more sense. However the reason why I've decided to implement it as a method on `Bytecode`, as I want to hide internal structure of `Bytecode` from outside users. So that way internal representation of `Bytecode` only matters for the structure itself

verification/src/solidity/verifier.rs
```diff
@@ -274,18 +329,73 @@ impl BytecodeWithConstructorArgs {
...
 
+impl Bytecode<CreationTxInput> {
+    /// Extract constructor arguments using the bytecode obtained as a result of local compilation.
+    /// If there are no constructor arguments, returns `Ok(None)`, otherwise returns `Ok`
+    /// with encoded constructor arguments. If the extraction fails, returns `Err`.
+    pub fn constructor_args(
+        &self,
+        compiled_bytecode: &Bytecode<CompilationResult>,
+    ) -> Result<Option<bytes::Bytes>, VerificationError> {
+        if let Some(constructor_args) = self
```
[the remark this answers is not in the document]
leviska: well it's just strange to see this pattern for me (I don't really see difference with match in term of explaining flow), but if you think it's better, you can keep it

verification/src/solidity/verifier.rs
```diff
@@ -311,10 +420,137 @@ impl Verifier {
...
     /// with compiler output received when compiling source data locally.
     ///
-    /// If verification succeeds return [`Ok`], otherwise when verification
-    /// fails return an [`VerificationError`] inside [`Err`].
-    pub fn verify(&self, output: CompilerOutput) -> Result<(), VerificationError> {
-        todo!()
+    /// Iterates through all contracts received from local compilation and
+    /// returns [`VerificationSuccess`] with corresponding file path and contract name
+    /// if any contract  the contract that succeeds the verification. Otherwise, returns [`None`].
+    pub fn verify(&self, output: CompilerOutput) -> Option<VerificationSuccess> {
```
leviska, line 426: is this used somewhere? or is it for future? function returns strange thing for me, but usage could explain it. If it's for the future, it's fine, I'll look into it when we'll use it

leviska: I mean for me the part "if any contract the contract that succeeds the verification" is strange, not the return type

verification/src/solidity/verifier.rs
```diff
@@ -311,10 +420,137 @@ impl Verifier {
...
+        let deployed_bytecode = {
+            let bytes = contract
+                .get_deployed_bytecode_bytes()
+                .ok_or(VerificationError::MissedLibrary)?;
+            DeployedBytecode::try_from(bytes.0.clone())
+                .map_err(|err| VerificationError::InvalidDeployedBytecode(err.to_string()))?
+        };
+        let bytecode = {
+            let bytes = contract.get_bytecode_bytes().expect(
+                "contract compiled locally and has valid deployed bytecode: bytecode must be valid",
```
leviska, line 460: why do we panic here? is this something impossible?

leviska (approved):

verification/src/solidity/verifier.rs
rimrakhimov: > is this used somewhere? or is it for future? function returns strange thing for me, but usage could explain it. If it's for the future, it's fine, I'll look into it when we'll use it

The structure returns information obtained from the contract succeeded verification. It would be used further

PR #32, opened 2022-06-17
version_list endpoint

### LINKED ISSUE #33: Add handler for list of all compiler versions

sevenzing:
closes #33

### DIFF
--- a/verification/src/compiler/compilers.rs
+++ b/verification/src/compiler/compilers.rs
@@ -1,8 +1,10 @@
-use crate::compiler::{download_cache::DownloadCache, fetcher::Fetcher, version::CompilerVersion};
+use crate::compiler::{CompilerVersion, DownloadCache, Fetcher};
 use anyhow::anyhow;
 use ethers_solc::{error::SolcError, CompilerInput, CompilerOutput, Solc};
 use std::fmt::{Debug, Display};
 use thiserror::Error;
+
+use super::fetcher::VersionList;
 
 #[derive(Debug, Error)]
 pub enum CompilersError {
@@ -42,5 +44,11 @@ impl<T: Fetcher> Compilers<T> {
         let output = solc.compile(&input)?;
 
         Ok(output)
+    }
+}
+
+impl<T: VersionList> Compilers<T> {
+    pub fn all_versions(&self) -> Vec<&CompilerVersion> {
+        self.fetcher.all_versions()
     }
 }
--- a/verification/src/compiler/compilers.rs
+++ b/verification/src/compiler/compilers.rs
@@ -1,10 +1,8 @@
-use crate::compiler::{CompilerVersion, DownloadCache, Fetcher};
+use crate::compiler::{CompilerVersion, DownloadCache, Fetcher, VersionList};
 use anyhow::anyhow;
 use ethers_solc::{error::SolcError, CompilerInput, CompilerOutput, Solc};
 use std::fmt::{Debug, Display};
 use thiserror::Error;
-
-use super::fetcher::VersionList;
 
 #[derive(Debug, Error)]
 pub enum CompilersError {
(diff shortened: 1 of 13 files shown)


verification/src/http_server/handlers/verification/solidity/version_list.rs
```diff
@@ -0,0 +1,21 @@
...
+    web::{self, Json},
+    Error,
+};
+
+pub async fn get_version_list(
+    cache: web::Data<DownloadCache<GithubFetcher>>,
+) -> Result<Json<VersionsResponse>, Error> {
+    let mut versions = cache.fetcher().all_versions();
+    // sort in descending order
+    versions.sort_by(|x, y| y.cmp(x));
```
leviska, line 17: `versions.sort()` ?

sevenzing: by defalut it will sort in ascending order

leviska, line 17: actually, I think that it would be better, if `all_versions` returned already sorted vector. There is no obvious way to me to make it clear in the trait (except commenting), but the idea is that this list is changed very rarely (once a day?) and we can sort it once a day and reuse the sorted vector after that 
and we could implement it in the download cache, for example?
but okay, this is some ammount of work that we probably don't need right now, I'll just make an issue about that

verification/src/compiler/fetcher.rs
```diff
@@ -6,4 +6,5 @@ use std::path::PathBuf;
 pub trait Fetcher {
     type Error;
     async fn fetch(&self, ver: &CompilerVersion) -> Result<PathBuf, Self::Error>;
+    fn all_versions(&self) -> Vec<&CompilerVersion>;
```
leviska, line 9: I think it would be better to split the trait into two `Fetcher` and `VersionList` or something

leviska: so in the tests, for example, you wouldn't need to impl empty function for mock fetchers

verification/src/http_server/handlers/verification/solidity/version_list.rs
leviska: oh, I missed that. may be it would be more readable to do `x.cmp(y).reverse()`, because it is very easy to miss, that it's `y.cmp(x)`, and not `x.cmp(y)` https://doc.rust-lang.org/std/cmp/enum.Ordering.html#method.reverse

sevenzing: Hm, maybe just change HashMap to BTreeMap? Then, we will need just to reverse list in handler function

leviska (approved):

leviska (approved):

verification/src/compiler/fetcher.rs
```diff
@@ -7,3 +7,7 @@ pub trait Fetcher {
     type Error: Send + Sync + 'static;
     async fn fetch(&self, ver: &CompilerVersion) -> Result<PathBuf, Self::Error>;
 }
+
+pub trait VersionList {
+    fn all_versions(&self) -> Vec<&CompilerVersion>;
```
rimrakhimov, line 12: As I remember @sevenzing suggested to return `BTreeSet<_>` in some place. What if we return `BTreeSet` here creating a contract between the trait and its users that the method would return versions in a sorted order?

@leviska What do you think?

verification/src/compiler/compilers.rs
```diff
@@ -44,3 +46,9 @@ impl<T: Fetcher> Compilers<T> {
         Ok(output)
     }
 }
+
+impl<T: VersionList> Compilers<T> {
```
rimrakhimov, line 50: It would be better to implement `VersionList` for `Compilers`

verification/src/compiler/fetcher.rs
sevenzing: I suggested use BTreeMap for type of `releases` in `CompilerFetcher` so that  versions are always sorted

But I don't think we need to return BTreeSet and introduce some contract in trait. don't see enough reason to do it

verification/src/compiler/compilers.rs
sevenzing: don't think that `impl<T: VersionList> VersionList for Compilers<T>` looks readable now) but I don't mind, okay

rimrakhimov (approved):

PR #34, opened 2022-06-17
Add version order

sevenzing:


verification/src/compiler/version.rs
```diff
@@ -123,6 +151,22 @@ impl Display for CompilerVersion {
     }
 }
 
+impl Ord for CompilerVersion {
+    fn cmp(&self, other: &Self) -> Ordering {
+        (self.version(), self.date(), self.commit()).cmp(&(
```
leviska, line 156: How we compare `date` with release and nightly? For example, `v0.4.0` and `v0.4.0-nightly..`? 
Looking at docs, it seems, like `None` will be less, than `Some()`, so we'll have
`0.3.9, 0.4.0, 0.4.0-nightly, 0.4.1,...`, which isn't right

And actually I'd really prefer not to use general methods here and compare variants one by one, something like
```rust

// impl Ord for ReleaseVersion and NightlyVersion

// release
(&left.version, &left.commit).cmp(&(&right.version, &right.commit))

// nightly
(&left.version, &left.date, &left.commit).cmp(...)

// for CompilerVersion
fn cmp_release_nightly(left: Release, right: Nightly) -> Ordering {
  (&left.version, 1/*, &left.commit*/).cmp(&(&right.version, 0/*, &right.commit*/)) // any release is later than any nightly with the same version
  // you can leave out the commit
}

fn cmp(&self, other: &Self) -> Ordering {
  match self {
    ReleaseVersion(left) => {
      match other {
        ReleaseVersion(right) => {
          left.cmp(right)
        },
        NightlyVersion(right) => {
          cmp_release_nightly(left, right)
        },
      }
    },
    NightlyVersion(left) => {
      match other {
        ReleaseVersion(_) => {
          cmp_release_nightly(right, left).reverse()
        }
        NightlyVersion(right) => {
          left.cmp(right)
        }
      }
    }
  }
}
```

verification/src/compiler/version.rs
```diff
@@ -209,4 +253,93 @@ mod tests {
...
+        let mut versions: Vec<CompilerVersion> = versions
+            .iter()
+            .map(|s| CompilerVersion::from_str(s).expect("invalid version"))
+            .collect();
+        versions.sort();
+        let versions: Vec<String> = versions.iter().map(|v| v.to_string()).collect();
+        assert_eq!(
+            versions,
+            vec![
+                "solc-v0.5.2+commit.1df8f40c",
```
leviska, line 306: yep, this is wrong

verification/src/compiler/version.rs
```diff
@@ -209,4 +253,93 @@ mod tests {
...
+            "solc-v0.6.2-nightly.2020.1.14+commit.6dbadf69",
+            "solc-v0.6.2-nightly.2020.1.13+commit.408458b7",
+        ];
+        let mut versions: Vec<CompilerVersion> = versions
+            .iter()
+            .map(|s| CompilerVersion::from_str(s).expect("invalid version"))
+            .collect();
+        versions.sort();
+        let versions: Vec<String> = versions.iter().map(|v| v.to_string()).collect();
+        assert_eq!(
```
leviska, line 303: give an already sorted array and check `assert_eq!(versions, sorted)`
then also random shuffle it like 10-100 times and check `assert_eq!(versions, shuffled_then_sorted)`

verification/src/compiler/version.rs
leviska: Or, if you want your way, you can reorder fields in our structs and enum, and/or make some new structs, if needed, for which `derive(Ord)` will do the sorting
Something similar to your impl, but more cleaner
```rust
#[derive(PartialOrd, Ord)]
/*private*/ struct VersionSort {
  version: &Version,
  release: bool,
  date: Option<&Date>,
  commit: &[u8],
}

// and then
impl From<ReleaseVersion> for VersionSort;
impl From<NightlyVersion> for VersionSort;
impl From<CompilerVersion> for VersionSort;

fn cmp(self, other) -> Ordering {
  self.into::<VersionSort>.cmp(other.into())
}
```

leviska: @rimrakhimov what do you think?

verification/src/compiler/version.rs
leviska: but actually it would be better to have 3 independent tests, which compares
`release <=> release`
`nightly <=> release`
`release <=> nightly`
and in each test to have several comparisons (at least one for each Ordering)

verification/src/compiler/version.rs
rimrakhimov: @sevenzing proposed the idea to introduce a new function `CompilerVersion::is_release(&self) -> bool` and use that value as a second argument of a tuple. Personally, I support that idea, as it seems to be clearer.

verification/src/compiler/version.rs
rimrakhimov: I support with the idea with 3 tests one per each possible case. But shouldn't the last one be `nightly <=> nightly` instead of `release <=> nightly`?

verification/src/compiler/version.rs
```diff
@@ -99,6 +104,29 @@ pub enum CompilerVersion {
...
 
+impl CompilerVersion {
+    fn version(&self) -> &Version {
+        match self {
+            CompilerVersion::Nightly(v) => &v.version,
+            CompilerVersion::Release(v) => &v.version,
+        }
+    }
+
+    fn date(&self) -> Option<NaiveDate> {
```
rimrakhimov, line 115:
```suggestion
    fn date(&self) -> Option<&NaiveDate> {
```
Why not to return just a reference?

verification/src/compiler/version.rs
```diff
@@ -209,4 +266,141 @@ mod tests {
...
+            "solc-v4.3.2+commit.b8d736ae",
+            "solc-v5.1.1+commit.b8d736ae",
+            "solc-v5.3.1+commit.b8d736ae",
+            "solc-v6.3.2+commit.b8d736ae",
+            "solc-v7.1.1+commit.b8d736ae",
+            "solc-v7.1.2+commit.b8d736ae",
+            "solc-v7.2.1+commit.b8d736ae",
+            "solc-v7.2.2+commit.b8d736ae",
+            "solc-v7.3.1+commit.b8d736ae",
+            "solc-v7.3.2+commit.b8d736ae",
```
rimrakhimov, line 311: May we use actual commit hashes?

rimrakhimov:
@sevenzing It would be great also to add a test which compares two Release versions with the same version and different commit hashes, and two Nightly versions with the same version and date and different commit hashes. Just to indicate in tests that we considered such scenario and it was our deliberate choice to compare such versions on commit hashes (even though it does not make much sense)

verification/Cargo.toml
```diff
@@ -10,6 +10,8 @@ actix-web = "4"
 actix-rt = "2.7.0"
 reqwest = { version = "0.11", features = ["json"] }
 clap = { version = "3", features = ["derive"] }
+chrono = "0.4"
+rand = "0.8"
```
rimrakhimov, line 14: We can put it inside `[dev-dependencies]`

verification/src/compiler/version.rs
```diff
@@ -1,6 +1,10 @@
+use chrono::NaiveDate;
 use semver::Version;
-use std::{fmt::Display, str::FromStr};
+use std::{cmp::Ordering, fmt::Display, str::FromStr};
 use thiserror::Error;
+extern crate rand;
```
rimrakhimov, line 5: What is the purpose of that line here?

verification/src/compiler/version.rs
```diff
@@ -209,4 +266,141 @@ mod tests {
...
+    }
+
+    fn test_shuffle_and_sort(sorted: Vec<&str>) {
+        let sorted_versions = to_versions(sorted.clone());
+        let mut shuffled_versions = sorted_versions.clone();
+        shuffled_versions.shuffle(&mut thread_rng());
+        shuffled_versions.sort();
+        let shuffled: Vec<String> = shuffled_versions.iter().map(|v| v.to_string()).collect();
+        assert_eq!(shuffled, sorted);
+    }
```
rimrakhimov, line 284:
```suggestion
    fn test_shuffle_and_sort(sorted: Vec<&str>) {
        let to_versions: fn(Vec<&str>) -> Vec<CompilerVersion> = |versions| {
            versions
                .iter()
                .map(|s| CompilerVersion::from_str(s).expect("invalid version"))
                .collect()
        };
        let sorted_versions = to_versions(sorted.clone());
        let mut shuffled_versions = sorted_versions.clone();
        shuffled_versions.shuffle(&mut thread_rng());
        shuffled_versions.sort();
        let shuffled: Vec<String> = shuffled_versions.iter().map(|v| v.to_string()).collect();
        assert_eq!(shuffled, sorted);
    }
```

Nit: I believe putting `to_version` function inside a closure would be clearer to read, as `to_versions` is not used anywhere else

verification/src/compiler/version.rs
```diff
@@ -209,4 +266,141 @@ mod tests {
...
+            .collect()
+    }
+
+    fn test_shuffle_and_sort(sorted: Vec<&str>) {
+        let sorted_versions = to_versions(sorted.clone());
+        let mut shuffled_versions = sorted_versions.clone();
+        shuffled_versions.shuffle(&mut thread_rng());
+        shuffled_versions.sort();
+        let shuffled: Vec<String> = shuffled_versions.iter().map(|v| v.to_string()).collect();
+        assert_eq!(shuffled, sorted);
```
rimrakhimov, line 283: Why not to compare just `shuffled_versions` and `sorted_versions`?

sevenzing: in case of wrong order, test will show unreadable error:
изображение

rimrakhimov: Could you please add a small comment describing that?

rimrakhimov (approved):

rimrakhimov (approved):

verification/src/compiler/version.rs
```diff
@@ -209,4 +265,184 @@ mod tests {
...
+        );
+        assert!(
+            ver("solc-v0.5.14-nightly.2019.12.9+commit.d6667560")
+                > ver("solc-v0.5.2+commit.1df8f40c")
+        );
+    }
+
+    fn test_shuffle_and_sort(sorted: Vec<&str>, times: usize) {
+        let sorted_versions: Vec<CompilerVersion> = sorted
+            .clone()
```
leviska, line 318: `.into_iter()` ?

sevenzing: I need to use `sorted` after that at line 330, into_iter() takes ownership

sevenzing: oh, u are right, will change it to `iter().map()`

PR #41, opened 2022-06-22
New compiler fetcher

sevenzing:


verification/example_config.toml
```diff
@@ -6,3 +6,6 @@ addr = "0.0.0.0:8043"
 api_url = "https://sourcify.dev/server/"
 verification_attempts = 3
 request_timeout = 10
+
+[compiler]
+compilers_list_url = "https://binaries.soliditylang.org/linux-amd64/"
```
rimrakhimov, line 11: What if we rename `[verifier]` section into `[solidity]` (if vyper to be added, it would be another config anyway) and move the parameter into `[solidity]`?

rimrakhimov: And does not specifying the actual file seems a little confusing. What if we use the full path (including `/list.json` segment) allowing any file name to be used as a list? And if the `path` has no schema prefix, we will use the url without that last segment

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,295 @@
...
+        Ok(releases)
+    }
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct CompilerInfo {
+    pub path: DownloadPath,
+    pub long_version: String,
+    pub sha256: String,
```
rimrakhimov, line 59: I suggest to use more restrictive H256 (https://docs.rs/ethereum-types/latest/ethereum_types/struct.H256.html) type here

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+        Ok(releases)
+    }
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct CompilerInfo {
+    pub path: DownloadPath,
+    pub long_version: String,
```
rimrakhimov, line 58: The same here, couldn't we use a `CompilerVersion` as a type here?

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,295 @@
...
+    #[error("error parsing tag name into version: {0}")]
+    TagName(ParseError),
+    #[error("error parsing 'path' field: {0}")]
+    Path(url::ParseError),
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+pub struct ListJson {
+    pub builds: Vec<CompilerInfo>,
+}
```
rimrakhimov, line 33: Maybe keep builds just inside a `CompilerFetcher`? Why do we need a separate structure? We actually require that information even when downloading compilers (e.g. not implemented now but are going to add validating checsum on download)

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,295 @@
...
+            releases.insert(version, download_url);
+        }
+        Ok(releases)
+    }
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct CompilerInfo {
+    pub path: DownloadPath,
```
rimrakhimov, line 57: What if we would keep the final URL here (prepending paths without schema with url prefix on initialization)?

verification/src/solidity/compiler_fetcher.rs
leviska: may be just use array of u8? I don't really want to include this dependency for such simple thing

rimrakhimov: We may use something more lightweight: https://docs.rs/primitive-types/0.11.1/primitive_types/index.html or https://docs.rs/h256/0.1.0/h256/struct.H256.html (personally I prefer `primitive_types`; we may also consider using H160 for addresses, for example, or H128 for MD5)

.gitignore
```diff
@@ -19,3 +19,4 @@ Session.vim
 
 .pre-commit-config.yaml
 config.toml
+test_config.toml
```
rimrakhimov, line 22: Why is `test_config.toml` here, if it is required for testing purposes?

verification/src/config.rs
```diff
@@ -64,10 +71,18 @@ impl Default for SourcifyConfiguration {
 impl Config {
     pub fn parse() -> Result<Self, config::ConfigError> {
         let args = cli::Args::parse();
+        Self::from_file(args.config_path)
+    }
+
+    pub fn test() -> Result<Self, config::ConfigError> {
+        Self::from_file(PathBuf::from_str("test_config.toml").unwrap())
+    }
```
rimrakhimov, line 79: It's usually considered a bad practice to add a functionality required only for testing purposes. 

In our case it seems that `Config` has too broad functionality. I suggest to extract Args parsing functionality outside of the Config and leave `Config::from_file` as the only way to initialize Config structu

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,331 @@
...
+            })
+            .await
+            .map_err(FetchError::Shedule)??;
+        }
+
+        Ok(file)
+    }
+}
+
+mod types {
```
rimrakhimov, line 119: What if we name the module `deserialization_types` and move `CompilerInfo` outside? So that we will just separate intermediate types required for deserialization and types used by the CompilerFetcher

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,330 @@
...
+        );
+        assert_has_version(&releases,
+            "0.4.16+commit.d7661dd9",
+            "https://binaries.soliditylang.org/linux-amd64/download/files/solc-linux-amd64-v0.4.16+commit.d7661dd9"
+        );
+    }
+
+    #[tokio::test]
+    async fn list_download_releases() {
+        let config = Config::from_file(PathBuf::from_str("test_config.toml").unwrap()).unwrap();
```
rimrakhimov, line 306: Nit: Wouldn't it be easier just to make default config and then change the `solidity.compilers_list_url` to required url in case if `target_os="macos"` (or windows?)?

rimrakhimov (approved):

.gitignore
leviska, line 22: isn't this used in tests?

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,330 @@
+use self::deserialized_types::{DeserializedCompilerInfo, DownloadPath, ListJson};
```
leviska, line 1: ?

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+#[derive(Default)]
+pub struct CompilerFetcher {
+    releases: HashMap<CompilerVersion, CompilerInfo>,
+    folder: PathBuf,
+}
+
+impl CompilerFetcher {
+    pub async fn new(compilers_list_url: &Url, folder: PathBuf) -> Result<Self, ListError> {
```
leviska, line 57: let's may be make this type abstract about the list by implementing the new like
```rust
pub fn new(releases: HashMap..., folder: PathBuf) -> Self
```
and we could move `list_releases` from impl CompilerFetcher to just function and probably rename it to `fetch_compiler_list`

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,330 @@
...
+
+        Ok(file)
+    }
+}
+
+mod deserialized_types {
+    use super::*;
+
+    #[derive(Debug, Deserialize, PartialEq)]
+    pub struct ListJson {
```
leviska, line 146: see previous comment, rename it to `json::List`?

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,330 @@
...
+        pub builds: Vec<DeserializedCompilerInfo>,
+    }
+
+    #[derive(Debug, Deserialize, PartialEq)]
+    #[serde(rename_all = "camelCase")]
+    pub struct DeserializedCompilerInfo {
```
leviska, line 152: see previous comment, rename it to `json::CompilerInfo`?

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    #[derive(Debug, Deserialize, PartialEq)]
+    #[serde(untagged)]
+    pub enum DownloadPath {
+        Url(Url),
+        Filename(String),
+    }
+
+    impl ListJson {
+        pub fn into_releases(
```
leviska, line 167: `impl TryFrom<ListJson> for HashMap<.., ..>` ?

verification/src/config.rs
```diff
@@ -29,13 +27,20 @@ impl Default for ServerConfiguration {
...
+    pub compilers_list_url: Url,
 }
 
-impl Default for VerifierConfiguration {
+impl Default for SolidityConfiguration {
     fn default() -> Self {
-        Self { enabled: true }
+        Self {
+            compilers_list_url: Url::try_from(
+                "https://raw.githubusercontent.com/blockscout/solc-bin/main/list.json",
```
rimrakhimov, line 39: Let's maybe define default link to be dependent on the OS user runs the service. And use ethereum/solc-bin list.json for MacOS and Windows

rimrakhimov: Something like 
```
#[cfg(target_os = "linux")]
const DEFAULT_COMPILER_LIST: &'static str = "https://raw.githubusercontent.com/blockscout/solc-bin/main/list.json";
#[cfg(target_os = "macos")]
const DEFAULT_COMPILER_LIST: &'static str = "https://solc-bin.ethereum.org/macosx-amd64/list.json";
#[cfg(target_os = "windows")]
const DEFAULT_COMPILER_LIST: &'static str = "https://solc-bin.ethereum.org/windows-amd64/list.json";

impl Default for SolidityConfiguration {
    fn default() -> Self {
        Self {
            compilers_list_url: Url::try_from(
                DEFAULT_COMPILER_LIST,
            )
            .expect("valid url"),
            enabled: true,
        }
    }
}
```
That way we may create valid versions of list.json for all main OS

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,330 @@
...
+            })
+            .await
+            .map_err(FetchError::Shedule)??;
+        }
+
+        Ok(file)
+    }
+}
+
+mod deserialized_types {
```
leviska, line 142: we could potentially also serialize this types, so the naming imo not good.

may be just `mod json`, and not write `use self::json..`, but access it directly like `let a = json::List` or `json::CompilerInfo`

.gitignore
sevenzing: it is, but I need compiler_list_url to be `binaries.soliditylang.org/macosx-amd64/list.json` always (in `config.toml` too) and it would be just convinient for me not to commit this file

verification/src/config.rs
sevenzing: okay, let me remove `test_config.toml` for now we don't need it

verification/src/solidity/compiler_fetcher.rs
sevenzing: изображение

sevenzing: I need url for this (

verification/src/solidity/compiler_fetcher.rs
sevenzing: okaay, I can add `super::CompilerInfo` when it needed, but for me it will be move unclear

.gitignore
leviska: there is a concept of local and public gitignore https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally

verification/src/compiler/version.rs
```diff
@@ -108,25 +108,25 @@ pub enum CompilerVersion {
 }
 
 impl CompilerVersion {
-    fn version(&self) -> &Version {
+    pub fn version(&self) -> &Version {
```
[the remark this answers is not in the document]
sevenzing: convinient for tests

verification/src/solidity/compiler_fetcher.rs
sevenzing: @leviska Maybe introduce new type `struct Releases(HashMap<CompilerVersion, CompilerInfo>);`  and change Fetcher to
```
pub struct CompilerFetcher {
    releases: Releases,
    folder: PathBuf,
}
```

sevenzing: Then we can make `impl TryFrom<(ListJson, &Url)> for Releases`

verification/src/compiler/version.rs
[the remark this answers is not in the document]
leviska: > Ok, I thought about this and it seems like just my quirk, so think about this just as my thoughts, not a review comment.

I kinda dislike this functions, because instead of writing (for example)
```rust
let (semver, date, commit) = match ver {
  CompilerVersion::Nightly(v) => {
    (&v.version, Some(&date), &v.commit)
  }
  CompilerVersion::Release(v) => {
    (&v.version, None, &v.commit)
  }
}
```
you will probably write 
```rust
let semver = ver.version();
let date = ver.date();
let commit = ver.commit();
```
and you may ask "what's wrong with it, it's much clearer!", but in the second version there is 3 match clauses, instead of one, which, first of all, _blahblah performance blahblah_, but for me it's something about the semantics, like why do we need 6 matches instead of two.

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,326 @@
...
+            .await
+            .map_err(FetchError::Shedule)??;
+        }
+
+        Ok(file)
+    }
+}
+
+mod json {
+    use super::*;
```
rimrakhimov, line 155: Nit: what if we import only those structures that are required (`H256`, `Url`, `CompilerVersion`, and `Deserialize`)?

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,326 @@
...
+            })
+            .await
+            .map_err(FetchError::Shedule)??;
+        }
+
+        Ok(file)
+    }
+}
+
+mod json {
```
rimrakhimov, line 154: Nit: I suggest to move that module on the top of the file, so that users may easily see what json module we are using and not to confuse it with any external crate

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,326 @@
...
+    #[error("fetching list json returned error: {0}")]
+    ListJsonFetch(reqwest::Error),
+    #[error("cannot parse list json file: {0}")]
+    ParseListJson(reqwest::Error),
+    #[error("error parsing 'path' field: {0}")]
+    Path(url::ParseError),
+}
+
+#[derive(Default)]
+pub struct Releases(HashMap<CompilerVersion, CompilerInfo>);
```
rimrakhimov, line 26: Isn't `Release` too similar with `ReleaseVersion` compilers? I cannot come up with anything better, but maybe there is still a name that would be less associated with non-nightly versions?

rimrakhimov (approved):

verification/src/solidity/compiler_fetcher.rs
sevenzing: thought about it too but there was `releases` variable in github fetcher before, so decided to leave this name.

maybe `Compilers` / `CompilerVersions` / `Versions` ?

verification/src/main.rs
```diff
@@ -1,8 +1,10 @@
-use verification::{run_http_server, Config};
+use clap::Parser;
+use verification::{run_http_server, Args, Config};
 
 #[tokio::main]
 async fn main() -> std::io::Result<()> {
     env_logger::init();
-    let config = Config::parse().expect("Failed to parse config");
+    let args = Args::parse();
```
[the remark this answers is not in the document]
sevenzing: okay, but I will add `default()` instead of `new()`
изображение

verification/src/solidity/compiler_fetcher.rs
sevenzing: I prefer `Versions`, but if you don't agree, I can reset last [commit](https://github.com/blockscout/blockscout-rs/pull/41/commits/4a0133377df4b628e173ed445bb374f4e22c9b01)

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -0,0 +1,329 @@
...
+            .map_err(ListError::ListJsonFetch)?
+            .json()
+            .await
+            .map_err(ListError::ParseListJson)?;
+
+        Releases::try_from((list_json_file, compilers_list_url)).map_err(ListError::Path)
+    }
+}
+
+impl TryFrom<(json::List, &Url)> for Releases {
```
leviska, line 75: May be make a struct/type alias, which will serve as more descriptive naming, like
```rust
pub struct UrledList(List, Url);
// Or?
pub type UrledList = (List, Url);

impl TryFrom<UrledList>...

// Usage
let releases: Releases = UrledList(list, url).into()
```

And same for the CompilerInfo

sevenzing: urled list? sound like list of urls

leviska (approved):

PR #42, opened 2022-06-23
Implement aggregation of GET-requests

kuksag:
Added basic implementation of GET-requests handler, URL-builder, requests propagation, response combiner.

It allows us to aggregate GET-requests over API of maintained networks (e.g. https://blockscout.com/eth/mainnet/api-docs).

It's implemented in the way that a fixed thread pool is created on every request and API-calls are made.
Maintained instances and number of workers can be configured via `config.rs`.

There are 2 unit tests on `build_urls` and `merge_responses`, 1 integration test that makes a correct API-call over 3 networks.


multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+pub fn run(settings: Settings) -> Result<Server, std::io::Error> {
+    let listener = TcpListener::bind(settings.server.addr)?;
+
+    let server = HttpServer::new(move || {
+        // We want to pass the settings to the handler function, so we create clones: s1, s2
+        let s1 = settings.blockscout.clone();
+        App::new().route(
+            "/{_}", // We want to match every GET-request regardless of URL
```
kuksag, line 103: Apparently, there is an `App::default_service()`.
I will look into it.

multichain-api-gateway/src/config.rs
```diff
@@ -0,0 +1,38 @@
...
+#[derive(serde::Deserialize, Clone)]
+pub struct ServerSettings {
+    pub addr: SocketAddr,
+}
+
+#[derive(serde::Deserialize, Clone)]
+pub struct Settings {
+    pub server: ServerSettings,
+    pub blockscout: BlockScoutSettings,
+}
```
rimrakhimov, line 33: An interesting problem with that solution is that the service will fail to run if any of the fields would be missing and that is probably not the behavior you want (e.g. `blocscout.base_url` may be missing but we may safely run the service with `https://blockscou.com` as a default value). You may look at https://serde.rs/field-attrs.html#default for example, as one of the solutions

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+            web::get().to(move |request| {
+                let s2 = s1.clone();
+                router_get(request, s2)
```
rimrakhimov, line 106: You may consider using `web::Data<>` (https://actix.rs/docs/application/)

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+        .instances
+        .iter()
+        .map(|Instance(net, subnet)| {
+            let mut url = settings.base_url.clone();
+            url.set_path(&format!("{}/{}{}", net, subnet, path));
+            url.set_query(Some(query));
+            (Instance(net.clone(), subnet.clone()), url)
+        })
+        .collect::<Vec<_>>()
+}
```
rimrakhimov, line 24: I suggest to create urls (without query part) only once on service initialization. You can do that by creating a new structure (e.g. `Gateway`, but I'm not sure about naming) with the following layout: 
```
struct Gateway {
    apis: Vec<Url>,
    concurrent_requests: usize,
}
```
and which would be initialized from `BlockScoutSettings` via implementing `TryFrom` trait. You may initialize that structure in the `run`, so that url parsing would happen only once and shutdown the server if any variant could not result in valid url

rimrakhimov (commented):
@kuksag thank you for you contribution! I've left some comments and suggestions below. I'll look more thoroughly later today, and add additional feedback if would be any.

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+            let client = &client;
+            async move {
+                let resp = client.get(url).send().await.unwrap();
+                (instance.clone(), resp.bytes().await)
+            }
+        })
+        .buffer_unordered(concurrent_requests)
+        .collect::<Vec<_>>()
+        .await
+        .iter()
```
rimrakhimov, line 43: Is it actually required here? Why not to collect and await at the very end?

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    stream::iter(urls)
+        .map(|(instance, url)| {
+            let client = &client;
+            async move {
+                let resp = client.get(url).send().await.unwrap();
+                (instance.clone(), resp.bytes().await)
+            }
+        })
+        .buffer_unordered(concurrent_requests)
```
rimrakhimov, line 40: Nit: not sure but wouldn't use of `for_each_concurrent` give the same result but be more clear?

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+    concurrent_requests: usize,
+) -> Vec<(Instance, String)> {
+    let client = Client::new();
+
+    stream::iter(urls)
+        .map(|(instance, url)| {
+            let client = &client;
+            async move {
+                let resp = client.get(url).send().await.unwrap();
+                (instance.clone(), resp.bytes().await)
```
rimrakhimov, line 37:
```suggestion
            async {
                let resp = client.get(url).send().await.unwrap();
                (instance.clone(), resp.bytes().await)
```
You may do it the following way without copying the reference

rimrakhimov (commented):
I suggest to get rid of `serde_json::Map` return types. It makes the code more difficult to read and modify with not allowing to easily change the underlying data representation.

I suggest to create `type Responses = HashMap<String, HashMap<String, Value>>;` alias and return it from `merge_responses` function. `handle_default_request` may return `Json<Responses>`. Deserialization implemented on HashMap will serialize you responses automatically into json format.

multichain-api-gateway/src/lib.rs
kuksag: Yes, they would give the same result. 
But since I want to call another `map` after making requests, it's easier for me to stick with `buffer_unordered`. 
Otherwise, if I use `for_each_concurrent`, I have to collect it to `vector` and then make `.to_iter().map()` calls.

kuksag:
@rimrakhimov, here are milestones that were completed:
* Added CLI arguments
Now we can specify a path to the config file and pass CLI-arguments, that will overwrite options from the config 
* Moved URL-APIs building before running the server
It may not affect efficiency, but it will prevent us from using wrong base URL
* Fixed other remarks. I hope now it looks a little better

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/// Assumptions:
+/// 1. All api calls expected to have trailing path "/api"
+///     e.g. "<base_url>/<...>/<...>/api?<query>"
+/// 2. First two segments of path of api call expected to be (network, chain)
+///     e.g. "<base_url>/<network>/<chain>/<...>/<...>"
+/// Taking it to account, we expect the following api call urls:
+///     e.g. <base_url>/<network>/<chain>/api?<query>   
+impl TryFrom<BlockScoutSettings> for APIsEndpoints {
```
rimrakhimov, line 32: It would be better to implement `From` here if no errors may occur during conversion.

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    Json(merge_responses(responses))
+}
+
+pub async fn router_get(
+    request: HttpRequest,
+    apis_endpoints: Data<APIsEndpoints>,
+) -> impl Responder {
+    let json =
+        handle_default_request(request.query_string(), apis_endpoints.get_ref().clone()).await;
+    HttpResponse::Ok().json(json)
```
rimrakhimov, line 106: You may return just a `json` variable here without wrapping it under `HttpResponse::Ok`, as `Json<T>` already implements `Responder` trait

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+pub fn run(settings: Settings) -> Result<Server, std::io::Error> {
+    let listener = TcpListener::bind(settings.server.addr)?;
+
+    let apis_endpoints: APIsEndpoints = settings.block_scout.try_into().unwrap();
+
+    let server = HttpServer::new(move || {
+        App::new()
+            .app_data(Data::new(apis_endpoints.clone()))
+            .default_service(web::route().to(router_get))
+    })
```
rimrakhimov, line 118: Nit: The lambda function passed inside `HttpServer::new` is executed for every thread `actix-web` creates. Thus, here `apis_endpoints` cloning occurs for every thread created by `actix`. In general, it may not be a big problem, as the structure does not contain any complex fields. 

But, otherwise, you may want wrap the structure into `Data` earlier (e.g. `let apis_endpoints = Data::new(settings.block_scout.try_into().unwrap());`) and pass into `app_data` as already wrapped value.

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+fn enrich_apis(query: &str, apis_endpoints: &mut APIsEndpoints) {
+    apis_endpoints
+        .apis
+        .iter_mut()
+        .for_each(|(_, url)| url.set_query(Some(query)))
+}
+
+async fn handle_default_request(query: &str, mut apis_endpoints: APIsEndpoints) -> Json<Responses> {
+    enrich_apis(query, &mut apis_endpoints);
+    let responses = make_requests(apis_endpoints).await;
```
rimrakhimov, line 96: Why not to pass query into `make_requests` and call `set_query` before making a call?

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+            apis.push((Instance(net, subnet), url));
+        }
+        Ok(Self {
+            apis,
+            concurrent_requests: settings.concurrent_requests,
+        })
+    }
+}
+
+async fn make_requests(apis_endpoints: APIsEndpoints) -> Vec<(Instance, String)> {
```
rimrakhimov, line 49: Probably may be implemented as a method of the `APIsEndpoints` struct

multichain-api-gateway/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+use serde_json::Value;
+use url::Url;
+
+use crate::config::{BlockScoutSettings, Instance, Settings};
+
+mod cli;
+pub mod config;
+
+#[derive(Clone)]
+pub struct APIsEndpoints {
```
rimrakhimov, line 20: I believe `ApiEndpoints` would be more consistent with existing guidelines (https://rust-lang.github.io/api-guidelines/naming.html)

rimrakhimov (commented):
I believe it looks better now, thank you. I've added a some more comments below. You may decide not to address all of them, but at least have a look.

Write to me when you believe you are ready to merge the code, I'll do it for you as you may not have enough rights to do it for now.

Also, as a general suggestion it would be nice to implement Debug trait on the structures you use (especially in config module). It is usually expected for types to have `Debug` implemented, as it may be easily derived, there is almost no overhead and it allows to print the structure if you want to see what it contains for some reason (e.g. logging, or you just testing your application manually to ensure that it works the way you expect)

multichain-api-gateway/src/config.rs
```diff
@@ -0,0 +1,108 @@
...
+            addr: SocketAddr::from_str("0.0.0.0:8044").unwrap(),
+        }
+    }
+}
+
+#[derive(Deserialize, Clone, Default)]
+pub struct Settings {
+    pub server: ServerSettings,
+    pub block_scout: BlockScoutSettings,
+}
```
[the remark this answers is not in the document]
kuksag: Yes, thank you, I probably didn't think of it. I've added `#serde(default)` now.

PR #43, opened 2022-06-24
Add multi part enpoint

sevenzing:
remove FlattenSource verification and replace it with more comprehensive MultiFilesInput


verification/src/http_server/routers/solidity.rs
```diff
@@ -29,8 +29,8 @@ impl Router for SolidityRouter {
     fn register_routes(&self, service_config: &mut web::ServiceConfig) {
         service_config.app_data(self.cache.clone()).service(
             web::scope("/verify")
-                .route("/flatten", web::post().to(flatten::verify))
-                .route("/standard_json", web::post().to(standard_json::verify)),
+                .route("/", web::post().to(files_input::verify))
```
sevenzing, line 32: decided to leave this verification on `/api/v1/solidity/verify/`. We can move it to `/verify/files/` or something, but i don't think it's needed

verification/src/http_server/handlers/verification/solidity/types.rs
```diff
@@ -18,40 +18,47 @@ pub struct VerificationRequest<T> {
...
+                Some(EvmVersion::from_str(&input.evm_version).map_err(anyhow::Error::msg)?);
         } else {
             // `Settings::default()` sets the value to the latest available evm version (`Some(London)` for now)
             settings.evm_version = None
         }
+
+        let sources: Sources = input
+            .sources
+            .into_iter()
+            .map(|(name, content)| (name, Source { content }))
```
sevenzing, line 57: It would be great to find method of deserialization to `Source` struct, but i didnt come up with idea

verification/src/http_server/routers/solidity.rs
sevenzing: should I rename everything to "multi-part style". I mean rename structs to `MultiPart`, file to `multi-part.rs` and so on?

verification/src/http_server/routers/solidity.rs
```diff
@@ -31,8 +31,8 @@ impl Router for SolidityRouter {
             .app_data(self.cache.clone())
             .service(
                 web::scope("/verify")
-                    .route("/flatten", web::post().to(flatten::verify))
-                    .route("/standard_json", web::post().to(standard_json::verify)),
+                    .route("/multi_part/", web::post().to(multi_part::verify))
+                    .route("/standard_json/", web::post().to(standard_json::verify)),
```
rimrakhimov, line 35:
```suggestion
                    .route("/multi-part", web::post().to(multi_part::verify))
                    .route("/standard_json", web::post().to(standard_json::verify)),
```

rimrakhimov: It is better to use hyphens instead of underscores in REST paths

And what is the reason to add slashes at the end?

rimrakhimov: > maybe replace underscore in standard_json too?

I believe it would be better to do that in a separate PR (e.g. #46 which is related to making types and routes to correspond to the discussed Api)

rimrakhimov (approved):

leviska (approved):

PR #45, opened 2022-06-26
Add Readme

rimrakhimov:


verification/README.md
```diff
@@ -0,0 +1,210 @@
...
+    "B.sol": "pragma solidity ^0.8.14; contract B {}"
+  },
+  // Version of the EVM to compile for
+  "evm_version": "default",
+  // If present, optimizations are enabled with specified number of runs, 
+  // otherwise optmimizations are disabled
+  "optimization_runs": "200",
+  // If present, specify addresses of the libraries.
+  "contract_libraries": {
+    "MyLib": "0x123123..."
```
sevenzing, line 106: I think we should accept double map, like

```json
"contract_libraries": {
  "A.sol": {
    "MyLib": "0x123123..."
  }
}
```

rimrakhimov: We've decided to accept single level map here, as usually users do not specify what file the library belongs to. We will handle different cases internally, by bruteforcing different combinations of file and library names

verification/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Installing through cargo
+Another way to install the binary without cloning the repository is to use cargo straightway:
+```
+cargo install --git https://github.com/blockscout/blockscout-rs --bin verification
+```
+In that case, you can run the binary using just `verification`.
+
+## Configuration
+Currently, the service uses a configuration file the path to which is specified via CLI flag `--config-path=[path]`.
```
leviska, line 28: drop the "currently", jsut `(the?) Service uses...`

verification/README.md
```diff
@@ -0,0 +1,248 @@
...
+The configuration file may contain the following options:
+```toml
+[server]
+# IP address and port number the server should listen to
+addr = "0.0.0.0:8043"
+
+[solidity]
+# if false solidity verification is not supported
```
leviska, line 36: `enables solidity verification. when disabled, will not load solidity required data, like compilers list`

verification/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+[sourcify]
+# if false sourcify verification is not supported
+enabled = true
+# Sourcify API endpoint
+api_url = "https://sourcify.dev/server/"
+# number of failing attempts the server makes to Sourcify API 
+verification_attempts = 3
+# the maximum period the service is waiting for the Sourcify response
+request_timeout = 10
```
leviska, line 49: units? `period in seconds`?

verification/README.md
```diff
@@ -0,0 +1,248 @@
...
+```
+For all keys omitted from the configuration file default values from the example above are used.
+
+# Api
+
+---
```
leviska, line 55: github already draws the line 

image

leviska (approved):

verification/README.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Verification Failure
+If verification fails because of invalid verification data provided to it from outside,
+the service returns 200 with the failure status:
+```json5
+{
+  // Message indicating the reason for failure
+  "message": "Compilation error: contracts/3_Ballot.sol:4:1: ParserError: Expected pragma, import directive or contract/interface/library/struct/enum/constant/function definition.\n12312313vddfvfdvfd\n^------^",
+  // Status of 1 indicates verification failure
+  "status": 1
```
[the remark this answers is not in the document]
rimrakhimov: But the zero success code was described in the input before. 
Although, added a note that non-zero status indicates an error code

verification/README.md
rimrakhimov: I believe that it is better not to say `enables solidity verification` as it actual a default behavior. The flag mainly disables solidity verification, doesn't it?

verification/README.md
```diff
@@ -0,0 +1,244 @@
...
+### Input
+No input required
+
+### Output
+
+```json5
+{
+  // List of all available versions in descending order
+  "builds": ["0.8.15-nightly.2022.5.27+commit.095cc647","0.8.15-nightly.2022.5.25+commit.fdc3c8ee",..]
+}
```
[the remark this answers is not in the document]
sevenzing: I don't like this way. what if we will need to pass some other versions information (maybe last release, or stable release or smth)?  Or split builds into releases and nightly? 

If we use vector, we restrict ourself to only list of strings and nothing else.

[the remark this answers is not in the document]
sevenzing: also it's not that hard to write `response['data']['builds']` instead of `response['data']`

leviska (approved):

verification/README.md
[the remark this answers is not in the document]
leviska: I agree with Lev

PR #48, opened 2022-06-28
Update Compilers: add returning error if compilation fails

### LINKED ISSUE #49: Map result from solc correctly to include compilation error
If compilation results in error, errors are stored in `CompilerOutput.errors` and is considered as success now. We need to parse those errors and if there are any with `Error` severity level, return `Err` as a result of compilation

rimrakhimov:
Closes #49

### DIFF
--- a/verification/src/compiler/compilers.rs
+++ b/verification/src/compiler/compilers.rs
@@ -1,6 +1,10 @@
 use crate::compiler::{download_cache::DownloadCache, fetcher::Fetcher, version::CompilerVersion};
 use anyhow::anyhow;
-use ethers_solc::{error::SolcError, CompilerInput, CompilerOutput, Solc};
+use ethers_solc::{
+    artifacts::{self, Severity},
+    error::SolcError,
+    CompilerInput, CompilerOutput, Solc,
+};
 use std::fmt::{Debug, Display};
 use thiserror::Error;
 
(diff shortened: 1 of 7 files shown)


verification/src/compiler/compilers.rs
```diff
@@ -50,3 +67,131 @@ impl<T: VersionList> VersionList for Compilers<T> {
...
+
+    use async_once_cell::OnceCell;
+    use ethers_solc::artifacts::{Source, Sources};
+    use std::default::Default;
+
+    async fn global_compilers() -> &'static Compilers<CompilerFetcher> {
+        static COMPILERS: OnceCell<Compilers<CompilerFetcher>> = OnceCell::new();
+        COMPILERS
+            .get_or_init(async {
+                let url = "https://solc-bin.ethereum.org/linux-amd64/list.json"
```
sevenzing, line 85: *somewhere in the world a macos user is crying*

why not to read from config?

rimrakhimov: Hmm, makes sense

> why not read from config?

Because config is broader entity which is not related to Compilers module at all as required data from config is passed to Compilers separately as parameters. 

I believe it may make sense to extract the `DEFAULT_COMPILER_LIST` constant in a separate module and import it in Config as well as in Compilers

sevenzing: right, but why in seprate module? maybe here, in `compilers.rs` file?

rimrakhimov: Because we reuse it in both `compilers.rs` and `config.rs`

sevenzing: okay, its deal :)

leviska (approved):

sevenzing (approved):

PR #51, opened 2022-06-29
Add refresh compiler versions

sevenzing:
added cron job that will refresh all compilers info (`HashMap<CompilerVersion, CompilerInfo>`)

### DIFF
--- a/verification/src/compiler/compilers.rs
+++ b/verification/src/compiler/compilers.rs
@@ -1,6 +1,5 @@
 use crate::compiler::{CompilerVersion, DownloadCache, Fetcher, VersionList};
 use anyhow::anyhow;
-use async_trait::async_trait;
 use ethers_solc::{
     artifacts::{self, Severity},
     error::SolcError,
@@ -63,10 +62,9 @@ impl<T: Fetcher> Compilers<T> {
     }
 }
 
-#[async_trait]
-impl<T: VersionList + Send + Sync> VersionList for Compilers<T> {
-    async fn all_versions(&self) -> Vec<CompilerVersion> {
-        self.fetcher.all_versions().await
+impl<T: VersionList> VersionList for Compilers<T> {
+    fn all_versions(&self) -> Vec<CompilerVersion> {
+        self.fetcher.all_versions()
     }
 }
 
--- a/verification/src/compiler/compilers.rs
+++ b/verification/src/compiler/compilers.rs
@@ -1,5 +1,6 @@
 use crate::compiler::{CompilerVersion, DownloadCache, Fetcher, VersionList};
 use anyhow::anyhow;
+use async_trait::async_trait;
 use ethers_solc::{
     artifacts::{self, Severity},
     error::SolcError,
@@ -62,16 +63,17 @@ impl<T: Fetcher> Compilers<T> {
     }
 }
 
-impl<T: VersionList> VersionList for Compilers<T> {
-    fn all_versions(&self) -> Vec<&CompilerVersion> {
-        self.fetcher.all_versions()
+#[async_trait]
+impl<T: VersionList + Send + Sync> VersionList for Compilers<T> {
+    async fn all_versions(&self) -> Vec<CompilerVersion> {
+        self.fetcher.all_versions().await
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::solidity::{CompilerFetcher, Releases};
+    use crate::solidity::CompilerFetcher;
     use std::{env::temp_dir, str::FromStr};
 
     use crate::consts::DEFAULT_COMPILER_LIST;
@@ -84,10 +86,9 @@ mod tests {
         COMPILERS
             .get_or_init(async {
                 let url = DEFAULT_COMPILER_LIST.try_into().expect("Getting url");
-                let releases = Releases::fetch_from_url(&url)
+                let fetcher = CompilerFetcher::new(url, None, temp_dir())
                     .await
                     .expect("Fetch releases");
-                let fetcher = CompilerFetcher::new(releases, temp_dir()).await;
                 let compilers = Compilers::new(fetcher);
                 compilers
             })
--- a/verification/src/config.rs
+++ b/verification/src/config.rs
@@ -31,7 +31,7 @@ impl Default for ServerConfiguration {
 pub struct SolidityConfiguration {
     pub enabled: bool,
     pub compilers_list_url: Url,
-    pub cron_schedule: String,
+    pub refresh_versions_schedule: String,
 }
 
 impl Default for SolidityConfiguration {
@@ -39,7 +39,7 @@ impl Default for SolidityConfiguration {
         Self {
             compilers_list_url: Url::try_from(DEFAULT_COMPILER_LIST).expect("valid url"),
             enabled: true,
-            cron_schedule: "0 0 * * * * *".to_string(), // every hour
+            refresh_versions_schedule: "0 0 * * * * *".to_string(), // every hour
         }
     }
 }
--- a/verification/src/config.rs
+++ b/verification/src/config.rs
@@ -1,5 +1,6 @@
 use crate::consts::DEFAULT_COMPILER_LIST;
 use config::{Config as LibConfig, File};
+use cron::Schedule;
 use serde::Deserialize;
 use std::{net::SocketAddr, num::NonZeroUsize, path::PathBuf, str::FromStr};
 use url::Url;
@@ -31,7 +32,8 @@ impl Default for ServerConfiguration {
 pub struct SolidityConfiguration {
     pub enabled: bool,
     pub compilers_list_url: Url,
-    pub refresh_versions_schedule: String,
+    #[serde(with = "serde_with::rust::display_fromstr")]
+    pub refresh_versions_schedule: Schedule,
 }
 
 impl Default for SolidityConfiguration {
@@ -39,7 +41,7 @@ impl Default for SolidityConfiguration {
         Self {
             compilers_list_url: Url::try_from(DEFAULT_COMPILER_LIST).expect("valid url"),
             enabled: true,
-            refresh_versions_schedule: "0 0 * * * * *".to_string(), // every hour
+            refresh_versions_schedule: Schedule::from_str("0 0 * * * * *").unwrap(), // every hour
         }
     }
 }
--- a/verification/src/config.rs
+++ b/verification/src/config.rs
@@ -31,6 +31,7 @@ impl Default for ServerConfiguration {
 pub struct SolidityConfiguration {
     pub enabled: bool,
     pub compilers_list_url: Url,
+    pub cron_schedule: String,
 }
 
 impl Default for SolidityConfiguration {
@@ -38,6 +39,7 @@ impl Default for SolidityConfiguration {
         Self {
             compilers_list_url: Url::try_from(DEFAULT_COMPILER_LIST).expect("valid url"),
             enabled: true,
+            cron_schedule: "0 0 * * * * *".to_string(), // every hour
         }
     }
 }
--- a/verification/src/http_server/routers/solidity.rs
+++ b/verification/src/http_server/routers/solidity.rs
@@ -16,7 +16,7 @@ impl SolidityRouter {
     pub async fn new(config: SolidityConfiguration) -> anyhow::Result<Self> {
         let fetcher = CompilerFetcher::new(
             config.compilers_list_url.clone(),
-            Some(&config.refresh_versions_schedule),
+            Some(config.refresh_versions_schedule),
             "compilers/".into(),
         )
         .await?;
--- a/verification/src/http_server/routers/solidity.rs
+++ b/verification/src/http_server/routers/solidity.rs
@@ -15,7 +15,7 @@ pub struct SolidityRouter {
 impl SolidityRouter {
     pub async fn new(config: SolidityConfiguration) -> anyhow::Result<Self> {
         let fetcher = CompilerFetcher::new(
-            config.compilers_list_url.clone(),
+            config.compilers_list_url,
             Some(config.refresh_versions_schedule),
             "compilers/".into(),
         )
--- a/verification/src/http_server/routers/solidity.rs
+++ b/verification/src/http_server/routers/solidity.rs
@@ -5,7 +5,7 @@ use crate::{
     compiler::Compilers,
     config::SolidityConfiguration,
     http_server::handlers::{multi_part, standard_json, version_list},
-    solidity::{CompilerFetcher, Releases},
+    solidity::CompilerFetcher,
 };
 
 pub struct SolidityRouter {
@@ -14,10 +14,12 @@ pub struct SolidityRouter {
 
 impl SolidityRouter {
     pub async fn new(config: SolidityConfiguration) -> anyhow::Result<Self> {
-        let releases = Releases::fetch_from_url(&config.compilers_list_url)
-            .await
-            .map_err(anyhow::Error::msg)?;
-        let fetcher = CompilerFetcher::new(releases, "compilers/".into()).await;
+        let fetcher = CompilerFetcher::new(
+            config.compilers_list_url.clone(),
+            Some(&config.cron_schedule),
+            "compilers/".into(),
+        )
+        .await?;
         let compilers = Compilers::new(fetcher);
         Ok(Self {
             cache: web::Data::new(compilers),
--- a/verification/src/http_server/routers/solidity.rs
+++ b/verification/src/http_server/routers/solidity.rs
@@ -16,7 +16,7 @@ impl SolidityRouter {
     pub async fn new(config: SolidityConfiguration) -> anyhow::Result<Self> {
         let fetcher = CompilerFetcher::new(
             config.compilers_list_url.clone(),
-            Some(&config.cron_schedule),
+            Some(&config.refresh_versions_schedule),
             "compilers/".into(),
         )
         .await?;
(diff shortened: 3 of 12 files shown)


verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
     }
 }
 
+#[derive(Default)]
+struct RefreshableCompilerVersions {
```
leviska, line 112: I think this type isn't needed

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+    versions: Arc<RwLock<CompilerVersions>>,
```
leviska, line 113: we don't have any awaits inside locks, so make this `parking_lot::RwLock`
also may be include full name, do not `use parking_lot::RwLock`

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+                };
+            })
+        })?;
+        let scheduler = JobScheduler::new()?;
+        scheduler.add(job)?;
+        scheduler.start()
+    }
+
+    async fn refresh_versions(
+        versions: Arc<RwLock<CompilerVersions>>,
```
leviska, line 180: `&self` or `self`?

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+    fn spawn_refresh_job(
+        &self,
+        versions_list_url: Url,
+        cron_schedule: &str,
+    ) -> Result<(), JobSchedulerError> {
+        log::info!("spawn version refresh job with schedule {}", cron_schedule);
+        let versions = Arc::clone(&self.versions);
+
+        let job = Job::new_async(cron_schedule, move |_, _| {
+            let versions_list_url = versions_list_url.clone();
```
leviska, line 162: do you need to clone here?

verification/src/solidity/compiler_fetcher.rs
leviska, line 112: let's remove this and just include arc and mutex inside `CompilerVersions`

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -142,8 +228,8 @@ fn create_executable(path: &Path) -> Result<File, std::io::Error> {
 impl Fetcher for CompilerFetcher {
     type Error = FetchError;
     async fn fetch(&self, ver: &CompilerVersion) -> Result<PathBuf, Self::Error> {
-        let compiler_info = self
-            .releases
+        let compiler_versions = self.compiler_versions.versions.read().await;
```
leviska, line 231:
```rust
let compiler_info = {
  let compiler_versions = self.compiler_versions.versions.read().await;
  compiler_versions
            .0
            .get(ver)
            .ok_or_else(|| FetchError::NotFound(ver.clone()))?.clone()
}
```
this will heavilly reduce lock usage

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+        let fetched_versions = CompilerVersions::fetch_from_url(versions_list_url)
+            .await
+            .map_err(anyhow::Error::msg)?;
+        let need_to_update = {
+            let versions_lock = versions.read().await;
+            fetched_versions.0 != versions_lock.0
+        };
+        if need_to_update {
+            let mut versions_lock = versions.write().await;
+            log::debug!(
```
leviska, line 192: log::info

leviska, line 192: let's move the log out of the lock zone, something like
```rust
if need_to_update {
  let (old_len, new_len) = {
    let mut versions_lock = versions.write().await;
    let old_len = versions_lock.len();
    *versions_lock = fetched_versions;
    let new_len = versions_lock.len();
    (old_len, new_len)
  }
  log::debug!(
      "found new compiler versions. old length: {}, new length: {}",
      old_len,
      new_len,
  );
}
```

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+
+    async fn refresh_versions(
+        versions: Arc<RwLock<CompilerVersions>>,
+        versions_list_url: &Url,
+    ) -> anyhow::Result<()> {
+        let fetched_versions = CompilerVersions::fetch_from_url(versions_list_url)
+            .await
+            .map_err(anyhow::Error::msg)?;
+        let need_to_update = {
+            let versions_lock = versions.read().await;
```
leviska, line 187: about naming the variables with mutexes
usually it's something like
```rust
let variable_lock: Mutex<Type>
let variable_guard: MutexGuard<Type>
let variable: &Type
```
or you can just name the `variable_guard` as `variable`, if the name is free
let's rename them everywhere

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+        let scheduler = JobScheduler::new()?;
+        scheduler.add(job)?;
+        scheduler.start()
+    }
+
+    async fn refresh_versions(
+        versions: Arc<RwLock<CompilerVersions>>,
+        versions_list_url: &Url,
+    ) -> anyhow::Result<()> {
+        let fetched_versions = CompilerVersions::fetch_from_url(versions_list_url)
```
leviska, line 183: add `log::info("looking for new compilers versions")`

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+        if need_to_update {
+            let mut versions_lock = versions.write().await;
+            log::debug!(
+                "found new compiler versions. old length: {}, new length: {}",
+                versions_lock.0.len(),
+                fetched_versions.0.len()
+            );
+            *versions_lock = fetched_versions
+        } else {
+            log::debug!("no new versions found")
```
leviska, line 199: `log::info`

verification/src/compiler/compilers.rs
```diff
@@ -62,16 +63,17 @@ impl<T: Fetcher> Compilers<T> {
     }
 }
 
-impl<T: VersionList> VersionList for Compilers<T> {
-    fn all_versions(&self) -> Vec<&CompilerVersion> {
-        self.fetcher.all_versions()
+#[async_trait]
+impl<T: VersionList + Send + Sync> VersionList for Compilers<T> {
+    async fn all_versions(&self) -> Vec<CompilerVersion> {
```
leviska, line 68: if you replace the lock to `parking_lot`, you can replace back to the usual function

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +108,97 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+    ) -> anyhow::Result<()> {
+        let fetched_versions = CompilerVersions::fetch_from_url(versions_list_url)
+            .await
+            .map_err(anyhow::Error::msg)?;
+        let need_to_update = {
+            let versions_lock = versions.read().await;
+            fetched_versions.0 != versions_lock.0
+        };
+        if need_to_update {
+            let mut versions_lock = versions.write().await;
```
leviska, line 191: ok, so in general case, if you check something in read lock, then you need to check it again in write lock, because someone can take the lock between read and write, and then change the value and break the condition
for us, it's probably not needed, because even if two tasks trying to change the variable, it's hard to tell which to take, so we can leave it
but I really want to see the comment here about it, so something like
`// we're skipping the check again, because even if somebody changes the variable, there is no point to check it again`

leviska:
don't see any tests for the refresh part

verification/src/solidity/compiler_fetcher.rs
sevenzing: maybe do only 1 check after .write() lock then?

verification/src/solidity/compiler_fetcher.rs
sevenzing: Oh, didn't notice, sure

verification/src/solidity/compiler_fetcher.rs
sevenzing: Couldn't solve compiler errors without it(

verification/src/solidity/compiler_fetcher.rs
sevenzing: I wanted to do this but then I have to copy `self` to `tokio_cron_scheduler::Job` function and compiler ask `&self` to be `'static` in `spawn_refresh_job`

verification/src/solidity/compiler_fetcher.rs
sevenzing: then CompilerVersions::fetch_from_url will return some unnecessary arc and mutex.

in `refresh_versions` func I need new actual HashMap, I can only get it from `CompilerVersions::fetch_from_url` but it returns hashmap with mutex, which I don't need

verification/src/solidity/compiler_fetcher.rs
sevenzing: I wanted to do this but then I have to copy `self` to `tokio_cron_scheduler::Job` function and compiler ask `&self` to be `'static` in `spawn_refresh_job`

verification/src/solidity/compiler_fetcher.rs
leviska: You mean do it without read lock? It will be worse in performance

verification/src/config.rs
```diff
@@ -31,13 +31,15 @@ impl Default for ServerConfiguration {
 pub struct SolidityConfiguration {
     pub enabled: bool,
     pub compilers_list_url: Url,
+    pub cron_schedule: String,
```
rimrakhimov, line 34: Wouldn't be more indicative if we name it compilers_refresh_schedule?

rimrakhimov: Also, do we want to allow users specify that no refreshing is required at all?

verification/src/solidity/mod.rs
```diff
@@ -2,4 +2,4 @@ mod compiler_fetcher;
 pub mod svm_fetcher;
 mod verifier;
 
-pub use compiler_fetcher::{CompilerFetcher, Releases};
+pub use compiler_fetcher::{CompilerFetcher, CompilerVersions};
```
rimrakhimov, line 5:
```suggestion
pub use compiler_fetcher::CompilerFetcher;
```

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +107,99 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+                };
+            })
+        })?;
+        let scheduler = JobScheduler::new()?;
+        scheduler.add(job)?;
+        scheduler.start()
+    }
+
+    async fn refresh_versions(
+        versions: Arc<parking_lot::RwLock<CompilerVersions>>,
```
rimrakhimov, line 147: Are you sure that using `parking_lot::RwLock` and not `tokio::sync::RwLock` cannot cause deadlocks?

verification/src/config.rs
sevenzing: how? it's not possible to make this field Option<String>, do you suggest to introduce new field like `compilers_refresh_enabled` ?

verification/src/solidity/compiler_fetcher.rs
sevenzing: decided to move logic of `CompilerVersions::fetch_from_url` to separate function `try_fetch_versions`

verification/src/scheduler.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+fn time_till_next_call(schedule: &Schedule) -> std::time::Duration {
+    let default = std::time::Duration::from_millis(500);
+    let now = Utc::now();
+
+    schedule
+        .upcoming(chrono::Utc)
+        .take(1)
+        .next()
```
rimrakhimov, line 32:
```suggestion
        .next()
```

rimrakhimov (approved):

verification/src/scheduler.rs
```diff
@@ -0,0 +1,61 @@
...
+) -> tokio::task::JoinHandle<()>
+where
+    F: (FnMut() -> Fut) + Send + 'static,
+    Fut: Future + Send + 'static,
+    <Fut as futures::Future>::Output: Send,
+{
+    tokio::spawn(async move {
+        loop {
+            let sleep_duration = time_till_next_call(&schedule);
+            log::debug!("sleep for {:?}", sleep_duration);
```
leviska, line 18: `scheduled next compiler list update in {:?}`

verification/src/scheduler.rs
```diff
@@ -0,0 +1,60 @@
+use cron::Schedule;
+use futures::Future;
+
+use chrono::Utc;
+
+pub fn spawn_job_with_schedule<F, Fut>(
```
leviska, line 6: may be just `scheduler::spawn_job`?
also you already take schedule as argument, so it looks like
`spawn_job_with_schedule(schedule: Scheduler...)`
vs
`spawn_job(schedule: Scheduler...) // spawn_job with (schedule,... )`

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -104,15 +116,75 @@ impl TryFrom<(json::CompilerInfo, &Url)> for CompilerInfo {
...
+        let fetched_versions = try_fetch_versions(versions_list_url)
+            .await
+            .map_err(anyhow::Error::msg)?;
+        let need_to_update = {
+            let versions = self.0.read();
+            fetched_versions != *versions
+        };
+        if need_to_update {
+            let (old_len, new_len) = {
+                let mut versions = self.0.write();
```
leviska, line 145: where is mah comment
something something `// we don't need to check condition again, we can just override the value`

verification/src/http_server/routers/solidity.rs
```diff
@@ -14,10 +14,12 @@ pub struct SolidityRouter {
 
 impl SolidityRouter {
     pub async fn new(config: SolidityConfiguration) -> anyhow::Result<Self> {
-        let releases = Releases::fetch_from_url(&config.compilers_list_url)
-            .await
-            .map_err(anyhow::Error::msg)?;
-        let fetcher = CompilerFetcher::new(releases, "compilers/".into()).await;
+        let fetcher = CompilerFetcher::new(
+            config.compilers_list_url.clone(),
```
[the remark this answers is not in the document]
sevenzing: we need to clone it somewhere before passing to refresh job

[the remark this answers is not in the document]
sevenzing: hm, maybe you are right and it's better to clone in CompilerFetcher::new

verification/src/scheduler.rs
sevenzing: this function doesn't know about compiler list. I thought about passing name of job as function argument (&str) and print it in logs, what do you think?

verification/src/solidity/compiler_fetcher.rs
```diff
@@ -50,40 +57,45 @@ pub enum ListError {
...
 }
 
-#[derive(Default)]
-pub struct Releases(HashMap<CompilerVersion, CompilerInfo>);
+type CompilerVersionsMap = HashMap<CompilerVersion, CompilerInfo>;
+
+#[derive(Default, Clone)]
+pub struct CompilerVersions(Arc<parking_lot::RwLock<CompilerVersionsMap>>);
+
+impl From<CompilerVersionsMap> for CompilerVersions {
```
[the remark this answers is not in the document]
sevenzing: agree to remove From, but CompilerVersionsMap occur 3 times...

verification/src/scheduler.rs
rimrakhimov: @leviska It is not a problem here, but in general as Rust does not have function overloading that approach may not be valid. If another spawning method were there which would accept something else instead of `Schedule` (e.g. `&str`) we could not create the method `spawn_job(_: &str)`

verification/src/solidity/compiler_fetcher.rs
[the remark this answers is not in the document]
rimrakhimov: I believe using `RwLock` is better here, as that lock would be taken for every `all_versions` and `fetch` requests. Using `Mutex` may become a source of bottleneck and may open a vulnerability allowing to slow down the service by sending a usual `all_versions` requests

verification/src/scheduler.rs
rimrakhimov: > I thought about passing name of job as function argument (&str) and print it in logs, what do you think?

I believe it may be a good idea

verification/src/http_server/routers/solidity.rs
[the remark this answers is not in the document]
sevenzing: right, I see

sevenzing:
🥳

PR #52, opened 2022-06-30
Integrate compilation and verification

rimrakhimov:
Continues #27.

Combines code compilation and its verification. Add tests for the whole process


sevenzing:
I don't like name for `handlers.rs`. It's not a handler. (in out code style hander is function that passes to actix) 

maybe change filename and add word `bruteforce` to `compile_and_verify_handler` function? like `compile_and_verify_bruteforce`, or something else

verification/src/http_server/handlers/verification/solidity/handlers.rs
```diff
@@ -0,0 +1,106 @@
...
+                    input.compiler_input,
+                    input.compiler_version,
+                    verification_success,
+                ));
+                return Ok(VerificationResponse::ok(verification_result));
+            }
+            Err(CompileAndVerifyError::VerifierInitialization(err)) => {
+                return Err(error::ErrorBadRequest(err))
+            }
+            Err(CompileAndVerifyError::Compilation(CompilersError::Fetch(err))) => {
```
sevenzing, line 67: I don't like idea of `Err(...)` here. I think that user input errors should be returned as `VerificationResponse::err`.

`Err(...)` should be use in case of very strange error that shouldn't happen (like invalid bytecode, or invalid compiler versions)

verification/tests/solidity_multiple.rs
```diff
@@ -0,0 +1,435 @@
...
+        response.status().is_success(),
+        "Invalid status code (success expected): {}",
+        response.status()
+    );
+
+    let verification_response: VerificationResponse = read_body_json(response).await;
+
+    assert_eq!(
+        verification_response.status,
+        VerificationStatus::Ok,
```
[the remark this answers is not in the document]
rimrakhimov: It is not the final version, and I'll  add negative cases later today

rimrakhimov:
> I don't like name for handlers.rs. It's not a handler. (in out code style hander is function that passes to actix)
>
> maybe change filename and add word bruteforce to compile_and_verify_handler function? like compile_and_verify_bruteforce, or something else

verification/src/http_server/handlers/verification/solidity/handlers.rs
rimrakhimov: There are different kind of errors that may occur here. `CompilersError::Fetch` may be returned in two cases: 1) if user supplied invalid compiler version which is not in our list of all versions, and 2) if user provided valid version but downloading for some reason failed (e.g. link for download is invalid). In case (2) we certainly should return `InternalError`. In case (1) however (I believe that is the case you've meant), I believe we should return not 200 with `VerificationResponse::err` but  `400 BadRequest`. It is because we provide the user with a list of compilers we are ready to accept, and if they still provide us something invalid should be considered as their fault, I believe.

verification/src/http_server/handlers/verification/solidity/multi_part.rs
```diff
@@ -18,13 +20,16 @@ pub async fn verify(
 ) -> Result<Json<VerificationResponse>, Error> {
     let params = params.into_inner();
 
-    let input = CompilerInput::try_from(params.content).map_err(error::ErrorBadRequest)?;
+    let compiler_input = params.content.try_into().map_err(error::ErrorBadRequest)?;
```
sevenzing, line 23: why not `CompilerInput::try_from`? I thought `From` is more clear than `Into`

verification/tests/solidity_multiple.rs
```diff
@@ -0,0 +1,389 @@
...
+    async fn contract_from_factory() {
+        let contract_dir = "contract_from_factory";
+        let test_input = TestInput::new("ContractFromFactory", "v0.4.26+commit.4563c3fc")
+            .with_optimization_runs(200)
+            .has_constructor_args();
+        test_success(contract_dir, test_input).await;
+    }
+}
+
+mod basic_error_tests {
```
sevenzing, line 269: I would also add test with wrong compiler version, but it's not super important right now

sevenzing (approved):

verification/src/http_server/handlers/verification/solidity/handlers.rs
```diff
@@ -0,0 +1,106 @@
...
+    let compiler_output = compilers
+        .compile(&input.compiler_version, &input.compiler_input)
+        .await?;
+    verifier
+        .verify(compiler_output)
+        .ok_or(CompileAndVerifyError::NoMatchingContracts)
+}
+
+fn pre_v0_6_0(compiler_version: &CompilerVersion) -> bool {
+    let last_non_0_6_0_version =
```
leviska, line 103: we can just compare semvers, and also semvers implement constrains, so you can do something like
```rust
let req = VersionReq::parse("<0.6.0").unwrap();
if req.matches(compiler_version.version()) {
```
also I don't like the name of the function and suggest just remove the function and do the versionreq variant

verification/src/http_server/handlers/verification/solidity/handlers.rs
```diff
@@ -0,0 +1,106 @@
...
+                return Err(error::ErrorBadRequest(err))
+            }
+            Err(CompileAndVerifyError::Compilation(CompilersError::Fetch(err))) => {
+                return Err(error::ErrorInternalServerError(err))
+            }
+            Err(CompileAndVerifyError::Compilation(CompilersError::Internal(err))) => {
+                return Err(error::ErrorInternalServerError(err))
+            }
+            err @ Err(CompileAndVerifyError::Compilation(CompilersError::Compilation(_))) => {
+                return Ok(VerificationResponse::err(err.expect_err("Matched to Err")))
```
leviska, line 74: let's remove useless `expect`s. Document the expect thoughtfully, or just put `unwrap()/unwrap_err()`.
Writing `expect("expected value")` doesn't give any new information relative to the `unwrap()`

verification/src/http_server/handlers/verification/solidity/handlers.rs
```diff
@@ -0,0 +1,106 @@
...
+        Vec::from([input.compiler_input.settings.metadata.clone()])
+    } else if pre_v0_6_0(&input.compiler_version) {
+        Vec::from([None])
+    } else {
+        Vec::from(BYTECODE_HASHES.map(|hash| Some(SettingsMetadata::from(hash))))
+    };
+
+    for metadata in bruteforce_metadata {
+        input.compiler_input.settings.metadata = metadata;
+        match compile_and_verify(compilers, &input).await {
```
leviska, line 55: can't we implement `from CompileAndVerifyError into actix_web::Error` and do something like
```rust
match compile_and_verify(compilers, &input).await {
  Ok(..)..,
  Err(NoMatchingContracts) => {},
  Err(err) => err.into(),
}
```

verification/src/http_server/handlers/verification/solidity/handlers.rs
```diff
@@ -0,0 +1,106 @@
+use crate::{
```
leviska, line 1: rename this to `compile_and_verify` and remove this prefix from types, so we could use it like `compile_and_verify::Input` and etc.

also I'd like to rename it to something else, may be just `verify` or `verify_full` or something, but my ideas are worse, than current

verification/src/http_server/handlers/verification/solidity/handlers.rs
```diff
@@ -0,0 +1,106 @@
...
+            Err(CompileAndVerifyError::VerifierInitialization(err)) => {
+                return Err(error::ErrorBadRequest(err))
+            }
+            Err(CompileAndVerifyError::Compilation(CompilersError::Fetch(err))) => {
+                return Err(error::ErrorInternalServerError(err))
+            }
+            Err(CompileAndVerifyError::Compilation(CompilersError::Internal(err))) => {
+                return Err(error::ErrorInternalServerError(err))
+            }
+            err @ Err(CompileAndVerifyError::Compilation(CompilersError::Compilation(_))) => {
```
leviska, line 73: can't we use this at first and then do just 
```rust
Err(CompileAndVerifyError::Compilation(err)) => }
return Err(error::ErrorInternalServerError(err))
}
```

verification/src/http_server/handlers/verification/solidity/handlers.rs
```diff
@@ -0,0 +1,106 @@
...
+enum CompileAndVerifyError {
+    #[error("{0:#}")]
+    VerifierInitialization(#[from] VerifierInitializationError),
+    #[error("{0:#}")]
+    Compilation(#[from] CompilersError),
+    #[error("No contract could be verified with provided data")]
+    NoMatchingContracts,
+}
+
+pub(crate) async fn compile_and_verify_handler<T: Fetcher>(
```
leviska, line 37: just a random idea, may be make this (and other functions) as `impl CompileAndVerifyInput`?

verification/tests/solidity_multiple.rs
```diff
@@ -0,0 +1,398 @@
...
+    #[actix_rt::test]
+    async fn issue_4758() {
+        let contract_dir = "issue_4758";
+        let test_input = TestInput::new("CS3_1OnChainShop", "v0.8.4+commit.c7e474f2")
+            .with_optimization_runs(200)
+            .has_constructor_args();
+        test_success(contract_dir, test_input).await;
+    }
+
+    // runs only for linux, as other compiler lists do not have nightly builds
```
leviska, line 351: right now nightly builds will be available only in s3, for which other developers don't have access

I have an idea to push all the nightly builds, which are used in tests in solc-bin releases, so we could use them without access to s3, we can discuss this

verification/tests/solidity_multiple.rs
```diff
@@ -0,0 +1,398 @@
...
+    async fn contract_from_factory() {
+        let contract_dir = "contract_from_factory";
+        let test_input = TestInput::new("ContractFromFactory", "v0.4.26+commit.4563c3fc")
+            .with_optimization_runs(200)
+            .has_constructor_args();
+        test_success(contract_dir, test_input).await;
+    }
+}
+
+mod basic_error_tests {
```
leviska, line 278: drop the `basic`?

verification/tests/solidity_multiple.rs
```diff
@@ -0,0 +1,398 @@
...
+    );
+
+    assert!(
+        verification_response.message.contains(expected_message),
+        "Invalid message: {}",
+        verification_response.message
+    );
+}
+
+mod basic_tests {
```
leviska, line 203: drop the `basic`?

verification/src/http_server/handlers/verification/solidity/handlers.rs
leviska: about naming
ok, basically this module verifies the contracts, but the main problem is that we're already have `verify` module or something
what's the difference then? you can say that this module compiles the code, but I think it is more like a implementation detail, and not the api description
instead, this module verifies the _source code_, and the other compares _bytecodes_
renaming that module is optional, but using this logic we could name this just `verify_source` or something
and then we'd say, that how exactly we transform source code into bytecode is impl detail

verification/src/http_server/handlers/verification/solidity/handlers.rs
rimrakhimov: Okay, agree that using `unwrap_err` the reason why it should not fail is quite obvious in that case

verification/src/http_server/handlers/verification/solidity/handlers.rs
rimrakhimov: In that case we would have to put some `actix_web::Error` in correspondence to `CompileAndVerifyError::NoMatchingContracts` which actually does not correspond to any actix_web error. So, I prefer to make this conversion explicit

sevenzing (approved):

verification/tests/solidity_multiple_types.rs
```diff
@@ -0,0 +1,85 @@
...
+            contract_libraries: Default::default(),
+            has_constructor_args: false,
+
+            source_code: None,
+            creation_tx_input: None,
+            deployed_bytecode: None,
+        }
+    }
+
+    pub fn with_evm_version(self, evm_version: &'static str) -> Self {
```
leviska, line 37: I understand that this is a code for test, and it can be not fully optimized, but still, passing self by reference removes all the copying/moving, while this variant still moves each variable
https://godbolt.org/z/1oqzjhfWj
https://godbolt.org/z/c7h6WeKEj
So I suggest changing this to reference (everywhere)

Also I really don't understand why
```rust
let input = TestInput::new(..).with_evm_version(something);
```
is better, than
```rust
let input = TestInput::new(..);
input.evm_version = something;
```
or
```rust
let input = TestInput{
  evm_version,
  optimization_runs: 200,
  ..TestInput::new(..),
}
```

verification/src/http_server/handlers/verification/solidity/contract_verifier.rs
```diff
@@ -0,0 +1,105 @@
...
+
+pub struct Input<'a> {
+    pub compiler_version: CompilerVersion,
+    pub compiler_input: CompilerInput,
+    pub creation_tx_input: &'a str,
+    pub deployed_bytecode: &'a str,
+}
+
+#[derive(Error, Debug)]
+enum CompileAndVerifyError {
```
leviska, line 26: ContractVerifyError?

verification/src/http_server/handlers/verification/solidity/contract_verifier.rs
```diff
@@ -0,0 +1,105 @@
...
+            Err(CompileAndVerifyError::NoMatchingContracts) => {}
+        }
+    }
+    // In case of any other error the execution will not get to this point
+    Ok(VerificationResponse::err(
+        CompileAndVerifyError::NoMatchingContracts,
+    ))
+}
+
+async fn compile_and_verify<T: Fetcher>(
```
leviska, line 73: `verify_single_contract`?

verification/src/http_server/handlers/verification/solidity/contract_verifier.rs
```diff
@@ -0,0 +1,105 @@
...
+{
+    let compiler_output = compilers
+        .compile(&input.compiler_version, &input.compiler_input)
+        .await?;
+    verifier
+        .verify(compiler_output)
+        .ok_or(CompileAndVerifyError::NoMatchingContracts)
+}
+
+fn settings_metadata(
```
leviska, line 89: can you comment what this function does? something very deep connected to the contracts it seems, but the comment explaining at least something (and/or may be an url where we could read more about settings metadata or smth) would really help

leviska (approved):

verification/tests/solidity_multiple_types.rs
rimrakhimov: Passing self by value allows us to create a chain of calls which initiates the structure instead of making a separate statement for each field. Likewise, using setters on construction allows us to hide some implementation details not important for the initialization (e.g. we pass a `String` as a deployed_bytecode in `TestInput::with_deployed_bytecode`, while internally the method sets corresponding field to be `Some(deployed_bytecode)`. In my opinion, altogether they make the initialization logic more explicit and readable. 

Regarding optimization I believe clearness is more preferable here in comparison with performance gain which we get by using references. However, I've updated the code so that it has only one more move operation in comparison with using references while leaving the user experience the same (https://godbolt.org/z/MMnn8bTzq)

verification/src/http_server/handlers/verification/solidity/contract_verifier.rs
rimrakhimov: The main difference between `compile_and_verify_handler` and `compile_and_verify` is that the former converts errors returned during verification into `web_actix::Error` that can be returned to the user. That is the main difference and not the fact that `compile_and_verify_handler` may bruteforce between different bytecode hashes or anything else. It is just a byproduct of the fact that we need a `CompileAndVerifyError` to distinguish whether the error may be related to incorrect bytecode hash or the verification will result in that error for any hash. 

`compile_and_verify` is just a function that makes it convenient to return `CompileAndVerifyError` which may be further converted into `actix_web::Error`

verification/src/http_server/handlers/verification/solidity/contract_verifier.rs
```diff
@@ -0,0 +1,105 @@
...
+) -> Result<VerificationSuccess, CompileAndVerifyError>
+where
+    <T as Fetcher>::Error: Debug + Display,
+{
+    let compiler_output = compilers
+        .compile(&input.compiler_version, &input.compiler_input)
+        .await?;
+    verifier
+        .verify(compiler_output)
+        .ok_or(CompileAndVerifyError::NoMatchingContracts)
```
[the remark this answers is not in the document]
rimrakhimov: I do not think we need it right now. For now there is no need to return the reason why bytecode verification failed, as it does not matter for the client. There is only a fact that no contract corresponds to the provided deployed bytecode and creation tx input that is important for the client.

In the future we may be interested in parsing some of the errors occurring during verification (e.g. a contract library may be missing), but right now we are not doing that

PR #54, opened 2022-07-04
Add POST-requests handler to multichain-api-gateway

kuksag:
This lets us make POST requests to our API (e.g. [contract verify request](https://blockscout.com/eth/mainnet/api-docs#:~:text=More%20Details-,Verify,-a%20contract%20with%20its)).
It's done in the way that the incoming request body is passed through to the following calls.
Added 1 integration test that makes a call as described in an example [here](https://blockscout.com/eth/mainnet/api-docs#:~:text=More%20Details-,Verify,-a%20contract%20with%20its).


multichain-api-gateway/src/lib.rs
```diff
@@ -45,13 +46,31 @@ impl From<BlockscoutSettings> for ApiEndpoints {
...
+                    .header(
+                        "Content-Type",
+                        headers
+                            .get("Content-Type")
+                            .unwrap_or(&"application/json".parse().unwrap()),
+                    )
+                    .body(body.clone())
+                    .send()
+                    .await
+                    .unwrap();
```
sevenzing, line 73: usually unwrap is not best practice for http server. maybe return Result<> from this function and handle error somewhere above?

multichain-api-gateway/src/lib.rs
```diff
@@ -45,13 +46,31 @@ impl From<BlockscoutSettings> for ApiEndpoints {
...
             .map(|(instance, mut url)| async {
                 url.set_query(Some(query));
-                let resp = client.get(url).send().await.unwrap();
+
+                let resp = client
+                    .request(method.clone(), url)
+                    .header(
+                        "Content-Type",
+                        headers
+                            .get("Content-Type")
```
sevenzing, line 67: hm, but what if I would like to pass other headers? Like `Authorization: <auth_token>`

I see that `headers()` accepts `http::HeaderMap`, however we have `actix_web::http::header::HeaderMap`, it's uncomfortable(((
but maybe we have convert one type to another? what do you think? 👀

multichain-api-gateway/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 actix-web = "4"
+awc = { version = "3.0.0", features = ["openssl"] }
```
kuksag, line 8: Somehow default awc can't work with ssl
https://github.com/actix/actix-web/issues/1045

multichain-api-gateway/src/lib.rs
```diff
@@ -2,12 +2,13 @@ use std::{collections::HashMap, net::TcpListener, str};
 
 use actix_web::{
     dev::Server,
+    http::{header::HeaderMap, Method},
     web,
-    web::{Data, Json},
+    web::{Bytes, Data, Json},
     App, HttpRequest, HttpServer, Responder,
 };
+use awc::Client;
```
kuksag, line 10: I decided to try replace "reqwest" client with actix-web client (awc) since we use actix-web server. It takes almost the same lines of code.
Let me know what you think; it's easy to rollback to "reqwest" if needed.

multichain-api-gateway/src/lib.rs
```diff
@@ -45,58 +46,82 @@ impl From<BlockscoutSettings> for ApiEndpoints {
...
-        let client = Client::new();
+    async fn make_requests(
+        self,
+        query: &str,
+        method: &Method,
+        body: Bytes,
+        headers: &HeaderMap,
+    ) -> Vec<(Instance, String)> {
+        let client = Client::builder()
+            .timeout(std::time::Duration::from_secs(60))
```
kuksag, line 57: Default value for a timeout is 1 second. 
[This](https://blockscout.com/eth/mainnet/api-docs#:~:text=curl%20%2Dd%20%27%7B%22addressHash,contract%26action%3Dverify%22) request takes 48 second on my pc

multichain-api-gateway/src/lib.rs
```diff
@@ -45,58 +46,82 @@ impl From<BlockscoutSettings> for ApiEndpoints {
...
+                let str_response = match response {
+                    Ok(mut response) => match response.body().await {
+                        Ok(bytes) => match str::from_utf8(bytes.as_ref()) {
+                            Ok(str) => str.to_string(),
+                            Err(utf8_error) => utf8_error.to_string(),
+                        },
+                        Err(payload_error) => payload_error.to_string(),
+                    },
+                    Err(send_request_error) => send_request_error.to_string(),
+                };
```
kuksag, line 81: I think the whole block is looking kind of messy, but I don't know what to do with it, since:
- `response.body().await` is needed to be called outside of any `map`
- chain computation via `and_then` can't be used because we have different result and error types on every step

multichain-api-gateway/src/lib.rs
```diff
@@ -45,58 +46,82 @@ impl From<BlockscoutSettings> for ApiEndpoints {
...
+                let request_headers = request.headers_mut();
+                for (key, value) in headers {
+                    request_headers.insert(key.clone(), value.clone());
+                }
+                let response = request.send_body(body.clone()).await;
+                let str_response = match response {
+                    Ok(mut response) => match response.body().await {
+                        Ok(bytes) => match str::from_utf8(bytes.as_ref()) {
+                            Ok(str) => str.to_string(),
+                            Err(utf8_error) => utf8_error.to_string(),
```
kuksag, line 76: We can use `unwrap_or_else` here, but I'm not sure if it will add more readability

multichain-api-gateway/src/lib.rs
```diff
@@ -45,58 +46,82 @@ impl From<BlockscoutSettings> for ApiEndpoints {
...
-    json_resonses
+    json_responses
         .into_iter()
         .for_each(|(Instance(net, subnet), value)| {
             let kv_subnet = result.entry(net).or_insert_with(HashMap::new);
             kv_subnet.insert(
                 subnet,
-                serde_json::from_str(&value).unwrap_or_else(|e| Value::String(e.to_string())),
+                serde_json::from_str(&value)
+                    .unwrap_or_else(|e| Value::String(format!("{}: {}\n", e, value))),
```
kuksag, line 102: Before that, when the server returned an error, the error was replaced with`expected value at column 1 at row 1` (json parsing failed). Now we preserve both errors

multichain-api-gateway/src/lib.rs
rimrakhimov: Personally, in my opinion, it would be more readable. You would just have to map success result into String before unwrapping

multichain-api-gateway/src/lib.rs
rimrakhimov: I suggest to create a new private method 
```
async fn make_request(request: ClientRequest, body: Bytes) -> Result<String, Box<dyn Error>>
```
inside an `ApiRequest` and use `?` to convert between errors there

PR #59, opened 2022-07-11
preparing to implement s3 fetcher

leviska:
renamed compiler fetcher into list fetcher
moved list fetcher into compilers
made traits use dyn Fetcher, instead just the trait


verification/src/compiler/compilers.rs
```diff
@@ -14,13 +14,13 @@ pub enum CompilersError {
     Compilation(Vec<String>),
 }
 
-pub struct Compilers<T> {
+pub struct Compilers<T: ?Sized> {
     cache: DownloadCache,
-    fetcher: T,
+    fetcher: Arc<T>,
```
rimrakhimov, line 19: Imo, it would be better to drop generics here and use `Arc<dyn Fetcher>`

verification/src/compiler/fetcher.rs
```diff
@@ -1,13 +1,22 @@
...
+    Shedule(tokio::task::JoinError),
 }
 
-pub trait VersionList {
-    fn all_versions(&self) -> Vec<CompilerVersion>;
+#[async_trait]
+pub trait Fetcher: Send + Sync + 'static {
+    async fn fetch(&self, ver: &Version) -> Result<PathBuf, FetchError>;
+    fn all_versions(&self) -> Vec<Version>;
 }
```
rimrakhimov, line 22: What is the reason behind removing Error as associated type?

rimrakhimov (commented):
I suggest to create a single public error enum implemented for the whole compiler module which combines all types of compiler errors (`CompilersError`, `FetchError`), put that error in the `mod.rs` and return it instead in the module. That way we will get rid of hiding `FetchError`s inside `CompilersError` and may expose just a single `compiler::Error` type as part of module public interface. That may allow to reduce number of abstractions over the module and make the code more straightforward and easier to read.

verification/src/compiler/fetcher.rs
leviska: Because then we'll still use `dyn Fetcher<Error=FetchErrror>`, and if we're using that, why not just include it into the trait

verification/src/compiler/fetcher.rs
```diff
@@ -1,13 +1,22 @@
...
+#[derive(Error, Debug)]
+pub enum FetchError {
+    #[error("version {0} not found")]
+    NotFound(Version),
+    #[error("couldn't fetch the file: {0}")]
+    Fetch(anyhow::Error),
+    #[error("couldn't create file: {0}")]
+    File(std::io::Error),
+    #[error("tokio sheduling error: {0}")]
+    Shedule(tokio::task::JoinError),
```
rimrakhimov, line 15:
```suggestion
    Schedule(tokio::task::JoinError),
```

verification/src/compiler/fetcher.rs
```diff
@@ -1,13 +1,22 @@
...
 }
 
-pub trait VersionList {
-    fn all_versions(&self) -> Vec<CompilerVersion>;
+#[async_trait]
+pub trait Fetcher: Send + Sync + 'static {
```
rimrakhimov, line 19: Seems that `'static` bound is not needed

verification/src/compiler/compilers.rs
```diff
@@ -1,26 +1,26 @@
...
+use std::{fmt::Debug, sync::Arc};
+use thiserror::Error as DeriveError;
 
-#[derive(Debug, Error)]
-pub enum CompilersError {
+#[derive(Debug, DeriveError)]
+pub enum Error {
     #[error("Error while fetching compiler: {0:#}")]
-    Fetch(anyhow::Error),
+    Fetch(FetchError),
```
rimrakhimov, line 10:
```suggestion
    Fetch(#[from] FetchError),
```
Allows to get rid of `map_err` and use only `?`

rimrakhimov (approved):

verification/src/compiler/compilers.rs
leviska: why we're not using it everywhere then? :thinking:

sevenzing: good question :) 
I didn't write it somewhere because if you had 2 similar #[from] errors it wouldn't work. 
but yes, we should use it when it's possible

sevenzing (approved):

leviska:
we've moved verification into it's own repo, you can see this pr here https://github.com/blockscout/verification/pull/1

PR #65, opened 2022-07-28
Add metrics

sevenzing:


leviska (approved):

verification/example_config.toml
```diff
@@ -9,3 +9,7 @@ request_timeout = 10
 [solidity]
 compilers_list_url = "https://raw.githubusercontent.com/blockscout/solc-bin/main/list.json"
 refresh_versions_schedule = "0 0 * * * * *"
+
+[metrics]
```
leviska, line 13: May be add "enabled" option, so people could disable metrics altogether (or even disable them by default)

verification/src/http_server/metrics.rs
```diff
@@ -0,0 +1,102 @@
...
+pub struct Metrics {
+    endpoint: PrometheusMetrics,
+    middleware: PrometheusMetrics,
+}
+
+impl Metrics {
+    pub fn new(endpoint: String) -> Self {
+        let shared_registry = build_registry();
+
+        let endpoint = PrometheusMetricsBuilder::new("verification_metrics_endpoint")
```
rimrakhimov, line 75:
```suggestion
        let endpoint = PrometheusMetricsBuilder::new("verification_metrics")
```
I believe `endpoint` suffix is not required

verification/src/http_server/metrics.rs
```diff
@@ -0,0 +1,102 @@
...
+            .registry(shared_registry.clone())
+            .endpoint(&endpoint)
+            .build()
+            .unwrap();
+        let middleware = PrometheusMetricsBuilder::new("verification")
+            .registry(shared_registry)
+            .build()
+            .unwrap();
```
rimrakhimov, line 83: I suggest to name those variables as metrics_middleware and verification_middleware rather than endpoint and middleware. The main idea here is that they are both a middlewares in the sense of corresponding HTTP servers, but one of them belongs to metrics server and another to the actual verification server

verification/example_config.toml
rimrakhimov: I agree that it would be better to disable them by default

sevenzing: I think it should be enabled by default, as we did with solidity and sourcify 
image

rimrakhimov: It is not the main logic of the service and I believe users will not want to have it opened by default. So, I believe auto-enabling of `solidity` and `sourcify` should not be the reason to enable metrics by default

sevenzing: ok, changed to `enabled: false` in default

verification/src/http_server/handlers/verification/solidity/standard_json.rs
```diff
@@ -31,7 +34,7 @@ pub async fn verify(
         creation_tx_input: &params.creation_bytecode,
         deployed_bytecode: &params.deployed_bytecode,
     };
-    compile_and_verify_handler(&compilers, input, false)
-        .await
-        .map(Json)
+    let response = compile_and_verify_handler(&compilers, input, true).await?;
```
rimrakhimov, line 37: Why it became `true` while being `false` before?

verification/src/http_server/metrics.rs
```diff
@@ -0,0 +1,91 @@
...
+use prometheus::{
+    register_histogram_with_registry, register_int_counter_vec_with_registry,
+    register_int_counter_with_registry, Histogram, IntCounter, IntCounterVec, Registry,
+};
+use std::net::SocketAddr;
+
+use crate::VerificationStatus;
+
+lazy_static! {
+    pub static ref REGISTERY: Registry = Registry::new();
```
rimrakhimov, line 13:
```suggestion
    pub static ref REGISTRY: Registry = Registry::new();
```

rimrakhimov (approved):

verification/src/http_server/handlers/verification/solidity/standard_json.rs
sevenzing: oops, copypaste

verification/src/http_server/mod.rs
```diff
@@ -1,22 +1,50 @@
...
+    let metrics_endpoint = config.metrics.endpoint.clone();
+
     log::info!("Verification server is starting at {}", socket_addr);
     let app_router = Arc::new(
         AppRouter::new(config)
             .await
             .expect("couldn't initialize the app"),
     );
-    HttpServer::new(move || App::new().configure(configure_router(&*app_router)))
+    let metrics = Metrics::new(metrics_endpoint);
```
leviska, line 26: can we just not create this if `metrics_enabled == false` ?

verification/src/http_server/mod.rs
```diff
@@ -1,22 +1,50 @@
...
+                .wrap(middleware.clone())
+                .configure(configure_router(&*app_router))
+        })
         .bind(socket_addr)?
         .run()
-        .await
+    };
+    let server_future = tokio::spawn(async move { server_future.await });
+    let metrics_future = tokio::spawn(async move {
+        if metrics_enabled {
```
leviska, line 39: I guess it's easier to check if here, but I'd prefer to actually do nothing, if metrics not enabled, including not spawning a thread. You can do this using vec
```rust
let futures = vec![tokio::spawn(server_future...)];
if metrics_enabled {
  futures.push(tokio::spawn(metrics...));
}
```

verification/tests/server.rs
```diff
@@ -0,0 +1,43 @@
...
+    config.metrics.enabled = true;
+    let base = format!("http://{}", config.server.addr);
+    let metrics_base = format!("http://{}", config.metrics.addr);
+    let _server_handle = {
+        let config = config.clone();
+        tokio::spawn(async move { run_http_server(config).await })
+    };
+
+    let sleep_between = Some(tokio::time::Duration::from_millis(100));
+    let attempts = NonZeroUsize::new(20).unwrap();
```
leviska, line 20: let's make this high (`100` == 10 secs), so it'll be like a timeout, and not some arbitrary "well this probably takes something around 2 secs on my machine may be" :)

verification/src/lib.rs
```diff
@@ -17,3 +18,4 @@ pub use http_server::{
     handlers::verification::{VerificationResponse, VerificationResult, VerificationStatus},
     run as run_http_server, AppRouter, Router,
 };
+pub use network::make_retrying_request;
```
leviska, line 21: may be add a new line between mods and use?

sevenzing: @leviska, sorry, didn't get it. where do you suggest adding new line? after 20th line?

verification/src/http_server/mod.rs
sevenzing: we need this value in `server_future `

verification/src/http_server/mod.rs
```diff
@@ -1,22 +1,47 @@
...
+    };
+    let mut futures = vec![tokio::spawn(async move { server_future.await })];
+    if metrics_enabled {
+        futures.push(tokio::spawn(async move {
+            metrics.run_server(metrics_addr).await
+        }))
+    }
+    for future in future::join_all(futures).await.into_iter() {
+        future??
+    }
```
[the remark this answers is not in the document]
sevenzing: changed `join_all` to `select_all`

verification/src/lib.rs
leviska: yes

PR #66, opened 2022-07-28
Add tracing and jaeger

sevenzing:
jaeger usage examples:

image

### DIFF
--- a/verification/example_config.toml
+++ b/verification/example_config.toml
@@ -15,3 +15,6 @@ compilers_list_url = "https://solc-bin.ethereum.org/linux-amd64/list.json"
 [metrics]
 endpoint = "/metrics"
 addr = "0.0.0.0:6060"
+
+[tracing]
+jaeger_agents = ["127.0.0.1:6831"]
--- a/verification/example_config.toml
+++ b/verification/example_config.toml
@@ -17,6 +17,7 @@ enabled = false
 endpoint = "/metrics"
 addr = "0.0.0.0:6060"
 
-[tracing]
-enable_jaeger = false
-jaeger_agent = "localhost:6831"
+[jaeger]
+enabled = false
+agent_endpoint = "localhost:6831"
+
--- a/verification/example_config.toml
+++ b/verification/example_config.toml
@@ -19,4 +19,4 @@ addr = "0.0.0.0:6060"
 
 [tracing]
 enable_jaeger = false
-jaeger_agent = "127.0.0.1:6831"
+jaeger_agent = "localhost:6831"
--- a/verification/example_config.toml
+++ b/verification/example_config.toml
@@ -13,8 +13,10 @@ refresh_versions_schedule = "0 0 * * * * *"
 compilers_list_url = "https://solc-bin.ethereum.org/linux-amd64/list.json"
 
 [metrics]
+enabled = false
 endpoint = "/metrics"
 addr = "0.0.0.0:6060"
 
 [tracing]
-jaeger_agents = ["127.0.0.1:6831"]
+enable_jaeger = false
+jaeger_agent = "127.0.0.1:6831"
--- a/verification/src/compiler/compilers.rs
+++ b/verification/src/compiler/compilers.rs
@@ -6,7 +6,7 @@ use crate::{
 use ethers_solc::{artifacts::Severity, error::SolcError, CompilerInput, CompilerOutput, Solc};
 use std::{fmt::Debug, path::PathBuf, sync::Arc};
 use thiserror::Error as DeriveError;
-use tracing::{instrument, Instrument};
+use tracing::instrument;
 
 #[derive(Debug, DeriveError)]
 pub enum Error {
@@ -39,10 +39,8 @@ impl Compilers {
         input: &CompilerInput,
     ) -> Result<CompilerOutput, Error> {
         let solc_path_result = {
-            let span = tracing::debug_span!("get solc_path from cache");
             self.cache
                 .get(self.fetcher.as_ref(), compiler_version)
-                .instrument(span)
                 .await
         };
         let solc_path = match solc_path_result {
--- a/verification/src/compiler/compilers.rs
+++ b/verification/src/compiler/compilers.rs
@@ -44,6 +44,8 @@ impl Compilers {
         let solc = Solc::from(solc_path);
         let output = {
             let _timer = metrics::COMPILE_TIME.start_timer();
+            let _span =
+                tracing::debug_span!("compile contract", ver = compiler_version.to_string());
             solc.compile(&input)?
         };
 
@@ -74,7 +76,7 @@ impl Compilers {
         match self.cache.load_from_dir(dir).await {
             Ok(_) => {}
             Err(e) => {
-                log::error!("error during local compilers loading: {}", e)
+                tracing::error!("error during local compilers loading: {}", e)
             }
         };
     }
--- a/verification/src/compiler/compilers.rs
+++ b/verification/src/compiler/compilers.rs
@@ -6,6 +6,7 @@ use crate::{
 use ethers_solc::{artifacts::Severity, error::SolcError, CompilerInput, CompilerOutput, Solc};
 use std::{fmt::Debug, path::PathBuf, sync::Arc};
 use thiserror::Error as DeriveError;
+use tracing::{instrument, Instrument};
 
 #[derive(Debug, DeriveError)]
 pub enum Error {
@@ -31,21 +32,31 @@ impl Compilers {
             fetcher,
         }
     }
-
+    #[instrument(name = "download_and_compile", skip(self), level = "debug")]
     pub async fn compile(
         &self,
         compiler_version: &compiler::Version,
         input: &CompilerInput,
     ) -> Result<CompilerOutput, Error> {
-        let solc_path = match self.cache.get(&*self.fetcher, compiler_version).await {
+        let solc_path_result = {
+            let span = tracing::debug_span!("get solc_path from cache");
+            self.cache
+                .get(self.fetcher.as_ref(), compiler_version)
+                .instrument(span)
+                .await
+        };
+        let solc_path = match solc_path_result {
             Err(FetchError::NotFound(version)) => return Err(Error::VersionNotFound(version)),
             res => res?,
         };
         let solc = Solc::from(solc_path);
         let output = {
             let _timer = metrics::COMPILE_TIME.start_timer();
-            let _span =
-                tracing::debug_span!("compile contract", ver = compiler_version.to_string());
+            let span = tracing::debug_span!(
+                "compile contract with ethers-solc",
+                ver = compiler_version.to_string()
+            );
+            let _guard = span.enter();
             solc.compile(&input)?
         };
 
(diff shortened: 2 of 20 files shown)


verification/src/tracer.rs
```diff
@@ -0,0 +1,24 @@
...
+        tracing_subscriber::EnvFilter::builder()
+            .with_default_directive(LevelFilter::INFO.into())
+            .from_env_lossy(),
+    );
+    let tracer = init_jaeger_tracer().expect("failed to init tracer");
+    tracing_subscriber::registry()
+        // output logs (tracing) to stdout with log level taken from env (default is INFO)
+        .with(stdout)
+        // output traces to jaeger with default log level (default is TRACE)
+        .with(tracing_opentelemetry::layer().with_tracer(tracer))
```
leviska, line 15: may be add tracing endpoint too? like with metrics

verification/src/tracer.rs
```diff
@@ -0,0 +1,24 @@
+use opentelemetry::{sdk::trace::Tracer, trace::TraceError};
+use tracing_subscriber::{filter::LevelFilter, layer::SubscriberExt, prelude::*};
+
+pub fn init_logs() {
```
leviska, line 4: May be add "enabled" option, so people could disable tracing altogether (or even disable them by default)
but leave the logs

sevenzing: what do you mean by `disable tracing`? we use tracing for logs. you mean disable jaeger?

verification/src/tracer.rs
```diff
@@ -0,0 +1,39 @@
...
+            // output traces to jaeger with default log level (default is DEBUG)
+            .with(
+                tracing_opentelemetry::layer()
+                    .with_tracer(tracer)
+                    .with_filter(LevelFilter::DEBUG),
+            )
+            .try_init()
+    } else {
+        registry.try_init()
+    }
```
sevenzing, line 29: don't like this way, but Idk how to make it without else

verification/src/config.rs
```diff
@@ -87,6 +88,22 @@ impl Default for MetricsConfiguration {
     }
 }
 
+#[derive(Deserialize, Clone)]
+#[serde(default)]
+pub struct TracingConfiguration {
```
sevenzing, line 93: should I rename it to JaegerConfiguration? not sure about it

verification/src/tracer.rs
leviska: doesn't this work
```rust
let registry = if config.enable_jaeger {
  let tracer = ...;
  registry.with(...)
} else {
  registry
}
registry.try_init().expect(..);
```

leviska (approved):

verification/src/tracer.rs
sevenzing: Of course I tried this, but problem is that `.with()` takes self and returns not Self but `SomeStructure<Self>`, and you cannot case `registry` to `registry.with(...)`

leviska: well it seems you can only do this using dyn (if you even can), which in your case is actually worse imo
but I'll provide it jfyi
something like
```rust
let registry: Box<dyn SubscriberInitExt> = Box::new(
  if ... {
    let tracer = ...;
    registry.with(..) as dyn SubscriberInitExt
  } else {
    registry as dyn SubscriberInitExt
  }
);
```
or something like that

verification/src/compiler/compilers.rs
```diff
@@ -31,19 +32,31 @@ impl Compilers {
...
         &self,
         compiler_version: &compiler::Version,
         input: &CompilerInput,
     ) -> Result<CompilerOutput, Error> {
-        let solc_path = match self.cache.get(&*self.fetcher, compiler_version).await {
+        let solc_path_result = {
+            let span = tracing::debug_span!("get solc_path from cache");
+            self.cache
+                .get(self.fetcher.as_ref(), compiler_version)
+                .instrument(span)
```
leviska, line 45: For the most cases imo it's better to put the span inside the function, if possible
Like here, just put it inside cache

verification/src/compiler/download_cache.rs
```diff
@@ -47,7 +49,8 @@ impl DownloadCache {
             }
             None => {
                 let _timer = metrics::COMPILER_FETCH_TIME.start_timer();
-                self.fetch(fetcher, ver).await
+                let span = tracing::debug_span!("fetch compiler", ver = ver.to_string());
+                self.fetch(fetcher, ver).instrument(span).await
```
leviska, line 53: it's fine as of now, but just a question
I saw there is a `#instrument` macro, which will automatically add the span. Is there something like it for the trait function? Not the impl block, but the trait, like
```rust
trait Foo {
  #[instrument]
  fn foo();
}
```
If it exists, then I suggest changing the span to such macro, If not, current impl is okay

verification/src/compiler/fetcher.rs
```diff
@@ -44,15 +45,12 @@ fn create_executable(path: &Path) -> Result<File, std::io::Error> {
...
 pub fn validate_checksum(bytes: &Bytes, expected: H256) -> Result<(), Mismatch<H256>> {
-    let start = std::time::Instant::now();
-
-    let found = Sha256::digest(bytes);
-    let found = H256::from_slice(&found);
-
-    let took = std::time::Instant::now() - start;
-    // TODO: change to tracing
-    log::debug!("check hashsum of {} bytes took {:?}", bytes.len(), took,);
+    let found = {
```
leviska, line 50: why change this to a code block?

verification/src/compiler/list_fetcher.rs
```diff
@@ -105,6 +106,7 @@ impl ListFetcher {
         Ok(Self { versions, folder })
     }
 
+    #[instrument(skip(self) level = "debug")]
```
leviska, line 109: `    #[instrument(skip(self), level="debug")]` (comma)?

verification/src/compiler/s3_fetcher.rs
```diff
@@ -97,6 +98,7 @@ impl S3Fetcher {
         })
     }
 
+    #[instrument(skip(self) level = "debug")]
```
leviska, line 101: comma?

verification/src/compiler/download_cache.rs
sevenzing: no, I think it's not possible to add instrument to trait function

verification/src/compiler/versions_fetcher.rs
```diff
@@ -84,7 +84,7 @@ impl<T: PartialEq + Send + Sync + 'static> VersionsRefresher<T> {
         fetcher: Arc<F>,
         cron_schedule: Schedule,
     ) {
-        log::info!("spawn version refresh job");
+        tracing::info!("spawn version refresh job");
         scheduler::spawn_job(cron_schedule, "refresh compiler versions", move || {
             let fetcher = fetcher.clone();
```
leviska, line 89: can we create new trace here? so we could search for scheduled things later in jaeger

verification/src/compiler/compilers.rs
sevenzing: this span just doesn't attach to current trace without it. here is more about it:
https://docs.rs/tracing/latest/tracing/struct.Span.html#in-asynchronous-code

verification/src/compiler/versions_fetcher.rs
```diff
@@ -32,7 +32,7 @@ impl<T> VersionsRefresher<T> {
 
 impl<T: PartialEq> VersionsRefresher<T> {
     pub async fn refresh<F: VersionsFetcher<Versions = T>>(&self, fetcher: &F) {
-        log::info!("looking for new compilers versions");
+        tracing::info!("looking for new compilers versions");
```
leviska, line 35: add span here may be

verification/src/compiler/versions_fetcher.rs
```diff
@@ -32,7 +32,7 @@ impl<T> VersionsRefresher<T> {
...
         let refresh_result = fetcher.fetch_versions().await;
```
leviska, line 36: add span for this await

verification/src/http_server/handlers/verification/solidity/contract_verifier.rs
```diff
@@ -68,13 +70,15 @@ pub(crate) async fn compile_and_verify_handler(
...
 
+#[instrument(skip(compilers, verifier, input), level = "debug")]
 async fn compile_and_verify(
     compilers: &Compilers,
     verifier: &Verifier,
     input: &Input<'_>,
 ) -> Result<VerificationSuccess, CompileAndVerifyError> {
     let compiler_output = compilers
         .compile(&input.compiler_version, &input.compiler_input)
+        .in_current_span()
```
leviska, line 81: is this doing anything? like, you've already got a span for whole function, and this call isn't a separate span, what happens if you add this line? O.o

verification/src/compiler/compilers.rs
leviska: I'm talking about just writing the macro for this function
image

verification/src/compiler/versions_fetcher.rs
leviska: (not the span, but a trace)

verification/src/compiler/fetcher.rs
```diff
@@ -72,21 +70,26 @@ pub async fn write_executable(
...
-                } else {
-                    Err(e)
-                }
-            })?;
-            let mut file = create_executable(file.as_path())?;
-            std::io::copy(&mut data.as_ref(), &mut file)?;
-            Ok(())
+        let span = tracing::debug_span!("save executable");
+        tokio::task::spawn_blocking(move || {
+            span.in_scope(|| {
```
leviska, line 75: I know you've talked about this, but what's the difference between `in_scope` and `enter`?
I mean, in the code it's just the same
https://docs.rs/tracing/latest/src/tracing/span.rs.html#785-788
https://docs.rs/tracing/latest/src/tracing/span.rs.html#1100-1103

verification/src/compiler/versions_fetcher.rs
leviska: I mean for the whole function, `#instrument` may be better

verification/example_config.toml
```diff
@@ -13,5 +13,10 @@ refresh_versions_schedule = "0 0 * * * * *"
 compilers_list_url = "https://solc-bin.ethereum.org/linux-amd64/list.json"
 
 [metrics]
+enabled = false
 endpoint = "/metrics"
 addr = "0.0.0.0:6060"
+
+[tracing]
+enable_jaeger = false
+jaeger_agent = "localhost:6831"
```
rimrakhimov, line 22: I suggest to name the section `jaeger` and use `enabled` and `agent_addr` as keys.

As another option we may use `tracing.jaeger` as a subsection, but not sure whether it make sense now, as we do not have any other options for `tracing` section

rimrakhimov (approved):

verification/src/compiler/fetcher.rs
sevenzing: changed to `.enter()`

leviska (approved):

PR #69, opened 2022-08-03
Change StandardJson

sevenzing:
Decided to accept standard json input as string for convenience

### DIFF
--- a/verification/src/http_server/handlers/verification/solidity/types.rs
+++ b/verification/src/http_server/handlers/verification/solidity/types.rs
@@ -72,7 +72,7 @@ impl TryFrom<StandardJson> for CompilerInput {
 
     fn try_from(input: StandardJson) -> Result<Self, Self::Error> {
         serde_json::from_str(&input.input)
-            .map_err(|e| anyhow::Error::msg(format!("cannot parse input: {}", e)))
+            .map_err(|e| anyhow::anyhow!("content is not valid standard json: {}", e))
     }
 }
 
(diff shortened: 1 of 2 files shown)


verification/src/http_server/handlers/verification/solidity/types.rs
```diff
@@ -64,12 +64,15 @@ impl TryFrom<MultiPartFiles> for CompilerInput {
...
 
-impl From<StandardJson> for CompilerInput {
-    fn from(input: StandardJson) -> Self {
-        input.input
+impl TryFrom<StandardJson> for CompilerInput {
+    type Error = anyhow::Error;
+
+    fn try_from(input: StandardJson) -> Result<Self, Self::Error> {
+        serde_json::from_str(&input.input)
+            .map_err(|e| anyhow::Error::msg(format!("cannot parse input: {}", e)))
```
rimrakhimov, line 75: Let's maybe provide more descriptive name? I suggest something like "content is not valid standard json: {}"

verification/src/http_server/handlers/verification/solidity/types.rs
```diff
@@ -208,7 +211,7 @@ mod tests {
             "deployed_bytecode": "0x6001",
             "creation_bytecode": "0x6001",
             "compiler_version": "v0.8.2+commit.661d1103",
-            "input":{"language":"Solidity","sources":{"./src/contracts/Foo.sol":{"content":"pragma solidity ^0.8.2;\n\ncontract Foo {\n    function bar() external pure returns (uint256) {\n        return 42;\n    }\n}\n"}},"settings":{"metadata":{"useLiteralContent":true},"optimizer":{"enabled":true,"runs":200},"outputSelection":{"*":{"*":["abi","evm.bytecode","evm.deployedBytecode","evm.methodIdentifiers"],"":["id","ast"]}}}}
```
rimrakhimov, line 211: Why? I suggest to better add into test that the valie StandardJson could be converted into CompilerInput

rimrakhimov (approved):

verification/src/http_server/handlers/verification/solidity/types.rs
```diff
@@ -64,12 +64,15 @@ impl TryFrom<MultiPartFiles> for CompilerInput {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct StandardJson {
```
leviska, line 66: do we really need this struct? can't we just make 
```json
"standard_json": "{json}"
```
and not 
```json
"standard_json": {
 "input": "{json}"
},
```
?
Is this future proofing?

sevenzing: we need StandardJson because `T` in `VerificationRequest<T>` should be struct with fields

rimrakhimov (approved):

leviska (approved):

PR #71, opened 2022-08-03
Update configuration setup

rimrakhimov:
Updates verification service configuration. PR establishes the following order:

1. Settings structure has default values. Initially, the structure is filled with those values.
1. Configuration file to be used may be setup with `VERIFICATION__CONFIG` environment variable. If provided, the service tries to read from that file and overwrites all values that are provided in the config.
 
    As an example and for convinience, `config/base.toml` contains the configuration with default values
1. Finally, environment variables with prefix `VERIFICATION` and `__` as a separator overwrites previous values and may be used for the final setup.


verification/config/base.toml
```diff
@@ -0,0 +1,31 @@
...
+#region = "region"
+#endpoint = "endpoint"
+## The only required field for the s3 fetcher
+#bucket = "bucket"
+
+[sourcify]
+enabled = true
+api_url = "https://sourcify.dev/server/"
+verification_attempts = 3
+request_timeout = 10
```
rimrakhimov, line 27: Thinking about naming those as `url`, `attempts` and `timeout` as well

verification/tests/settings.rs
```diff
@@ -0,0 +1,11 @@
+use pretty_assertions::assert_eq;
+use verification::Settings;
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_example_settings() {
```
sevenzing, line 6: maybe add `#[ignore]` to this test and test it with `cargo test -- --include-ignored` ?

sevenzing (approved):

verification/config/base.toml
sevenzing: agree

verification/config/base.env
```diff
@@ -0,0 +1,25 @@
+#VERIFICATION__SERVER__ADDR=0.0.0.0:8043
```
leviska, line 1: I think most env files taking the token without '#'
https://docs.docker.com/compose/environment-variables/#the-env-file
and `#string` is actually a comment

verification/dockerfile
```diff
@@ -5,6 +5,7 @@ WORKDIR /build_app
 # cache dependencies
 RUN cargo init
 COPY ./Cargo.toml ./
+# COPY ./Cargo.lock ./
```
leviska, line 8: remove?

verification/src/settings.rs
```diff
@@ -0,0 +1,173 @@
...
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct Settings {
+    pub server: ServerSettings,
+    pub solidity: SoliditySettings,
+    pub sourcify: SourcifySettings,
+    pub metrics: MetricsSettings,
+
+    #[serde(rename = "config")]
+    pub config_path: IgnoredAny,
```
leviska, line 18: why do we even need this? it's kinda strange to have config path in the config, we don't use it anywhere, ...?

verification/tests/settings.rs
leviska: but why it depends on os?...

leviska:
I don't quite understand why did you rename `config` into `settings`. A lot of changes just to change this name, also while the files with the actual config didn't changed too much, it seems like git couldn't detect, that this is still the old file so we're lost all history for it.

leviska (approved):

verification/tests/settings.rs
rimrakhimov: @leviska Because default `solidity.fetcher.list.list_url` is different for linux and macos/windows systems

verification/config/base.env
rimrakhimov: You are right, but the main idea here was just to have a file with all configuration variables and their default values. So that users could copy required variables if they would like to use envvars to customize their setup. 

If we do not comment those values, however, they will overwrite values written in `basic.toml` and no changes to `basic.toml` will take a place when the default docker-compose is running. In order to avoid that I commented values in `base.env`

rimrakhimov: I've added some comments describing that into `base.env`

verification/src/settings.rs
rimrakhimov: I've added a comment describing why it is required 

The main idea is that we deny all unknown fields, so that users could notice any errors (otherwise, if a user make any mistake in the name of the field we will just use a default value without any warning or notification). At the same time, path to configuration file is setup through `PREFIX__CONFIG` (`VERIFICATION__CONFIG`, in our case) environment variable where the prefix is used to discriminate variables our service should consider. Thus, if we remove `config_path` field, `config` value would be considered as unknown during deserialization

verification/tests/settings.rs
leviska: let's make this test ignore this field, or set it manually to the OS default, or check this field only on linux, and ignore only the field on other systems, not the whole test. 
I think there is no real reason why we can't test the config on the macos or windows, and one different line is not an excuse

verification/src/settings.rs
leviska: ah, it would fail during ENV deserialization?

rimrakhimov: Yeap

verification/tests/settings.rs
rimrakhimov: Made them run on other systems as well without checking `list_url`

sevenzing (commented):
please change `Configuration` section in `verification/README.md`

PR #73, opened 2022-08-04
fix docker compose

sevenzing:


rimrakhimov (approved):

verification/docker-compose.yaml
```diff
@@ -1,9 +1,12 @@
...
+    ## you can build image from source if you want
+    # build: .
+    image: ghcr.io/blockscout/verification:latest
     ports:
       - "8043:8043"
     command: -c /app/config.toml
     volumes:
-      - ./example_config.toml:/app/config.toml
\ No newline at end of file
+      - ./example_config.toml:/app/config.toml
```
leviska, line 11: add
```yaml
# optional: you can provide custom config
 - ./example_config.toml:/app/config.toml
# optional: provide volume or folder to store compilers between launches
 - ./compilers:/app/compilers
```

verification/docker-compose.yaml
```diff
@@ -1,9 +1,12 @@
 version: "3"
 services:
   verification:
-    build: .
+    ## you can build image from source if you want
```
leviska, line 4: `# you can build image locally, or use prebuilt images from registry`

verification/docker-compose.yaml
leviska: ah, not providing the config file will not work because of the `-c /app/config.toml`, isn't it? :(

leviska: then may be add before command
``# you can run the service with default config, for that you need to remove `command` parameter``

leviska (approved):

PR #74, opened 2022-08-05
fixed s3 hash computing

leviska:


verification/src/compiler/s3_fetcher.rs
```diff
@@ -221,7 +227,7 @@ mod tests {
 
         mock_get_object(
             "/solc-releases/v0.4.10%2Bcommit.f0d539ae/sha256.hash",
-            &expected_hash,
+            hex::encode(expected_hash).as_bytes(),
         )
```
rimrakhimov, line 231: Could you please also add a test, that checksum extraction works for hashes which start with "0x" prefix (e.g. `0xacc9d8073a71c9d53dc12d3d3aed361d6b04c63adc0dc811839669e9f5ab4a91`)

PR #78, opened 2022-08-11
added docker build and push for verification

leviska:


rimrakhimov (approved):

rimrakhimov:
Could we add a notion of `latest` image which would point to the image with the greatest tag (ignoring main images)?

leviska:
We can make that pushing the tag will also push it as latest, so latest will be not the greatest tag, but the... latest :)

leviska:
And due to tags can have any text, not specifically semver, it would be hard to compare tags to find the greatest

rimrakhimov (approved):

PR #83, opened 2022-08-15
Update verification: add docker-compose example with jaeger

rimrakhimov:


sevenzing:
maybe also add `grafana` and `prometheus` for metrics?

verification/docker-compose.yaml
```diff
@@ -2,8 +2,8 @@ version: "3"
 services:
```
sevenzing, line 2: move this file to `docker-compose` dir?

rimrakhimov:
@sevenzing

PR #89, opened 2022-08-16
Kuksag/edit multichain api config

kuksag:
Rewrite service configuration in a manner of "[smart-contract-verifier](https://github.com/blockscout/blockscout-rs/tree/main/smart-contract-verifier)" service.

This is a part of achieving consistency in a code style among services.


multichain-api-gateway/src/settings.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub instances: Vec<Instance>,
+
+    /// The number of concurrent requests to be made to the Blockscout API from a server's thread worker.
+    pub concurrent_requests: usize,
+
+    /// The timeout of waiting for response from the Blockscout API.
+    // Using chrono::Duration here instead of std::time::Duration,
+    // because there is no "Display" trait for std::time::Duration, hence trouble when calling "try_deserialize"
+    #[serde_as(as = "serde_with::DurationSeconds<i64>")]
+    pub request_timeout: chrono::Duration,
```
rimrakhimov, line 30: If you use `serde_with::DurationSeconds<u64>` you can make the `request_timeout` to be `std::time::Duration`

multichain-api-gateway/src/lib.rs
```diff
@@ -11,16 +11,16 @@ use futures::{stream, StreamExt};
 use serde_json::Value;
 use url::Url;
 
-use crate::config::{BlockscoutSettings, Instance, Settings};
+pub use crate::settings::{BlockscoutSettings, Instance, Settings};
 
 mod cli;
```
rimrakhimov, line 16: You may remove cli module and even `cli.rs` file, as they are not used anymore

multichain-api-gateway/src/settings.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+impl Default for BlockscoutSettings {
+    fn default() -> Self {
+        Self {
+            base_url: Url::parse("https://blockscout.com/").expect("should be correct base"),
+            instances: vec![],
```
rimrakhimov, line 37: I suggest to add some instances by default as well (e.g. eth/mainnet and xdai/mainnet)

rimrakhimov (commented):
I've added some small suggestions and remarks. One important concept, though, is the one with `base.toml` and `base.env` (which is missed). The main idea was to keep them as a basic examples of what server configuration options are available and what are default values. So, it would be great, if `base.toml` would contain all possible configuration options with default values set. Another configuration example is `base.env` which would be also great to provide with all options set to default.

Besides that, have you tried to set `blockscout.instance` vector through corresponding environment variable? I'm not sure it would work as it is, so would be great to check that before merging

kuksag:
> have you tried to set blockscout.instance vector through corresponding environment variable? 

I tried using the following grammar:
```shell
KEY=value
KEY="Some other value"
KEY=value1:value2
```
But it did nothing as I received something like:
```shell
Error: failed to parse config                                                                            
                                                                                                         
Caused by:                                                                                               
    invalid type: string "123:456", expected a sequence
```

multichain-api-gateway/config/base.toml
```diff
@@ -3,7 +3,7 @@ addr = "0.0.0.0:8044"
 
 [blockscout]
 base_url = "https://blockscout.com"
-concurrent_requests = 2
+concurrent_requests = 10
 instances = [
     ["eth", "mainnet"],
     ["xdai", "testnet"],
```
rimrakhimov, line 9: I believe, it does not work, if you try to run your service with `MULTICHAIN_API_GATEWAY__CONFIG=config/base.toml cargo run`

rimrakhimov: What maybe if you use not the array notation you use now, but just require to provide net and chain information in the way it is currently provided into url path (e.g. `eth/mainet`)? And then parse it with custom `#[serde(deserialize_with=custom_function)]` (https://serde.rs/field-attrs.html#deserialize_with) instead of `#[serde(with = "As::<Vec<DisplayFromStr>>")]` in BlockscoutSettings.instances part

rimrakhimov: Let's though fix that in another PR maybe

rimrakhimov (approved):

rimrakhimov (approved):

PR #90, opened 2022-08-18
download file to tmp first, and rename when everything is ok

leviska:


smart-contract-verifier/src/compiler/fetcher.rs
```diff
@@ -95,6 +97,8 @@ pub async fn write_executable(
     check_result??;
     save_result??;
 
+    tokio::fs::rename(&file_tmp, &file).await?;
```
rimrakhimov, line 100: What if `file` already exists?

smart-contract-verifier/src/compiler/fetcher.rs
```diff
@@ -106,14 +110,39 @@ mod tests {
...
+            .await
+            .expect_err("expected to fail with wrong checksum");
+        assert!(matches!(err, FetchError::HashMismatch(_)));
+
+        let dir = tmp_dir.path().join(version.to_string());
+        let file = dir.join("solc");
+        let mut tmp_file = file.clone();
+        tmp_file.set_extension("tmp");
+        assert!(!file.exists());
+        assert!(tmp_file.exists());
```
rimrakhimov, line 146: Shouldn't temporary file be deleted if checksum was invalid?

rimrakhimov (approved):

smart-contract-verifier/src/compiler/fetcher.rs
leviska: > Rename a file or directory to a new name, **replacing the original file if to already exists**.

https://doc.rust-lang.org/std/fs/fn.rename.html

smart-contract-verifier/src/compiler/fetcher.rs
leviska: we can make this, but there still will be cases, where tmp file could exists, for example if server will shutdown while saving the file.

we can try to delete file in some places, but I don't see the reason to do so, because we can't delete it in every case.  Specially seeing that checksum mismatch is something rare

rimrakhimov: Okay

rimrakhimov (approved):

PR #91, opened 2022-08-18
added solc validator, which runs solc and checks it's version

### LINKED ISSUE #103: Add checks for downloaded compilers
We may run `--version` on downloaded solc compilers to ensure that they work and return correct compiler version. That adds some level of confidence that the downloaded compiler is correct and works

leviska:
Closes #103


smart-contract-verifier/src/compiler/compilers.rs
```diff
@@ -98,14 +98,14 @@ mod tests {
     use std::{env::temp_dir, str::FromStr};
 
     use crate::consts::DEFAULT_COMPILER_LIST;
-    use async_once_cell::OnceCell;
     use ethers_solc::artifacts::{Source, Sources};
     use std::default::Default;
+    use tokio::sync::OnceCell;
```
rimrakhimov, line 103: What was the reason behind changing `async_once_cell::OnceCell` on `tokio::sync::OnceCell`?

smart-contract-verifier/src/compiler/compilers.rs
```diff
@@ -98,14 +98,14 @@ mod tests {
...
 
     async fn global_compilers() -> &'static Compilers {
-        static COMPILERS: OnceCell<Compilers> = OnceCell::new();
+        static COMPILERS: OnceCell<Compilers> = OnceCell::const_new();
         COMPILERS
-            .get_or_init(async {
+            .get_or_init(move || async {
```
rimrakhimov, line 108: Nit:
```suggestion
            .get_or_init(|| async {
```

smart-contract-verifier/src/compiler/fetcher.rs
```diff
@@ -27,11 +28,19 @@ pub enum FetchError {
...
+
+#[async_trait]
+pub trait FileValidator: Send + Sync {
+    async fn validate(&self, ver: &Version, path: &Path) -> Result<(), anyhow::Error>;
 }
 
 #[async_trait]
 pub trait Fetcher: Send + Sync {
     async fn fetch(&self, ver: &Version) -> Result<PathBuf, FetchError>;
+    fn with_validator(&mut self, validator: Arc<dyn FileValidator>);
```
rimrakhimov, line 43: Why not just to add a validator as a constructor argument into `ListFetcher::new()` and `S3Fetcher::new()`? It seems that the code would be simpler in that case, while you would get the same result. Additionally, you may accept not `Option<dyn FileValidator>` but `dyn FileValidator` itself into the constructor in that case, which I'm not sure that is better, but still give you the same result with less amount of code and abstractions

smart-contract-verifier/tests/solidity_multiple.rs
```diff
@@ -18,14 +17,15 @@ use std::{
...
 
 const CONTRACTS_DIR: &'static str = "tests/contracts";
 const ROUTE: &'static str = "/api/v1/solidity/verify/multiple-files";
 
 async fn global_app_router() -> &'static AppRouter {
-    static APP_ROUTER: OnceCell<AppRouter> = OnceCell::new();
+    static APP_ROUTER: OnceCell<AppRouter> = OnceCell::const_new();
     APP_ROUTER
-        .get_or_init(async {
+        .get_or_init(move || async {
```
rimrakhimov, line 28: Nit: 
```suggestion
        .get_or_init(|| async {
```

smart-contract-verifier/src/solidity/validator.rs
```diff
@@ -0,0 +1,102 @@
...
+    use tokio::sync::OnceCell;
+
+    fn default_version() -> Version {
+        Version::from_str("v0.8.9+commit.e5eed63a").unwrap()
+    }
+
+    async fn fetch_compiler() -> PathBuf {
+        static COMPILERS: OnceCell<PathBuf> = OnceCell::const_new();
+        COMPILERS
+            .get_or_init(move || async {
```
rimrakhimov, line 49: Nit: 
```suggestion
            .get_or_init(|| async {
```

smart-contract-verifier/tests/standard_json.rs
```diff
@@ -18,14 +17,15 @@ use std::{
...
 
 const CONTRACTS_DIR: &'static str = "tests/contracts";
 const ROUTE: &'static str = "/api/v1/solidity/verify/standard-json";
 
 async fn global_app_router() -> &'static AppRouter {
-    static APP_ROUTER: OnceCell<AppRouter> = OnceCell::new();
+    static APP_ROUTER: OnceCell<AppRouter> = OnceCell::const_new();
     APP_ROUTER
-        .get_or_init(async {
+        .get_or_init(move || async {
```
rimrakhimov, line 28: Nit:
```suggestion
        .get_or_init(|| async {
```

smart-contract-verifier/src/compiler/compilers.rs
leviska: why include some questionable additional dependency, when we already have it in tokio?

smart-contract-verifier/src/compiler/fetcher.rs
leviska: because I thought that validators should be a part of fetcher interface. Like, you can fetch and while fetching run some custom validator
And I think option is actually still needed, because we might use this fetchers for vyper compilation, for example, and there could be no validation
If we make validator as part of constructor, then if someone will add new fetcher, he could not include validator in it's class, but I think we'd like to require all fetcher to support validators

smart-contract-verifier/src/compiler/compilers.rs
rimrakhimov: Okay, so just because the second library is more authoritative

smart-contract-verifier/src/compiler/fetcher.rs
rimrakhimov: > because I thought that validators should be a part of fetcher interface. Like, you can fetch and while fetching run some custom validator

Yes, but it does not bring to the `Fetcher` trait any additional meaning. You can easily do nothing on that `with_validator` method, as well as to pass `None` to `write_executable` even if some validator has been set with `with_validator`

> And I think option is actually still needed, because we might use this fetchers for vyper compilation, for example, and there could be no validation

Maybe we may just create a `NoOpValidator` which can be used as a default validator for fetching data that could not be validated? I mean, we could leave the `Option` as well, but removing it may make the code more clear. Additionally, `write_executable` would require to pass `FileValidator` as an argument, so that all fetchers that uses that method (supposedly, all **should**) would be forced to pass something as a validator here and, thus, think about what it should be.

leviska: I think `Option<dyn Validator>` is much clearer that `dyn Validator = NoOpValidator`

sevenzing (approved):

smart-contract-verifier/src/compiler/fetcher.rs
rimrakhimov: Ok, let's leave `Option<dyn Validator>, but still I cannot see what benefit `Fetcher::with_validator` brings rather than adding that function into fetcher constructor. It does not enforce anything new in comparison with passing validator into constructor

leviska: As I probably said in PMs, the better way to do it is to make `write_executable` a struct, which will store the validator, accept it in contsructor, and pass this struct into the fetchers. But I thought that it's a little over engineering at current point at time and decided to make an easy solution before making something more complex :)

rimrakhimov (approved):

sevenzing (approved):

PR #93, opened 2022-08-22
Draft: smart contract service init structure

leviska:


smart-contract-service/src/proto/service.proto
```diff
@@ -214,29 +202,26 @@ message VerifyVyperResponse {
...
 message GetVerificationStatusResponse {
   enum Status {
-    PENDING = 0;
-    PASS = 1;
-    FAIL = 2;
-    UNKNOWN_ID = 3;
+    UNKNOWN_ID = 0;
+    PENDING = 1;
+    PASS = 2;
+    FAIL = 3;
```
rimrakhimov, line 224: Life would be much easier that way :upside_down_face:  

But, I'm not sure we could do that here, as current implementation has already established another order for answers (https://github.com/blockscout/blockscout/blob/master/apps/explorer/lib/explorer/chain/smart_contract/verification_status.ex#L62). I'm not sure whether we should or should not be backward compatible in that place, but while we do not know for sure I believe it's better to assume that we should. Otherwise, we may make a system that relies on `UNKNOWN_ID` being the default value and then find out that it is not true anymore

smart-contract-service/src/proto/service.proto
```diff
@@ -275,9 +260,10 @@ service SmartContractsService {
 
   // `POST /verify-vyper` - verifies vyper contracts
   rpc VerifyVyper(VerifyVyperRequest) returns (VerifyVyperResponse);
+}
 
-  // Async verification related endpoints
-
+// Async verification related endpoints
+service AsyncSmartContractService {
```
rimrakhimov, line 266: `SmartContractAsyncService`? To be consistent with methods naming

PR #94, opened 2022-08-22
Add solc-cli module to fix 0.4.10 versions

sevenzing:


smart-contract-verifier/src/solidity/solc_cli.rs
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    fn merge_libs(libraries: Libraries) -> BTreeMap<String, String> {
+        libraries
+            .libs
+            .into_iter()
+            .map(|(_, libs)| libs)
+            .reduce(|accum, item| accum.into_iter().chain(item).collect())
+            .unwrap_or_default()
+    }
```
rimrakhimov, line 60: Maybe we may do something like this to make it more clear?

```suggestion
    fn merge_libs(libraries: Libraries) -> BTreeMap<String, String> {
        let mut result = BTreeMap::new();
        libraries
            .libs
            .into_iter()
            .for_each(|(_, libs)| result.extend(libs));
        result
    }
```

smart-contract-verifier/src/solidity/solc_cli.rs
```diff
@@ -0,0 +1,493 @@
...
+mod types {
+    use super::serde_helpers;
+    use std::collections::{BTreeMap, HashMap};
+
+    use ethers_solc::{
+        artifacts::{Contract, Libraries},
+        error::SolcError,
+        CompilerInput, CompilerOutput,
+    };
+    use minicbor::encode::Write;
```
rimrakhimov, line 42: That is not what you want to import

```suggestion

```

smart-contract-verifier/src/solidity/solc_cli.rs
```diff
@@ -0,0 +1,493 @@
...
+                vec.push(libs.join(","))
+            }
+            vec
+        }
+    }
+
+    #[derive(Debug)]
+    pub struct InputFiles {
+        pub files_dir: TempDir,
+        pub file_names: Vec<String>,
```
rimrakhimov, line 104: Why not to make it `PathBuf`?
```suggestion
        pub file_names: Vec<PathBuf>,
```

smart-contract-verifier/src/solidity/solc_cli.rs
```diff
@@ -0,0 +1,493 @@
...
+
+            Ok(InputFiles {
+                files_dir,
+                file_names,
+            })
+        }
+    }
+
+    impl InputFiles {
+        pub fn build_string_args(&self) -> Result<&Vec<String>, SolcError> {
```
rimrakhimov, line 129: Nit: maybe just `build`?

smart-contract-verifier/src/solidity/solc_cli.rs
```diff
@@ -0,0 +1,493 @@
...
+        r#type: "".to_string(),
+        component: "".to_string(),
+        severity: Severity::Error,
+        error_code: None,
+        message,
+        formatted_message: None,
+    }
+}
+
+pub fn compile_using_cli(
```
rimrakhimov, line 221: Nit: personally, I believe it would be better to rename that function to `compile`, make `pub(crate) mod solc_cli` inside `mod.rs`, and use it as `solc_cli::compile` from the `compilers.rs`

smart-contract-verifier/src/solidity/solc_cli.rs
```diff
@@ -0,0 +1,493 @@
...
+    solc: &PathBuf,
+    input: &CompilerInput,
+) -> Result<CompilerOutput, SolcError> {
+    let output = {
+        let input_args = types::InputArgs::try_from(input)?;
+        let input_files = types::InputFiles::try_from(input)?;
+        Command::new(solc)
```
rimrakhimov, line 228: I suggest to use `tokio::process::Command::new` and mark the function as `async`

rimrakhimov (approved):

smart-contract-verifier/src/solidity/solc_cli.rs
sevenzing: eventually in `build_string_args` I have to convert it to string anyway

smart-contract-verifier/src/solidity/solc_cli.rs
sevenzing: `input_files.build()`  looks like "build input files" for me

sevenzing: ok maybe it make sense, not very important for me

smart-contract-verifier/src/solidity/solc_cli.rs
sevenzing: we have like 4 or 5 `compile` functions. don't you think it will confuse people?

smart-contract-verifier/src/solidity/solc_cli.rs
sevenzing: make sense

smart-contract-verifier/src/solidity/solc_cli.rs
```diff
@@ -0,0 +1,494 @@
...
+                "bin-runtime": "60606040526000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff168063481591811460505780636d4ce63c146070578063f3d33cf4146090575b6000565b34600057605a60b0565b6040518082815260200191505060405180910390f35b34600057607a60be565b6040518082815260200191505060405180910390f35b34600057609a60d3565b6040518082815260200191505060405180910390f35b60006414b230ce3890505b90565b600060c660d3565b60cc60b0565b0190505b90565b6000650712e7ae7c7190505b905600a165627a7a72305820a80a9599b36625e94a3eadfd5c31475a2c507be6d1a9fa9a35e9cd4c54bce1390029"
+            }
+        },
+        "version": "0.4.8+commit.60cc1668"
+    }
+    "#;
+
+    #[cfg(target_os = "linux")]
+    pub const LIST_JSON: &str = r#"
```
leviska, line 325: imo we could minimize lists(?) (make them one line)

leviska:
can you add a test with compile error (source code is broken) and execution error (for example, make `solc` file an empty file)?

I'd like to test that we're processing errors correctly

smart-contract-verifier/src/solidity/solc_cli.rs
sevenzing: I think it won't be readable for people who will try to understand why tests dont pass

leviska (approved):

smart-contract-verifier/src/solidity/solc_cli.rs
rimrakhimov: You can make `build_string_args` to return `Vec<PathBuf>` without any problem

smart-contract-verifier/src/solidity/solc_cli.rs
rimrakhimov: I actually believe that `compilers::compile` is excessive and сan just be directly inline inside `compilers::Compilers::compile`. So my idea was that after removal of `compilers::compile` we would have `solc_cli::compile` and `solc.compile` which does not seem actually confusing for me.

PR #97, opened 2022-08-24
Update verification algorithm

### LINKED ISSUE #102: Update verification algorithm
Current verification approach assumes that there is only one metadata part present in bytecode to be verified which could be extracted as the last bytes of deployed bytecode. But that is not always true. E.g. [issue 5636](https://github.com/blockscout/blockscout/issues/5636) presents the case where creation transaction input contains a creation code of another contract with its own metadata part. We cannot verify such contracts right now. 

It would be nice to update the metadata extraction mechanisms to support verification of bytecodes with multiple metadata parts

rimrakhimov:
PR updates verification algorithm and addresses verification of contracts with multiple metadata hashes (e.g. `contractCode` method call in [#5636](https://github.com/blockscout/blockscout/issues/5636)) that occurred with current approach

Description of a new approach could be found at: https://bronzed-activity-e6b.notion.site/Verification-Algorithm-Proposal-00caad53b71d4a18b68e952e1751517a

Closes #102


smart-contract-verifier/src/solidity/verifier.rs
```diff
@@ -424,116 +51,245 @@ impl Verifier {
...
+    ///
+    /// # Panics
+    ///
+    /// The function will panic if `remote_raw.len()` is less than `local_raw.len()`.
+    fn compare_bytecode_parts(
+        remote_raw: &Bytes,
+        local_raw: &Bytes,
+        local_parts: &Vec<BytecodePart>,
+    ) -> Result<(), VerificationErrorKind> {
+        assert!(
```
sevenzing, line 207: didn't get it, why not `return Err(...)` ?

smart-contract-verifier/src/solidity/verifier.rs
```diff
@@ -424,116 +51,245 @@ impl Verifier {
...
+            remote_raw.len() >= local_raw.len(),
+            "Local bytecode is greater than remote"
+        );
+
+        let mut i = 0usize; // keep track of current processing position of `remote_raw`
+
+        for part in local_parts {
+            match part {
+                BytecodePart::Main { raw } => {
+                    if raw != &remote_raw[i..i + raw.len()] {
```
sevenzing, line 218: as far as I understood, to compare this, `local_parts` has to be sorted basing on start, right? otherwise `&remote_raw[i..i + raw.len()]` will give some strange result. but I didn't find code where you sort your parts.

so is it sorted implicitly?

smart-contract-verifier/src/solidity/verifier.rs
rimrakhimov: It is not an error. It was added just to make explicit that precondition. It is the responsibility of the caller to ensure thtlat it never fails. I.e. currently it is used only in `compare_creation_tx_inputs` and there is condition that ensures that property. Probably it does worth to add a comment describing that line, thank you!

smart-contract-verifier/src/solidity/verifier.rs
rimrakhimov: Yes, we just parse the result of local compilation on Localbytecode initialization. And that parts are stored in the same order they are located in the bytecpde itself

smart-contract-verifier/src/solidity/errors.rs
```diff
@@ -0,0 +1,91 @@
+use crate::{types::Mismatch, DisplayBytes};
+use std::fmt::{Display, Formatter};
+use thiserror::Error;
+
+/// Errors that may occur during deployed bytecode and
+/// creation transaction input parsing.
+#[derive(Error, Clone, Debug, PartialEq, Eq)]
+pub enum BytecodeInitializationError {
```
leviska, line 8: may be `BytecodeInitError`? this looks very long in the code

smart-contract-verifier/src/http_server/handlers/verification/solidity/contract_verifier.rs
```diff
@@ -79,9 +79,31 @@ async fn compile_and_verify(
     let compiler_output = compilers
         .compile(&input.compiler_version, &input.compiler_input)
         .await?;
+    let compiler_output_modified = {
```
leviska, line 82: just checking: we can't compile only the original, check somehow easily for matches, and if nothing matched only then compile additional stuff? As I understood, we just threw away the old algo, so we need modified outputs for new?

smart-contract-verifier/src/solidity/bytecode.rs
```diff
@@ -0,0 +1,618 @@
...
+    fn parse_bytecode_parts(
+        raw: &Bytes,
+        raw_modified: &[u8],
+    ) -> Result<Vec<BytecodePart>, VerificationErrorKind> {
+        let mut parts = Vec::new();
+
+        let len = raw.len();
+
+        let mut i = 0usize;
+        while i < len {
```
leviska, line 200: let's remake this into more readable thing
basically you're searching for the common prefix and then do stuff with that. but you're doing stuff "while searching"
imo it will be much better if we decompose this stuff
1. Find common prefix, get the index of the last byte
2. Then do stuff
and we probably can use find methods on iterators (didn't test that but something along those lines)
```rust
let index = raw.iter().zip(raw_modified.iter()).enumerate().find(|(index, (left, right))| left != right).0;
// or at least one single for/while loop with only searching
```

smart-contract-verifier/src/http_server/handlers/verification/solidity/contract_verifier.rs
rimrakhimov: That's an interesting idea, thanks. Probably we can do that. However, I believe it would be better to make that at a separate PR and maybe after we understand that verification has started to take considerable amount of time we want to reduce

leviska (approved):

PR #99, opened 2022-08-26
Add vyper compilation

### LINKED ISSUE #100: Verification of Vyper contracts
Currently we support only verification of Solidity smart-contracts, while Blockscout also supports verification of Vyper contracts. 
That issue keeps track of current status of Vyper contracts verificaiton in _smart-contract-verifier_ service.

- [x] Implement download of Vyper compilers
- [x] Compilation of Vyper contracts
- [x] Verification of Vyper contracts
- [x] Add HTTP endpoints to handle `list_versions` and `verify` requests

sevenzing:
Added vyper compilation using solc compiler interface with minor changes

Changed version parsing to parse vyper.list.json

Changed `Compilers` struct to accept several evm compilers

Partially closes #100


smart-contract-verifier/src/vyper/compiler.rs
```diff
@@ -0,0 +1,190 @@
...
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+    pub struct SourceLocation {
+        pub file: String,
+    }
+
+    #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+    #[serde(rename_all = "camelCase")]
+    pub struct VyperError {
```
rimrakhimov, line 44: We may name it just an `Error` as we are working on in order not to repeat `vyper` module prefix
```suggestion
    pub struct Error {
```

smart-contract-verifier/src/vyper/compiler.rs
```diff
@@ -0,0 +1,190 @@
...
+    use ethers_solc::{
+        artifacts::{Contract, Severity},
+        CompilerOutput,
+    };
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+    pub struct SourceLocation {
+        pub file: String,
+    }
```
rimrakhimov, line 40: What about `lineno` and `col_offset` fields, and `SourceLocation` inside `VyperError` struct?

smart-contract-verifier/src/vyper/compiler.rs
```diff
@@ -0,0 +1,190 @@
...
+        pub formatted_message: Option<String>,
+    }
+
+    #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Default)]
+    pub struct VyperCompilerOutput {
+        #[serde(default, skip_serializing_if = "Vec::is_empty")]
+        pub errors: Vec<VyperError>,
+        #[serde(default)]
+        pub contracts: BTreeMap<String, BTreeMap<String, Contract>>,
+    }
```
rimrakhimov, line 58: What about `sources` field?

smart-contract-verifier/src/settings.rs
```diff
@@ -68,6 +69,29 @@ impl Default for SoliditySettings {
...
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct VyperSettings {
+    pub enabled: bool,
+    pub compilers_dir: PathBuf,
+    #[serde(with = "serde_with::rust::display_fromstr")]
+    pub refresh_versions_schedule: Schedule,
+    pub list_url: Url,
```
rimrakhimov, line 79: I suggest to make it the same way it is done for `SoliditySettings`: to make it through `pub fetcher: FetcherSettings` field

smart-contract-verifier/src/compiler/version.rs
```diff
@@ -10,50 +11,58 @@ pub enum ParseError {
...
+    /// `(v)*VERSION*(-*PRERELEASE*)+commit.*COMMITHASH*`, examples:
     /// `v0.8.9+commit.e5eed63a`
     /// `0.8.4+commit.dea1b9ec`
+    /// `0.8.4-beta.16+commit.dea1d`
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let (_prefix, major, minor, patch, commit_hash) = sscanf::scanf!(
+        let (_prefix, major, minor, patch, pre, commit) = sscanf::scanf!(
             s,
-            "{:/v?/}{}.{}.{}+commit.{}",
+            "{:/v?/}{}.{}.{}{:/.*/}+commit.{:/[A-Fa-f0-9]+/}",
```
rimrakhimov, line 35: Why not
```suggestion
            "{:/v?/}{}.{}.{}-{:/.*/}+commit.{:/[A-Fa-f0-9]+/}",
```
?

smart-contract-verifier/src/compiler/version.rs
```diff
@@ -10,50 +11,58 @@ pub enum ParseError {
...
     CommitHash(hex::FromHexError),
+    #[error("cannot parse semver: {0}")]
+    Semver(#[from] semver::Error),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct ReleaseVersion {
     pub version: semver::Version,
-    pub commit: [u8; 4],
+    pub commit: String,
```
rimrakhimov, line 21: I suggest to leave it as `[u8; 4]` but to add leading zeros in case if there is not enough bytes in the commit hash

smart-contract-verifier/src/compiler/compilers.rs
```diff
@@ -20,16 +24,33 @@ pub enum Error {
...
 
-pub struct Compilers {
+pub trait EvmCompilerAgent {
+    fn compile(
+        &self,
+        path: &Path,
+        ver: &compiler::Version,
+        input: &CompilerInput,
+    ) -> Result<CompilerOutput, SolcError>;
+}
```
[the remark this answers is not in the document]
sevenzing: I will add `solc-cli` feature very soon, I will need it there

smart-contract-verifier/src/vyper/compiler.rs
sevenzing: why do we need them?

smart-contract-verifier/src/compiler/version.rs
sevenzing: because minus sign is optional

sevenzing: it could be 
`{:/v?/}{}.{}.{}{:/-?/}{:/.*/}`, but I decided do it this way

smart-contract-verifier/src/compiler/version.rs
sevenzing: I did the exact same way you suggested, however there is some problems with that

+ what if in future we will have more that 4 bytes?
+ `v0.1.1+commit.12` and `v0.1.1+commit.1200` are the same, but what if they are not? 
+ we don't need this slice at all
So, I decided to just remove this slice, it just creates a lot of complications, but we don't need it

smart-contract-verifier/src/compiler/compilers.rs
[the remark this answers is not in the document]
rimrakhimov: Get you. In that case you may just do `_ver: &compiler::Version` instead of `let _ = ver`

smart-contract-verifier/src/vyper/compiler.rs
rimrakhimov: It would be useful for users if we could provide more information about error occurred

smart-contract-verifier/src/vyper/compiler.rs
rimrakhimov: We don't need now, but that would make it more consistent with `CompilerOutput`. Otherwise, it may result in strange outputs if someone would want to use that value but find out that it is empty for vyper even though it is presented in vyper standard output specification

smart-contract-verifier/src/compiler/version.rs
rimrakhimov: It's a small detail, but, according to the specification, Prerelease should be prefixed by `-` (https://docs.rs/semver/latest/semver/struct.Prerelease.html). Current regex allow to specify prerelease part without that prefix. Your second suggestion also does not fix that case.

smart-contract-verifier/src/compiler/version.rs
rimrakhimov: Okay, let's leave it as a string

smart-contract-verifier/src/compiler/version.rs
sevenzing: `{:/v?/}{}.{}.{}{:/(-.*)?/}`

smart-contract-verifier/src/compiler/compilers.rs
```diff
@@ -20,16 +24,33 @@ pub enum Error {
     Compilation(Vec<String>),
 }
 
-pub struct Compilers {
+pub trait EvmCompilerAgent {
```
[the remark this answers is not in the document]
sevenzing: I would argue with that but okay...

smart-contract-verifier/src/vyper/compiler.rs
sevenzing: we won't repear vyper module prefix, since this module is private

if we rename this, we have to rename VyperCompilerOutput also, then we will have a lot of name collisions, and write all solidity errors and outputs like `ethers_solc::artifacts::Error`. to be homogeneous we will have to write this prefix for ethers_solc::artifacts::Severity, ethers_solc::artifacts::Contract and so on.

I tried to write this way, it's just an unreadable mess

smart-contract-verifier/src/vyper/compiler.rs
sevenzing: how would you suggest to map vyper SourceLocation to Solidity? 
Vyper: 
image


Solidity: image

I still think that this is something we don't want to

smart-contract-verifier/src/vyper/compiler.rs
sevenzing: `contracts` field is just Map to AST struct. How would you suggest to map Vyper AST to Solidity AST?

rimrakhimov: Hmm, okay... Just thought that it was a usual BTreeMap that maps source file to its string representation

smart-contract-verifier/src/vyper/compiler.rs
rimrakhimov: My suggestion is to set `lineno` value to both `start` and `end`. But probably we could not do that now, I do not insist

rimrakhimov (approved):

rimrakhimov:
Should close compilation of Vyper contracts and addition of new HTTP endpoints from #100

smart-contract-verifier/src/vyper/compiler.rs
sevenzing: the thing is, start is not related to line number, it's first char in **one-line** code. so, it just makes no sense to put it there

smart-contract-verifier/src/compiler/version.rs
```diff
@@ -10,50 +11,58 @@ pub enum ParseError {
...
 }
 
 impl FromStr for ReleaseVersion {
     type Err = ParseError;
 
     /// Parses release version from string formated as
-    /// `(v)*VERSION*+commit.*COMMITHASH*`, examples:
+    /// `(v)*VERSION*(-*PRERELEASE*)+commit.*COMMITHASH*`, examples:
```
leviska, line 28: \> release version
\> prerelease

:thinking:

leviska (approved):

PR #122, opened 2022-09-13
Add async compilation with semaphore

### LINKED ISSUE #110: Make solc compilation async
Currently we use `Solc::compile` function to compile solidity contracts. But that function compiles contracts in synchronous blocking way. Probably we should use `Solc::async_compile`

sevenzing:
Closes #110


smart-contract-verifier/src/compiler/compilers.rs
```diff
@@ -67,13 +72,16 @@ where
...
         };
+
         let output = {
-            let _timer = metrics::COMPILE_TIME.start_timer();
             let span = tracing::debug_span!(
                 "compile contract with ethers-solc",
                 ver = compiler_version.to_string()
             );
-            let _guard = span.enter();
+            let _span_guard = span.enter();
```
leviska, line 81: let's add another metric which will count how long are we waiting for the queue

smart-contract-verifier/src/compiler/compilers.rs
```diff
@@ -39,17 +42,19 @@ pub struct Compilers<C> {
     cache: DownloadCache,
     fetcher: Arc<dyn Fetcher>,
     evm_compiler: C,
+    lock: Arc<Semaphore>,
```
leviska, line 45: that's not a lock, it's misleading. lock is usually a single owner.
it's best imo to call semaphores with what kind of resource it manages. in our case it's threads, so we could call this `threads` or something similar, and then it would look like `self.threads.acquire()`

smart-contract-verifier/src/compiler/compilers.rs
```diff
@@ -141,7 +149,8 @@ mod tests {
                 let fetcher = ListFetcher::new(url, temp_dir(), None, None)
                     .await
                     .expect("Fetch releases");
-                let compilers = Compilers::new(Arc::new(fetcher), SolidityCompiler::new());
+                let lock = Arc::new(Semaphore::new(1));
```
leviska, line 152: I think we can allow some multithreading in tests too. may be make it default (threads count) or just a larger number (2-4)?

smart-contract-verifier/src/vyper/compiler.rs
```diff
@@ -105,7 +105,8 @@ mod tests {
                 let fetcher = ListFetcher::new(url, PathBuf::from("compilers"), None, None)
                     .await
                     .expect("Fetch releases");
-                let compilers = Compilers::new(Arc::new(fetcher), VyperCompiler::new());
+                let lock = Arc::new(Semaphore::new(1));
```
leviska, line 108: may be increase here too

smart-contract-verifier/src/metrics.rs
```diff
@@ -23,4 +25,28 @@ lazy_static! {
         "contract compilation time in seconds",
     )
     .unwrap();
+    pub static ref COMPILE_THREADS_NUMBER: Gauge = register_gauge!(
```
leviska, line 28: that's very misleading name
I firstly thought that this is just constant and is equal to the config value
let's call this `COMPILES_IN_FLIGHT` and add description of `number of compilations currently running`

smart-contract-verifier/src/compiler/compilers.rs
leviska: and another for how many waiters are there (similar as `COMPILES_IN_FLIGHT`, make it `COMPILES_IN_QUEUE`)

smart-contract-verifier/src/compiler/compilers.rs
sevenzing: will change it to `threads_semaphore`

smart-contract-verifier/src/compiler/compilers.rs
sevenzing: 4 metrics for compilation :)

leviska (approved):

smart-contract-verifier-http/tests/settings.rs
```diff
@@ -1,23 +1,28 @@
...
-#[cfg(target_os = "linux")]
-fn rewrite_os_specific_example_settings(
-    _example_settings: &mut Settings,
-    _default_settings: &Settings,
-) {
-}
-
-// For other systems we just use the values from default settings.
-#[cfg(not(target_os = "linux"))]
 fn rewrite_os_specific_example_settings(
```
rimrakhimov, line 4:
```suggestion
fn rewrite_system_specific_example_settings(
```
That is not _os_specific_ rewritings, as, for example, `max_threads` are not specific to any OS

smart-contract-verifier/Cargo.toml
```diff
@@ -14,7 +14,7 @@ chrono = "0.4"
 cron = "0.11"
 ethabi = "17.2"
 ethers-core = "0.17"
-ethers-solc = { version = "0.17", features = ["svm-solc"] }
+ethers-solc = { version = "0.17", features = ["svm-solc", "async"] }
```
rimrakhimov, line 17:
```suggestion
ethers-solc = { version = "0.17", features = ["async"] }
```
Nit: actually we can remove `svm-solc` feature

smart-contract-verifier-http/tests/settings.rs
```diff
@@ -1,23 +1,28 @@
...
     default_settings: &Settings,
 ) {
-    // For now, only server address is os system dependant
-    example_settings.server.addr = default_settings.server.addr;
+    example_settings.compilers.max_threads = default_settings.compilers.max_threads;
+    #[cfg(not(target_os = "linux"))]
+    {
+        use std::mem::discriminant;
+
+        example_settings.solidity.compilers_dir = default_settings.solidity.compilers_dir.clone();
```
rimrakhimov, line 13: Nit: could you please add commented default `compilers_dir` for macos in `base.toml` and `base.env`, as it is currently done for `list_url`

smart-contract-verifier/src/compiler/compilers.rs
```diff
@@ -23,6 +24,8 @@ pub enum Error {
     Internal(#[from] SolcError),
     #[error("Compilation error: {0:?}")]
     Compilation(Vec<String>),
+    #[error("failed to acquire lock: {0}")]
+    AcquireError(#[from] AcquireError),
```
rimrakhimov, line 28:
```suggestion
    Acquire(#[from] AcquireError),
```

smart-contract-verifier/src/metrics.rs
```diff
@@ -23,4 +25,38 @@ lazy_static! {
         "contract compilation time in seconds",
     )
     .unwrap();
+    pub static ref COMPILES_IN_FLIGHT: Gauge = register_gauge!(
```
rimrakhimov, line 28: Maybe `COMPILATIONS_IN_FLIGHT` (the same for below metrics)? It seems that we are using nouns for metrics, while _compile_ is definitely a verb. 

It seems that `COMPILE_TIME` above is also wrong and should be `COMPILATION_TIME` but I believe it is already too late to fix in that PR as it is already in use. But one incorrect value _should not_ be a reason to continue using invalid words, imo

sevenzing: agree to change to COMPILATIONS_IN_FLIGHT, but COMPILE_TIME is ok for me, I don't know why you think it's wrong

rimrakhimov (approved):

PR #126, opened 2022-09-19
Smart contract verifier descriptive errors

### LINKED ISSUE #124: Extend contract verification errors
Currently, if provided source contracts were successfully compiled, but none of them corresponded to the actual `creation_tx_input` and `deployed_bytecode` we return a verification error with the following message: "No contract could be verified with provided data".

We need to analyze what more specific errors we can return for different cases. As one of the examples, if any of the resultant error are [`CompilerVersionMismatch`](https://github.com/blockscout/blockscout-rs/blob/main/smart-contract-verifier/src/verifier/errors.rs#L41), the compiler version provided by the user differs from the compiler version the contract on chain was compiled with. In that case we may return "Invalid compiler version (expected ...)" instead of "No contract could be..."

rimrakhimov:
Closes #124


smart-contract-verifier/src/verifier/errors.rs
```diff
@@ -38,7 +38,7 @@ pub enum VerificationErrorKind {
     #[error("cannot parse metadata")]
     MetadataParse(String),
     #[error("compiler versions included into metadata hash does not match: {0:?}")]
-    CompilerVersionMismatch(Mismatch<Option<String>>),
+    CompilerVersionMismatch(Mismatch<Option<semver::Version>>),
```
sevenzing, line 41: Can we remove Option here? I think you did it because of Option<Version> in base_verifier.rs, I propose this way:

`base_verifier.rs`
image

smart-contract-verifier-http/tests/solidity_multiple.rs
```diff
@@ -344,6 +344,19 @@ mod failure_tests {
...
+    async fn returns_compiler_version_mismatch() {
+        let contract_dir = "solidity_0.5.14";
+
+        // Another version
+        let test_input = TestInput::new("A", "v0.5.15+commit.6a57276f");
+        test_failure(contract_dir, test_input, "Invalid compiler version").await;
+
+        // // Another nightly version
+        // let test_input = TestInput::new("A", "v0.5.14-nightly.2019.11.18+commit.79af19db");
+        // test_failure(contract_dir, test_input, "Invalid compiler versioa").await;
```
leviska, line 358: why is it commented?

leviska (approved):

leviska (approved):

sevenzing (approved):

PR #135, opened 2022-10-06
fix warnings

leviska:


smart-contract-verifier-http/tests/solidity_multiple.rs
```diff
@@ -35,16 +35,6 @@ async fn global_app_router() -> &'static AppRouter {
...
 
-async fn local_app_router(mut settings: Settings) -> AppRouter {
-    settings.sourcify.enabled = false;
-    settings.metrics.enabled = false;
-    settings.vyper.enabled = false;
-    settings.jaeger.enabled = false;
-    AppRouter::new(settings)
-        .await
-        .expect("couldn't initialize the app")
-}
```
rimrakhimov, line 46: I believe it is better not to remove that function, and add `_` prefix instead (`_local_app_router`). The point is that though that function is not used right now, we still have a test case that uses that function (https://github.com/blockscout/blockscout-rs/blob/leviska/fix_warning/smart-contract-verifier-http/tests/solidity_multiple.rs#L383). It fails right now, and is commented therefore, but it still describes the problem needed to be solved. And if we remove `local_app_router` definition, that test case would point on something non-existent

Issue #128, opened 2022-09-19
Yul contracts verification


rimrakhimov:
Research whether we can add a verification of Yul contracts (https://docs.soliditylang.org/en/v0.8.17/yul.html).

It seems, that there is no Abi available in that case, but maybe there is some way to obtain it. Or at least, we have to get an idea what are implications of that.

rimrakhimov:
Closed in #136

PR #136, opened 2022-10-09
Add basic Yul verification capability

rimrakhimov:
Add verification of Yul contracts which do not have constructor arguments.

**Breaking**: make ABI field in verification response optional, and return 'null' if Yul contract was verified


leviska (approved):
may be add a test with several yul or yul/solidity files, but as of now seems good

rimrakhimov:
Close #128

Issue #133, opened 2022-10-03
Signatures aggregation service


rimrakhimov:
Currently there are several solutions that implement Ethereum signature databases and provides an API to search for the functions and events by their selectors (https://sig.eth.samczsun.com/, https://www.4byte.directory/). Those databases may contain very similar but not identical information. We would like to implement a service that a) makes a call to those databases to search for corresponding methods, aggregates responses and return aggregated data to the caller; b) submits new ABIs to those databases.

rimrakhimov:
Endpoint (a) could be called from the main blockscout instance when it needs to decode a transaction of unverified contract.

rimrakhimov:
Regarding endpoint (b) we have considered 3 possible options:

1) The aggregation service is responsible only for importing ABIs into Ethereum Signature Databases. Aggregation service is called by Blockscout instance after smart-contract-verifier returns verification result.
![smart-contract-verifier-proxy-1](https://user-images.githubusercontent.com/25086020/193656665-e8434488-738b-4531-a623-58e0c74ab329.svg)

rimrakhimov:
## Endpoint (b)

### Option 2

The _aggregation service_ is responsible only for importing ABIs into Ethereum Signature Databases. But data into `signatures-aggregator` is submitted by `smart-contract-verifier` directly. The `Blockscout` instance does not know anything about `signatures-aggregator` and requires no changes at all. 

![smart-contract-verifier-proxy-2](https://user-images.githubusercontent.com/25086020/193679208-6f873d70-49b0-497c-9fa5-8c510b0669cc.svg)

rimrakhimov:
## Endpoint (b)

### Option 3

In the third option, we have another microservice named `smart-contract-verifier-proxy`. In comparison with the second option, that service allows `smart-contract-verifier` not to to be changed and allows it to be  _"pure"_ in the sense that it just accepts requests from the clients and returns responses. That approach have one interesting 
![smart-contract-verifier-proxy-3](https://user-images.githubusercontent.com/25086020/193681473-73fcb594-b3bb-41ce-ac82-f36329424c82.svg)

sevenzing:
I suggest considering a few other short names for this service:

+ eth-sig-proxy
+ sig-documenter
+ sig-storage
+ sig-provider

rimrakhimov:
@sevenzing I personally like _sig-provider_ the most, but I suggest to create a poll inside a Blockscout team regarding the naming, when the service functionality would be finalized

[...]

PR #137, opened 2022-10-10
Signature provider

leviska:
impl #133


sig-provider/src/aggregator.rs
```diff
@@ -0,0 +1,96 @@
...
+};
+use std::sync::Arc;
+
+pub struct SignatureAggregator {
+    sources: Vec<Arc<dyn SignatureProvider + Send + Sync + 'static>>,
+}
+
+impl SignatureAggregator {
+    pub fn new(
+        sources: Vec<Arc<dyn SignatureProvider + Send + Sync + 'static>>,
```
sevenzing, line 16: Maybe rename all `provider` to `source` ? I mean, name of this microservice is `sig-provider`, so in order not to confuse this, it is better to call the 4bytes and sigeth structures as the "SignatureSource"

rimrakhimov (approved):

sevenzing (approved):

PR #139, opened 2022-10-13
smart-contract-verifier: add Middleware trait

### LINKED ISSUE #132: Middleware trait for smart-contract-verifier
There are situations when we may want to make some additional work with verified contracts. Currently we have several examples when that could be useful:

1. #133 describes how we can update `smart-contract-verifier` to send a request for a new service responsible for broadcasting method ABIs to 3-rd party Ethereum signature database solutions (option 2)
2. It is highly probable that we decide to import verified contracts into the Sourcify service as part of our possible collaboration 

Besides that, there could be other extensions users may wish to add to result processing (e.g. insert results into some custom database).

rimrakhimov:
Closes #132


sevenzing (approved):

smart-contract-verifier/src/solidity/client.rs
```diff
@@ -3,12 +3,24 @@ use crate::{compiler::Compilers, middleware::Middleware, verifier::Success};
...
+    ///
+    /// [`new_arc`]: Self::new_arc
     pub fn new(compilers: Compilers<SolidityCompiler>) -> Self {
+        Self::new_arc(Arc::new(compilers))
+    }
+
+    /// Initialize new client builder. [`new`] is more ergonomic if you don't need the `Arc`.
+    ///
+    /// [`new`]: Self::new
+    pub fn new_arc(compilers: Arc<Compilers<SolidityCompiler>>) -> Self {
```
leviska, line 23: but... why... is it really to spare `Arc::new(compilers)` when calling `new`?

smart-contract-verifier/src/solidity/client.rs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Returns a [`Client`] using this builder configuration.
+    pub fn build(self) -> Client {
+        Client::new(self.compilers, self.middleware_stack)
+    }
+}
+
+pub struct Client {
+    compilers: Arc<Compilers<SolidityCompiler>>,
+    middleware_stack: Box<[Arc<dyn Middleware<Success>>]>,
```
leviska, line 58: imo it's better to have a single middleware
if somebody want several middlewares, he can create a separate middleware with an array of middlewares :D 
the idea is that you try to add some complex logic (arrays) in middleware, which can actually be different, like somebody might filter some things, or dynamically register middlewares or something. You can try to implement everything here, but you obviously fail, so in my experience it's better to have just one thing and if somebody wants an array - just implement your own middleware as array, or hashmap, or something

leviska: imagine if middleware could return "ok/not ok" (like a filter) (just for an example). then you could implement an array which call every middleware and then checks if any returned "not ok". or you could call them each one by one while they return "ok" and stop on first "not ok". implementing it inside middleware allows both behaviors

smart-contract-verifier/tests/middleware_types.rs
```diff
@@ -0,0 +1,153 @@
+pub mod solidity {
```
leviska, line 1: don't quite understand how this relates to middleware testing

is it just functions for generating requests, which can be used not only for middleware testing?

smart-contract-verifier/src/solidity/client.rs
leviska, line 58: also why is it `Box<[]>` and not just a vec?

leviska (approved):

smart-contract-verifier/src/solidity/client.rs
rimrakhimov: For now, I can see 2 main middlewares to implement: the one that submits verified contract's Abi to signature provider, and the one that submits verified contracts to Sourcify. It is quite easy to imagine that someone (we) may need to send something else to some other services, and it seems to be the most used case. Accepting an array of middlewares provides the ergonomic way to do that and removes the burden of implementing a separate structure for the most probable case from the user.

smart-contract-verifier/src/solidity/client.rs
leviska: why do we need this function? it's just the same as `new(Arc::new(compilers))`

smart-contract-verifier/src/solidity/client.rs
rimrakhimov: Another solution is to accept the only one middleware, but also implement a separate structure (e.g., `Composition`) which could be used by users to combine several middlewares. That solution has several advantages for sure (such that we can get rid of builder for clients and significantly reduce amount of code; also we could write better tests with checking that middleware is actually called for the clients, and checking that all middlewares are processed for the `Composition` struct.

But for me the structure with a client with several middlewares is more common and more expected (you can find it in [`actix`](https://actix.rs/), [`reqwest_middleware`](https://docs.rs/reqwest-middleware/latest/reqwest_middleware/), and many more)

rimrakhimov: I'm still not sure that it is better than the single middleware with `Composition` ability, though

rimrakhimov: In general, it has two advantages: 1) it shrinks the vector to occupy only the space that it fits in (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_boxed_slice) and does not store the capacity value, thus, requiring less space; 2) it indicates to users that this value is fixed and could not be modified. 

In general, both of these points are not too significant by themselves, but considering that using `Box<[]>` has no overhead, I cannot see a reason not to use that

smart-contract-verifier/src/solidity/client.rs
rimrakhimov: `new(Compilers<T>)` is written for convenience purpose, as for the most initializations users do not care about keeping a reference to compilers. `new_arc(Arc<Compilers<T>>)` is used for those cases when that reference is important (currently it is used only in tests and probably we may add a `#[cfg(test)]` directive, but I personally don't see why can't be a part of the builder for usual cases as well)

smart-contract-verifier/src/solidity/client.rs
leviska: that's just preemptive optimization and imo not needed here. you can easily shrink_to_fit the vec, and we'll save 16 bytes on pointers, which is nothing. 
if we're talking about MiCrOoPtImIzAtIoN, we can talk about shrink_to_fit actually causing the reallocation, so creating the `Box<[]>` from vec kinda has _some_ overhead, and etc etc, but we don't want to talk about that :) 
it seems that the only way to make [boxed slice is from vec](https://stackoverflow.com/a/33704684), I really don't see any reason why not just use vec. It is a standard solution 
but choose whatever you think is better

leviska (approved):

sevenzing (approved):

PR #142, opened 2022-10-17
Moved sig provider proto to a separate crate

leviska:


sevenzing (approved):

rimrakhimov:
What was the reason to extract proto file into `src` directory?

.github/workflows/test-sig-provider.yml
```diff
@@ -0,0 +1,107 @@
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+
+name: Test Suite
```
rimrakhimov, line 7: Could you please change the test suite name so that it indicates that those tests are for `sig-provider` service

sig-provider/sig-provider-proto/build.rs
```diff
@@ -24,8 +24,8 @@ fn compile(
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let gens = Box::new(GeneratorList::new(vec![
         tonic_build::configure().service_generator(),
-        Box::new(ActixGenerator::new("proto/api_config_http.yaml").unwrap()),
+        Box::new(ActixGenerator::new("src/api_config_http.yaml").unwrap()),
```
rimrakhimov, line 27: What was the reason to extract proto file into src directory?

leviska: because that's a (proto) source file :) 
idk just wanted to try and forgot to think about it

rimrakhimov: Could you move `.proto` and `.yaml` files back to the proto directory, like they are in [sol-to-uml](https://github.com/blockscout/sol-to-uml/tree/main/visualizer-server/proto) project. It is quite a common structure to have proto files separately, and I believe there is no reason to break that. 

Nit: also, when building, the compilation requires us to have swagger directory and is failing if that is not the case. That is not a big problem, but it would be better if we could create a directory automatically in case it did not exist.

leviska: done

rimrakhimov (approved):

PR #148, opened 2022-10-21
Move smart-contract-verifier

sevenzing:


leviska (approved):

smart-contract-verifier/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY --from=cache $CARGO_HOME $CARGO_HOME
+RUN cargo build --release
+
+FROM ubuntu:20.04 as run
+RUN apt-get update && apt-get install -y libssl1.1 libssl-dev ca-certificates
+
+WORKDIR /app
+ENV APP_USER=app
+RUN groupadd $APP_USER \
+    && useradd -g $APP_USER $APP_USER
```
rimrakhimov, line 25: Could you please add comments, why those lines are required

smart-contract-verifier/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+COPY --from=build /app/target/release/smart-contract-verifier-http /app/smart-contract-verifier-http
+RUN chown -R $APP_USER:$APP_USER /app
```
rimrakhimov, line 28: Could you please add comment, why this line is required

smart-contract-verifier/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN cargo chef cook --release --recipe-path recipe.json
+
+FROM chef AS build
+
+COPY . .
+COPY --from=cache /app/target target
+COPY --from=cache $CARGO_HOME $CARGO_HOME
+RUN cargo build --release
+
+FROM ubuntu:20.04 as run
```
rimrakhimov, line 19: Nit:

In cargo-chef description they use slim version of debian system in their example:
```
# We do not need the Rust toolchain to run the binary!
FROM debian:buster-slim AS runtime
```
Maybe we can also reuse that approach to decrease the size of our docker image?

rimrakhimov (commented):
Could you please extract docker-compose file into the external `smart-contract-verifier` directory and update Readme to be in line with current installation and new links (at least [Building from sources](https://github.com/blockscout/blockscout-rs/tree/lymarenkolev/move-scv/smart-contract-verifier/smart-contract-verifier-http#building-from-source) section)

smart-contract-verifier/Dockerfile
sevenzing: I tried to use `debian:buster-slim`, `distroless` and also several "small-weight" images, but vyper compiler always return error `lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found`. 

`debian:bullseye-slim` image worked, but it had the same image size as ubuntu, so I decided to leave ubuntu

rimrakhimov (approved):

PR #149, opened 2022-10-24
new api based on processing whole txs and events, not only hash

leviska:


sig-provider/sig-provider-proto/proto/sig-provider.proto
```diff
@@ -7,18 +7,37 @@ option go_package = "github.com/blockscout/blockscout-rs/sig-provider";
 service SignatureService {
   rpc CreateSignatures(CreateSignaturesRequest)
       returns (CreateSignaturesResponse) {}
-  rpc GetFunctionSignatures(GetSignaturesRequest)
-      returns (GetSignaturesResponse) {}
-  rpc GetEventSignatures(GetSignaturesRequest) returns (GetSignaturesResponse) {
-  }
+}
```
leviska, line 10: we could potentially leave old functions to be used somewhere

rimrakhimov (approved):

PR #153, opened 2022-10-26
added function decoding

leviska:


sig-provider/sig-provider-server/src/service.rs
```diff
@@ -40,8 +39,10 @@ impl AbiService for Service {
         request: tonic::Request<GetFunctionAbiRequest>,
     ) -> Result<tonic::Response<GetFunctionAbiResponse>, tonic::Status> {
         let request = request.into_inner();
+        let bytes = hex::decode(request.tx_input)
+            .map_err(|e| tonic::Status::invalid_argument(e.to_string()))?;
```
rimrakhimov, line 43: `hex::decode` cannot parse hex strings which start with "0x" prefix. I suggest to copy `ethers_core::types::Bytes` implementation which allows both options. Later we will extract that into a separate lib crate, so that it could be reused in all our services, but before that I suggest to copy it inside your service crate

sig-provider/sig-provider/src/aggregator.rs
```diff
@@ -78,3 +96,120 @@ impl SourceAggregator {
...
+            let agg = Arc::new(SourceAggregator::new(vec![source.clone()]));
+
+            let function = agg
+                .get_function_abi(&hex::decode(input).unwrap())
+                .await
+                .unwrap();
+            assert_eq!(abi, function);
+        }
+    }
+}
```
rimrakhimov, line 215: Could you please also add tests for array, and tuple (struct)?

sig-provider/sig-provider/src/lib.rs
```diff
@@ -3,3 +3,6 @@ mod sources;
 
 pub use aggregator::SourceAggregator;
 pub use sources::{fourbyte, sigeth, SignatureSource};
+
+#[cfg(test)]
+pub use sources::mock;
```
rimrakhimov, line 8: Why not to use `mockall` crate?

leviska: because I didn't know/forgot about it :D

rimrakhimov: Get it🙂 You should probably give it a try

leviska: done

rimrakhimov (approved):

rimrakhimov:
Looks nice for me. Could you please just fix the linting pipeline before merge

sevenzing (approved):

PR #161, opened 2022-11-02
Sig provider event decoding

leviska:
Implementing event log decoding by searching signatures and then trying to parse the log using given signature


sevenzing (approved):

sig-provider/sig-provider/src/aggregator.rs
```diff
@@ -129,6 +153,48 @@ fn decode_txinput(args: &[ParamType], tx_args: &[u8]) -> Option<Vec<Token>> {
     Some(decoded)
 }
 
+fn decode_log(name: String, args: &[ParamType], raw: RawLog) -> Option<(Vec<Token>, Vec<bool>)> {
+    const MAX_PERMUTATIONS: usize = 10000;
+    // this is indeed can be very long
+    // there are better ways to iterate over valid indexed permutations
+    // but right now we think this is okayish
+    for (ind, perm) in (0..raw.topics.len() - 1)
```
rimrakhimov, line 161: Could you please add a description how that algorithm works and why it does it that way? So that it will be at least clear why such a complex actions are required in the first place

rimrakhimov (commented):
I can see two problems which you may face when parsing events. 
1) Events could be anonymous (https://docs.soliditylang.org/en/v0.8.17/contracts.html#events) in which case the first topic does not correspond to the event identifier. 
2) Dynamic-length types (e.g. string, bytes) when indexed are represented as a keccak-256 of their encoded representation (https://docs.soliditylang.org/en/latest/abi-spec.html#events).

Could you please add tests for those cases, so to make it clear how the service handles both situations?

rimrakhimov (approved):

Issue #171, opened 2022-11-09
Generate entity files


rimrakhimov:
Generate entities from the database schema created (https://www.sea-ql.org/SeaORM/docs/generate-entity/sea-orm-cli/).

Depends on #170

rimrakhimov:
Closed in #180

PR #180, opened 2022-11-10
Add initial migration and entities

### LINKED ISSUE #170: Initial database migration
For now we came up with the following database schema for the `eth-bytecode-db` service: https://dbdiagram.io/d/63622918c9abfc61116fbc37

We need to set up migrations and create an initial database migration with schema above

sevenzing:
Closes #170, #171


eth-bytecode-db/eth-bytecode-db/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+## Development
+
+> all commands should be executed in `eth-bytecode-db/eth-bytecode-db` dir
+
+### Postgres
+
+```bash
+docker run -p 5432:5432 --name eth-bytecode-postgres -e POSTGRES_PASSWORD=admin -d postgres
```
rimrakhimov, line 11: When trying to run a migration against database in the docker container, it returns an error with the following message - "database \"eth-bytecode-db\" does not exist". 

I suggest to either add a description of how the database can be created inside docker container before running the migrations, or to create a database if the one does not already exiist.

eth-bytecode-db/eth-bytecode-db/README.md
```diff
@@ -0,0 +1,47 @@
...
+export DATABASE_URL=postgres://postgres:admin@localhost:5432/eth-bytecode-db
+```
+
+### Migrations
+
+1. Install `sea-orm-cli`:
+
+```bash
+cargo install sea-orm-cli
```
rimrakhimov, line 20: Another option is to build `migration` crate and run migrations from that trait as `target/debug(release)/migration up (down, fresh)`. In that case it is not required to run migrations from `eth-bytecode-db/eth-bytecode-db` directory

eth-bytecode-db/eth-bytecode-db/migration/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+        manager
+            .get_connection()
+            .execute(Statement::from_string(
+                manager.get_database_backend(),
+                st.to_string(),
+            ))
+            .await
+            .map_err(|_| DbErr::Query(RuntimeErr::Internal(st.to_string())))?;
+    }
+    Ok(())
```
rimrakhimov, line 27: It make sense to execute those statements in one transaction, so that if some statement fails, all others would not be applied. Otherwise, it may lead to inconsistencies in the database when one migration was only partially applied, and we cannot even  downgrade it

You may have a look at - https://www.sea-ql.org/SeaORM/docs/0.8.x/advanced-query/transaction/#begin--commit--rollback

rimrakhimov (approved):

PR #202, opened 2022-11-16
Draft: middleware settings approach

rimrakhimov:


smart-contract-verifier/smart-contract-verifier-http/src/settings.rs
```diff
@@ -219,6 +221,25 @@ impl Default for CompilersSettings {
     }
 }
 
+#[serde_as]
+#[derive(Debug, Clone, Default, PartialEq, Eq, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct ExtensionsSettings {
+    #[serde_as(as = "serde_with::EnumMap")]
+    solidity: Vec<Extensions>,
```
leviska, line 229: do we need a vec of extensions? Can't we make this a struct? like
```rust
struct Extensions {
  sig_provider: sig_provider::Settings,
  sourcify: sourcify::Settings,
}
```

leviska: I think we don't need to abstract this out. It's good to see a list of supported extensions

rimrakhimov: We've discussed that question in private conversation, and come up that the struct could be used as well. I personally have no preference in that case. @leviska I believe it's up to you to choose whatever you find more appropriate )

leviska:
Implemented in #217

Issue #174, opened 2022-11-09
Update smart-contract-verifier interface


rimrakhimov:
Epic: #200 

Update smart-contract-verifier interface

Need to update smart-contract verifier so that it would be possible to obtain verification specific data required for `eth-bytecode-db` service.

- [x] Add compiler settings to verification response
https://github.com/blockscout/blockscout-rs/pull/179
- [x] Add creation tx input and deployed bytecode parts (main/meta) of the source to verification response
https://github.com/blockscout/blockscout-rs/pull/205
- [x] Grpc server
    - [x] Protobuf file and server initialization
    https://github.com/blockscout/blockscout-rs/pull/222
    - [x] Solidity service implementation
    https://github.com/blockscout/blockscout-rs/pull/224
    - [x] Vyper service implementation
    https://github.com/blockscout/blockscout-rs/pull/225
    - [x] Sourcify service implementation
    https://github.com/blockscout/blockscout-rs/pull/226
- [ ] Add verification via metadata

rimrakhimov:
Interaction with contract verification service currently can be described by the following diagram:
![image](https://user-images.githubusercontent.com/25086020/200923854-a15387eb-1db2-45d9-a02d-cb41546f8ada.png)

We need to update `smart-contract-verifier` to return parsed creation transaction input and runtime bytecode, as well as to return compiler settings.

rimrakhimov:
I suggest to add the following keys in the current result:

```json5
{
  // Compiler settings used in verification
  "compiler_settings": "{ ... }",
  // Parsed creation transaction input resultant from local compilation
  "local_creation_input_parts": [
    { type: "main", "data": "0xcafe.." },
    { type: "metadata", "data": "0x625354.." },
    ...
  ]
    // Parsed deployed bytecode resultant from local compilation
  "local_deployed_bytecode_parts": [
    { type: "main", "data": "0xcafe.." },
    { type: "metadata", "data": "0x625354.." },
    ...
  ]
}
```

PR #205, opened 2022-11-17
Verifier - add bytecode parts to response

rimrakhimov:
User story: #174

### DIFF
--- a/smart-contract-verifier/smart-contract-verifier-http/README.md
+++ b/smart-contract-verifier/smart-contract-verifier-http/README.md
@@ -233,7 +233,19 @@ If verification succeeds, the service returns 200 with a success status:
     "constructor_arguments": "0xcafecafecafe",
     // (optional) contract abi (https://docs.soliditylang.org/en/latest/abi-spec.html?highlight=abi#json);
     // is `null` for Yul contracts
-    "abi": "[ { ... } ]"
+    "abi": "[ { ... } ]",
+    // (optional) creation transaction input resultant from local compilation
+    // parsed and split on Main and Meta parts. Is null for Sourcify verification.
+    "local_creation_input_parts": [
+      { "type": "main", "data": "0x1234.." },
+      { "type": "meta", "data": "0xcafe.." }
+    ],
+    // (optional) deployed bytecode resultant from local compilation
+    // parsed and split on Main and Meta parts. Is null for Sourcify verification.
+    "local_deployed_bytecode_parts": [
+      { "type": "main", "data": "0x1234.." },
+      { "type": "meta", "data": "0xcafe.." }
+    ]
   },
   // Status of 0 indicates successful verification
   "status": 0


smart-contract-verifier/smart-contract-verifier-http/tests/solidity_multiple_types.rs
```diff
@@ -1,5 +1,6 @@
 #![allow(dead_code, unused)]
 
+use smart_contract_verifier::BytecodePart;
```
sevenzing, line 3: do you need this?

smart-contract-verifier/smart-contract-verifier-http/src/verification_response.rs
```diff
@@ -3,14 +3,36 @@ use serde::{Deserialize, Serialize};
...
     pub message: String,
     pub result: Option<VerificationResult>,
     pub status: VerificationStatus,
 }
 
-#[derive(Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(tag = "type")]
+pub enum BytecodePart {
+    Main { data: DisplayBytes },
```
sevenzing, line 16: You can skip this idea but I suggest add `order` field for Main and Meta, that will represent position in local bytecode

I understand that it can be obtained from the position of part in array, but I think it will be more visual and informative and doesn't require a lot of efforts from our side

smart-contract-verifier/smart-contract-verifier-http/src/verification_response.rs
```diff
@@ -23,6 +45,11 @@ pub struct VerificationResult {
     pub abi: Option<String>,
     pub sources: BTreeMap<String, String>,
     pub compiler_settings: String,
+
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
```
sevenzing, line 49: why? I don't like the idea that service might not return field even if field is empty.

smart-contract-verifier/smart-contract-verifier-http/src/verification_response.rs
```diff
@@ -3,14 +3,36 @@ use serde::{Deserialize, Serialize};
...
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
 pub struct VerificationResponse {
     pub message: String,
     pub result: Option<VerificationResult>,
     pub status: VerificationStatus,
 }
 
-#[derive(Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(tag = "type")]
```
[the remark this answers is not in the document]
rimrakhimov: We have a test here
https://github.com/blockscout/blockscout-rs/pull/205/files#diff-c91c0999946402b4fe047f1e433609c4274379e374ba0c17628175e410e65307R215

You are right - it would be encoded as `{"type": "Main"}`

smart-contract-verifier/smart-contract-verifier-http/src/verification_response.rs
rimrakhimov: I personally prefer not to have `order` as it would make it more difficult to parse it both on the client and the server side. Besides that, it potentially opens a possibility for some indexes to be missed or to be equal (e.g, 1, 2, 2, 4), while position in an array is always consecutive

leviska:
please update readme :)

sevenzing (approved):

PR #206, opened 2022-11-17
decode events correctly

### LINKED ISSUE #165: sig-provider: decode event arguments correctly
Right now decoding event will not fail if some data was left, which can lead to wrong guesses
I suggest adding additional functions in ethabi: https://github.com/rust-ethereum/ethabi/pull/284, and then update our code using this function

leviska:
Closes #165

### DIFF
--- a/sig-provider/sig-provider-proto/build.rs
+++ b/sig-provider/sig-provider-proto/build.rs
@@ -17,6 +17,14 @@ fn compile(
         .protoc_arg("grpc_api_configuration=proto/api_config_http.yaml,output_format=yaml,allow_merge=true,merge_file_name=sig-provider")
         .bytes(["."])
         .type_attribute(".", "#[actix_prost_macros::serde]");
+
+    for path in protos.iter() {
+        println!("cargo:rerun-if-changed={}", path.as_ref().display())
+    }
+    for path in includes.iter() {
+        println!("cargo:rerun-if-changed={}", path.as_ref().display())
+    }
+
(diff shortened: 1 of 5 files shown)


sig-provider/sig-provider-proto/build.rs
```diff
@@ -17,6 +17,14 @@ fn compile(
         .protoc_arg("grpc_api_configuration=proto/api_config_http.yaml,output_format=yaml,allow_merge=true,merge_file_name=sig-provider")
         .bytes(["."])
         .type_attribute(".", "#[actix_prost_macros::serde]");
+
+    for path in protos.iter() {
+        println!("cargo:rerun-if-changed={}", path.as_ref().display())
+    }
```
rimrakhimov, line 23: Just interested, what these lines are for?

rimrakhimov (approved):

sig-provider/sig-provider-proto/build.rs
leviska: by default cargo will always rebuild the crate with custom build.rs
this stops it rebuilding if this files didn't change from the last build

Hektoreth (approved):
Hi thanku for all is hard for Contract write ?? All is done becouse i no undorstand  nothing here i see in etherscan my mining but i dont know how fonish

sevenzing:
closed. more details here: https://github.com/blockscout/blockscout-rs/issues/980

PR #216, opened 2022-11-22
Extract display bytes into a separate lib

rimrakhimov:
Extract display bytes into a separate crate. Allows to enable that functionality without having to import ethers-core crate


libs/hex-display-bytes/src/bytes.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub bytes::Bytes,
+);
+
+fn bytes_to_hex(b: &Bytes) -> String {
+    hex::encode(b.0.as_ref())
+}
+
+impl Debug for Bytes {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        write!(f, "Bytes(0x{})", bytes_to_hex(self))
```
leviska, line 28: may be use formatter functions here? https://doc.rust-lang.org/std/fmt/trait.Debug.html#examples

something like `f.debug_tuple("Bytes").field(format!("0x{}", bytes_to_hex(self)))`

rimrakhimov: Didn't get why it is better. Could you please clarify?

rimrakhimov: In that case the final result would be `'Bytes("0xcafe..")'`, but we need to get `Bytes(0xcafe)` instead

PR #217, opened 2022-11-23
Sig provider middleware

### LINKED ISSUE #176: Middleware to proxy verified contract abis to sig-provider
Implement a middleware that would share abis of verified contracts to the `sig-provider` service

leviska:
Closes #176


smart-contract-verifier/smart-contract-verifier-http/src/routers/solidity.rs
```diff
@@ -83,7 +83,12 @@ impl SolidityRouter {
             compilers_threads_semaphore,
         );
         compilers.load_from_dir(&dir).await;
-        let client = SolidityClient::new(compilers);
+        let mut client = SolidityClient::new(compilers);
+        if let Some(extensions) = settings.extensions {
+            client = client.with_middleware(
+                sig_provider_extension::SigProvider::new(extensions.sig_provider).await?,
```
leviska, line 89: there is a problem, that if we enable this for all `solidity, sourcify, vyper`, then we'll have 3 instances of the middleware == 3 connections. not sure right now what's the best way to fix that

smart-contract-verifier/smart-contract-verifier-http/config/base.toml
```diff
@@ -23,6 +23,9 @@ list_url = "https://solc-bin.ethereum.org/linux-amd64/list.json"
 ## The only required field for the s3 fetcher
 #bucket = "bucket"
 
+# [solidity.extensions.sig_provider]
+# url = "http://127.0.0.1:9000/"
```
rimrakhimov, line 27: Nit: I suggest to make it the other way around. Not to specify extensions as part of solidity, vyper, and sourcify settings, but specify solidity, vyper, and sourcify as different kinds of extensions. E.g.
```
[extensions.vyper.sig_provider]
url = "0.0.0.0:1234"

[extensions.sourcify.sig_provider]
url = "0.0.0.0:1234"
```

Grouping them together makes them independent from the actual verifier settings. It makes it easier to remove them, and seems to be more explicit for me. Furthermore, that way we extend the current settings struct without modification of any of its field struct that seems more clear way to do that.

smart-contract-verifier/smart-contract-verifier-http/src/settings.rs
```diff
@@ -219,6 +225,12 @@ impl Default for CompilersSettings {
     }
 }
 
+#[derive(Deserialize, Clone, PartialEq, Eq, Debug)]
+#[serde(deny_unknown_fields)]
+pub struct Extensions {
+    pub sig_provider: sig_provider_extension::Config,
```
rimrakhimov, line 231: I suggest to make each extension optional, while initializing it only of corresponding setting was given. That is, to make `extension: Extension` field not optional for the corresponding structs, but make each extension optional in `Extensions`. So that when new extensions would be added, we could enable only some of them

smart-contract-verifier/sig-provider-extension/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+        })
+    }
+}
+
+struct SigProviderImpl {
+    connection: Channel,
+}
+
+impl SigProviderImpl {
+    async fn create_signatures(self: Arc<Self>, abi: String) {
```
rimrakhimov, line 37: Why not a usual:
```suggestion
    async fn create_signatures(&self, abi: String) {
```
?

smart-contract-verifier/smart-contract-verifier-http/src/routers/solidity.rs
rimrakhimov: I personally cannot see ways to fix that easily for now. If you also don't I suggest to merge the PR as it is, and think about that later. Maybe we may create a separate issue for that

rimrakhimov (approved):

rimrakhimov:
@leviska Could you please also add new configuration options into `config/base.env` file

sevenzing (approved):

PR #222, opened 2022-11-24
Verifier http/grpc server initialization

rimrakhimov:
User story: #174

Defines protobuf file and corresponding http transcoding for the smart-contract-verifier. The Http API was intentionally remained the same as for the [smart-contract-verifier-http](https://github.com/blockscout/blockscout-rs/tree/main/smart-contract-verifier/smart-contract-verifier-http) implementation for backwards compatibility.

Initializes Grpc and Http server for smart-contract-verifier. Actual implementation of endpoints will be added in the next PRs for the sake of PR simplicity.


smart-contract-verifier/smart-contract-verifier-proto/proto/smart-contract-verifier.proto
```diff
@@ -0,0 +1,131 @@
...
+    /// (https://docs.soliditylang.org/en/v0.8.17/using-the-compiler.html#input-description)
+    string compiler_settings = 9;
+    /// Constructor arguments used for deploying verified contract
+    /// (exists only for creation inputs)
+    optional string constructor_arguments = 10;
+    /// Contract abi (https://docs.soliditylang.org/en/latest/abi-spec.html?highlight=abi#json);
+    /// (does not exist for Yul contracts)
+    optional string abi = 11;
+
+    message bytecode_part {
```
leviska, line 113: oneof?

rimrakhimov: We want bytecode part to be encoded as 
```
{
  "local_creation_input_parts": [
    {
        "type": "main",
        "data": "0xdafda"
    }
  ]
}
```
to be backwards compatible. Oneof seems not to be an option in that case

leviska: what's wrong with this?
```json
{ "local_creation_input_parts": [ 
  { "main": "0xdafda" }
 ] }
```

smart-contract-verifier/smart-contract-verifier-server/src/run.rs
```diff
@@ -0,0 +1,150 @@
...
+};
+use std::{net::SocketAddr, sync::Arc};
+
+pub async fn run(settings: Settings) -> Result<(), anyhow::Error> {
+    init_logs(settings.jaeger);
+
+    let solidity_verifier = settings
+        .solidity
+        .enabled
+        .then(|| Arc::new(SolidityVerifierService::default()));
```
leviska, line 27: settings aren't used here?

rimrakhimov: Actual implementation of services will be in next PRs. So, for now we just use an empty struct as a template

smart-contract-verifier/smart-contract-verifier-proto/proto/smart-contract-verifier.proto
rimrakhimov: We've just implemented it that way in https://github.com/blockscout/blockscout-rs/pull/205 for Http server, as it was easier to implement serde serialization. And I've tried to make new API to correspond to the previous one. So the question was not over consideration here.

Furthermore, in my opinion, explicit tag is better in the sense that it makes the semantic of the field clearer. In that case, it is explicitly mentioned that "main" is the value of key "type", while "0xdafda" is the actual data. While in your approach "0xdafda" seems to be a value for the key "main" and it is more difficult to understand what is happening here without the context, imo.

leviska (approved):

Issue #185, opened 2022-11-11
[Tech story] Create a basic service that fetches data from db and provides API


ulyanas:
Epic: https://github.com/blockscout/blockscout-rs/issues/184

[Tech story] Create a basic service that fetches data from db and provides API 

Need to create a template for the service that fetches data from db and provides API endpoint with data suitable for d3 graph library. 

- [x] Create barebone structure
https://github.com/blockscout/blockscout-rs/pull/223
- [x] Make two handles with number and line types with mock data
https://github.com/blockscout/blockscout-rs/pull/229
- [x] Initialize database
https://github.com/blockscout/blockscout-rs/pull/233
- [x] Initialize blockscout-db migrations and entities to be used as interface to database
- [x] Add update task which counts statistics
- [x] Make use of data from database to fetch data

PR #223, opened 2022-11-25
stats barebone structure

leviska:
Part of #185


stats/stats-proto/proto/api_config_http.yaml
```diff
@@ -0,0 +1,10 @@
+type: google.api.Service
+config_version: 3
+
+http:
+  rules:
+    - selector: blockscout.stats.v1.Stats.GetStats
+      get: /health
```
sevenzing, line 7: typo?

sevenzing (approved):

stats/stats-proto/proto/api_config_http.yaml
leviska: yeah, I've fixed it, but probably still will change it later

rimrakhimov (approved):

PR #224, opened 2022-11-26
Verifier Solidity service implementation

rimrakhimov:
User story: #174 

Adds Solidity service implementation for Grpc/Http smart-contract-verifier server.

Depends on #222


smart-contract-verifier/smart-contract-verifier-server/src/types/solidity_multi_part.rs
```diff
@@ -0,0 +1,155 @@
...
+use serde::{Deserialize, Serialize};
+use smart_contract_verifier::{
+    solidity::multi_part::{MultiFileContent, VerificationRequest},
+    Version,
+};
+use smart_contract_verifier_proto::blockscout::smart_contract_verifier::v1::VerifySolidityMultiPartRequest;
+use std::{collections::BTreeMap, ops::Deref, path::PathBuf, str::FromStr};
+
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct VerifySolidityMultiPartRequestWrapper(VerifySolidityMultiPartRequest);
```
leviska, line 12: may be just write a func?

rimrakhimov: Maybe :) What I personally don't like about just writing a function is that I couldn't come up with a good names and structure for them, so that it would be easy to understand what is converted into what. 

I know that current solution is not ideal as well. And it's quite probable that you are right and usual function would be easier to read. But I'm not ready to change it right now, as it would make it more difficult to merge together all service implementations. 

I suggest to consider rewriting that into functions again into separate PR after all smart-contract-verifier related PRs will be merged

PR #228, opened 2022-11-28
Add mismatch and solidity-metadata libs

sevenzing:


rimrakhimov (approved):

libs/solidity-metadata/src/lib.rs
```diff
@@ -0,0 +1,317 @@
...
+    pub fn from_cbor(encoded: &[u8]) -> Result<(Self, usize), minicbor::decode::Error> {
+        let mut context = DecodeContext::default();
+        let result = minicbor::decode_with(encoded, &mut context)?;
+
+        Ok((result, context.used_size))
+    }
+}
+
+#[derive(Debug, Error)]
+enum ParseMetadataHashError {
```
leviska, line 25: It's common to have just `Error` type if it's imported as root (i.e. people will use this as `solidity_metadata::Error`

leviska: also make this public?

leviska (approved):

libs/solidity-metadata/src/lib.rs
sevenzing: main function returns `minicbor::decode::Error::custom`, so you can only get Box<dyn std::error::Error> for that. For now there is no sense making this error public and therefore changing  its name

PR #229, opened 2022-11-28
stats: mock handles

leviska:
Implements basic handles with mock data, implements #185


stats/stats-proto/proto/api_config_http.yaml
```diff
@@ -0,0 +1,12 @@
+type: google.api.Service
+config_version: 3
+
+http:
+  rules:
+    - selector: blockscout.stats.v1.StatsService.GetStats
+      get: /api/v1/stats
+    - selector: blockscout.stats.v1.StatsService.GetNewBlocks
+      get: /api/v1/new_blocks
```
rimrakhimov, line 9: I suggest to use `blocks/new` syntax, as there are several charts related to blocks. That may be easier to grasp, and may implicitly allow us to add some further extensions if needed, as, for example, asking for all charts related to blocks

rimrakhimov (approved):

sevenzing (approved):

PR #235, opened 2022-12-04
Searching bytecodes in database

sevenzing:
+ Added search source logic. See function `search::partial_match::find_partial_match_contract`
+ Added 2 binaries `database_populate` and `database_search` for local debugging
+ Added 1 big integration test that requires Postgres
+ Added `justfile` to run migrations, create/remove database, run tests move convenient


eth-bytecode-db/eth-bytecode-db/entity/src/sea_orm_active_enums.rs
```diff
@@ -2,14 +2,6 @@
...
 use sea_orm::entity::prelude::*;
 
-#[derive(Debug, Clone, PartialEq, Eq, EnumIter, DeriveActiveEnum)]
-#[sea_orm(rs_type = "String", db_type = "Enum", enum_name = "part_type")]
-pub enum PartType {
-    #[sea_orm(string_value = "main")]
-    Main,
-    #[sea_orm(string_value = "metadata")]
-    Metadata,
-}
```
sevenzing, line 12: don't know why, but seaorm changed order of enums and structs for some reason

eth-bytecode-db/eth-bytecode-db/migration/src/lib.rs
```diff
@@ -12,6 +13,7 @@ impl MigratorTrait for Migrator {
         vec![
             Box::new(m20220101_000001_initial::Migration),
```
leviska, line 14: by the way, do we need separate migrations even before the first release? may be merge them into one migration? just a thought

eth-bytecode-db/eth-bytecode-db/src/bin/database_populate.rs
```diff
@@ -0,0 +1,70 @@
+use entity::sources;
```
leviska, line 1: why not make this a benchmark? may be use criterion

eth-bytecode-db/eth-bytecode-db/src/bin/database_populate.rs
```diff
@@ -0,0 +1,70 @@
...
+            let permit = semaphore.clone().acquire_owned().await.unwrap();
+            let db = db.clone();
+            join_handles.push(tokio::spawn(async move {
+                let res = generate_and_insert(db.as_ref(), 1, ContractType::Small).await;
+                drop(permit);
+                res
+            }));
+        }
+
+        for id in 10..5020 {
```
leviska, line 37: why 10... :D

eth-bytecode-db/eth-bytecode-db/src/bin/database_populate.rs
sevenzing: Maybe, but inserting 500.000 contracts still takes a lot of time. Even 10.000 takes about 30 seconds

eth-bytecode-db/eth-bytecode-db/src/bin/database_populate.rs
sevenzing: Like, ID 1 is used alredy in prev block, it could be 2 or bigger, I decided to use 10... do you want to start with 2?)

eth-bytecode-db/eth-bytecode-db/migration/src/lib.rs
sevenzing: 1. It was convenient for me to UP and DOWN migration with indexes to test my big SQL with different indexes.

2. I understand your concern about big amount of migrations, but I think if migration was commited in main, then someone (staging server?) could pull this branch and apply migrations. In this case we shouldn't rewrite this migration

rimrakhimov: I agree with @leviska. I believe we may change it as we want before the first release. There is no eth-bytecode-db service on the staging server right now and merging all migrations into the first one seems to be a good idea for me. 

It would be better to do that in a separate PR, though, when verification code would be merged as well. There are some more migrations in verification code, so probably, it would be better to merge them together

eth-bytecode-db/eth-bytecode-db/src/search/candidates.rs
```diff
@@ -0,0 +1,103 @@
...
+        .filter(bytecode_parts::Column::PartId.is_in(part_ids))
+        .filter(bytecode_parts::Column::Order.eq(0))
+        .filter(bytecodes::Column::BytecodeType.eq(r#type))
+        .select_only()
+        .column_as(bytecodes::Column::Id, QueryAs::BytecodeId)
+        .into_values::<_, QueryAs>()
+        .all(db)
+        .await?;
+
+    let bytecodes_parts = bytecodes::Entity::find()
```
leviska, line 67: can't you make this and previous request into one? probably not with the orm, yeah?

eth-bytecode-db/eth-bytecode-db/src/search/candidates.rs
```diff
@@ -0,0 +1,103 @@
...
+    C: ConnectionTrait,
+{
+    let part_ids = PartCandidate::find_by_statement(Statement::from_sql_and_values(
+        db.get_database_backend(),
+        r#"
+        SELECT *
+        FROM parts
+        WHERE 
+        $1
+        LIKE encode("data", 'hex') || '%'
```
leviska, line 92: can't you make this one using orm? is there no "encode" function? not asking to do that, just questioning the motivation

eth-bytecode-db/eth-bytecode-db/tests/test_search.rs
```diff
@@ -0,0 +1,216 @@
...
+use migration::{DbErr, MigratorTrait};
+use sea_orm::{ConnectionTrait, Database, DatabaseConnection, Statement};
+use std::{collections::HashMap, str::FromStr};
+use url::Url;
+
+pub struct TestDbGuard {
+    conn: DatabaseConnection,
+}
+
+impl TestDbGuard {
```
leviska, line 17: may be make this into library in the future

eth-bytecode-db/eth-bytecode-db/migration/src/lib.rs
leviska: yeah, let's just merge them may be before release

eth-bytecode-db/eth-bytecode-db/src/bin/database_populate.rs
leviska: you can write preparation code that will not count inside benchmark, and you can make it configurable with env variable (so you'll insert the contracts only once) :)

eth-bytecode-db/eth-bytecode-db/src/bin/database_populate.rs
leviska: ok :D

eth-bytecode-db/eth-bytecode-db/src/search/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod bytecodes_comparison;
+mod candidates;
+mod partial_match;
+
+pub use partial_match::{find_partial_match_contract, BytecodeRemote, PartialMatchContract};
```
rimrakhimov, line 5: Would the full match search be added further? I thought it should be a part of the same `search` function to look for the full match first, and then for the partial match if the previous was not successful

eth-bytecode-db/eth-bytecode-db/src/search/candidates.rs
```diff
@@ -0,0 +1,103 @@
...
+
+impl BytecodeCandidate {
+    /// Compare self with remote bytecode.
+    /// Return Ok(()) if this candidate meets the requirements
+    pub fn is_match(&self, remote_data: &Bytes) -> Result<(), CompareError> {
+        let local =
+            LocalBytecode::new(&self.parts).expect("local bytecode should contain valid metadata");
+        let result = compare(remote_data, &local);
+        if result.is_err() {
+            tracing::debug!(error = ?result, "bytecode unmatch");
```
rimrakhimov, line 22:
```suggestion
            tracing::debug!(error = ?result, "bytecode mismatch");
```

eth-bytecode-db/eth-bytecode-db/justfile
```diff
@@ -0,0 +1,33 @@
+# just is like `make` but better https://github.com/casey/just
+
+# list all recipes
+default:
+    just --list --unsorted
+
+
+start-postgres docker_name="eth-bytecode-postgres" db_name="eth_bytecode_db":
+    # we run it in --rm mode, so all data will be deleted after stopping
+    docker run -p 5432:5432 --name {{docker_name}} -e POSTGRES_PASSWORD=admin --rm -d postgres
```
rimrakhimov, line 10: Also, not exactly sure, but what if we make a default port docker container is starting to be different (e.g., 5433). The default Postgres port could be frequently occupied by the native Postgres instance running on a user's machine. By specifying some non-typical port, we may try to avoid such situations

eth-bytecode-db/eth-bytecode-db/src/search/mod.rs
sevenzing: I didn't add logic of full match, so I name this function "partial match". 
To be honest, I thought we would only have a partial match search, in next PR I can add `full_match.rs` search and combine it with `partial_match.rs` in some `any_search.rs`

eth-bytecode-db/eth-bytecode-db/justfile
```diff
@@ -0,0 +1,33 @@
+# just is like `make` but better https://github.com/casey/just
+
+# list all recipes
+default:
+    just --list --unsorted
+
+
+start-postgres docker_name="eth-bytecode-postgres" db_name="eth_bytecode_db":
```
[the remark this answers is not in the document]
sevenzing: sure

eth-bytecode-db/eth-bytecode-db/src/search/candidates.rs
sevenzing: I think there is not encode and also `like` function is only for columns, not for arbitrary string

eth-bytecode-db/eth-bytecode-db/src/bin/database_populate.rs
sevenzing: I will add this point about benchmarking to main issue

PR #236, opened 2022-12-05
added reading charts from db

leviska:


stats/stats/src/lib.rs
```diff
@@ -1,14 +1,216 @@
...
 #[cfg(test)]
 mod tests {
     use super::*;
+    use migration::MigratorTrait;
+    use pretty_assertions::assert_eq;
+    use sea_orm::{ConnectionTrait, Database};
+    use url::Url;
+
+    async fn init_db(name: &str) -> DatabaseConnection {
+        let db_url = std::env::var_os("DATABASE_URL")
```
sevenzing, line 98: change to std::env::var
https://github.com/blockscout/blockscout-rs/pull/235#discussion_r1039751140

stats/stats/src/lib.rs
```diff
@@ -1,14 +1,216 @@
...
+    data: &HashMap<String, (String, u64)>,
+    name: &str,
+) -> Result<u64, ReadError> {
+    data.get(name)
+        .map(|(_, value)| *value)
+        .ok_or_else(|| ReadError::NotFound(name.into()))
+}
+
+pub async fn get_counters(db: &DatabaseConnection) -> Result<Counters, ReadError> {
+    let data = JsonValue::find_by_statement(Statement::from_string(
```
sevenzing, line 35: I would suggest to use additional custom model instead of JsonValue:
```rust
#[derive(Debug, FromQueryResult)]
pub struct ChartData {
    name: ...,
    date: ...,
     value: ...,
}
```
https://www.sea-ql.org/SeaORM/docs/basic-crud/raw-sql/

rimrakhimov (approved):

Issue #208, opened 2022-11-20
Implement verification of smart-contracts


rimrakhimov:
Epic: #200

Implement verification of smart-contracts

Need to implement smart-contracts bytecode verification via source code

- [x] Solidity verification (multi-part, standard json)
https://github.com/blockscout/blockscout-rs/pull/237
- [x] Vyper verification 
https://github.com/blockscout/blockscout-rs/pull/237
- [ ] Metadata file verification (Solidity)
- [x] Store verification results
https://github.com/blockscout/blockscout-rs/pull/237
- [x] Sourcify verification
https://github.com/blockscout/blockscout-rs/pull/243

rimrakhimov:
Sequence diagram:

![image](https://user-images.githubusercontent.com/25086020/202919961-b9abfa9b-de3f-49fe-a148-eaff380e8bfc.png)

PR #243, opened 2022-12-08
Add sourcify verification

rimrakhimov:
User story: https://github.com/blockscout/blockscout-rs/issues/208

Adds verification of smart-contracts via sourcify. Implemented as a proxy to smart-contract-verifier's corresponding endpoint, which parses response and converts it into Source. Do not have any side effects and does not add anything into database.

Depends on #241, #237


sevenzing (approved):

eth-bytecode-db/eth-bytecode-db/tests/verification_test_helpers/mod.rs
```diff
@@ -15,33 +16,61 @@ use eth_bytecode_db::verification::{
...
 use tonic::transport::Uri;
 
 type AddIntoServiceFn<Service, GrpcT> = Arc<dyn Fn(&mut Service, GrpcT, VerifyResponse)>;
 
 #[derive(Clone)]
-pub enum VerifierServiceType<GrpcT> {
+pub enum VerifierServiceType<Request, GrpcT>
+// where GrpcT: From<Request>
+{
     Solidity {
```
leviska, line 33: why not make this as trait, with `type Request;`, and then have here just 3 `Arc<dyn Trait<Request = Mock..>>`?

rimrakhimov: Not exactly `type Request`, but updated to use a trait instead of enum

PR #244, opened 2022-12-08
Blockscout migration and entities

leviska:
User story: #185 

Initialization of blockscout migrations and entities, which right now be used in stats service, but can be used in other services too
We need this for easy mocking and ORM functionality to access blockscout main db, for updating statistics from it


libs/blockscout-db/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "blockscout-db"
+version = "5.0.0"
```
rimrakhimov, line 3: It is more like 4.x.x, isn't it?

libs/blockscout-db/entity/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "entity"
+version = "0.1.0"
```
rimrakhimov, line 3: I suggest to have the same version for all crates inside `blockscout-db`

rimrakhimov (approved):

rimrakhimov:
@leviska Could you please fix the workflow and add a description what user story that PR is related to

sevenzing (approved):

PR #245, opened 2022-12-09
changed date format, removed precision and changed handles

leviska:


stats/stats-proto/proto/stats.proto
```diff
@@ -6,36 +6,25 @@ option go_package = "github.com/blockscout/blockscout-rs/stats";
 
 service StatsService {
   rpc GetCounters(CountersRequest) returns (Counters);
-  rpc GetNewBlocks(ChartRequest) returns (ChartInt);
+  rpc GetLine(ChartRequest) returns (Chart);
```
rimrakhimov, line 9: What is the reason why this endpoint is named `GetLine` and not `GetChart`. I just cannot get the idea

leviska: because in future we might get different charts, like pie charts, error bars or something. probably should also change structure name too

may be we could change the name to "points", because some other charts could still have the same structure (bar charts, etc)

rimrakhimov: Get it. I suggest to change http endpoint to be `charts/line` in that case. And maybe to name proto rpc method as `GetLineChart`. It may be easier to get the idea why "line" word is present in that case just looking on the proto and swagger definitions

rimrakhimov: Also, I suggest to name request structures to be `GetCountersRequest` and `GetLineChartRequest` correspondingly. "Get" prefix allows us to add some other types requests for those resources if would be required

stats/stats-proto/proto/stats.proto
```diff
@@ -5,37 +5,26 @@ package blockscout.stats.v1;
...
-message PointInt {
+// All integers are encoded as strings to prevent data loss
+message Point {
   string date = 1;
-  uint64 value = 2;
+  string value = 2;
 }
 
-message ChartInt { repeated PointInt chart = 1; }
+message LineData { repeated Point chart = 1; }
```
rimrakhimov, line 20: Nit: still why not `LineChart` if we are trying to `GetLineChart`? Wouldn't it be more consistent?

rimrakhimov (approved):

sevenzing (approved):

PR #247, opened 2022-12-12
Make test helpers optional to compile

rimrakhimov:
Add "test-bins" feature and make it required for test binaries and search test to compile. That allows us not to expose modules required only for testing purposes for library users. That way, the Api provided by the library is more concise.

That is a part of refactoring, where eventually we would like to extract the common functionality from verification and search modules.

### DIFF
--- a/eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
+++ b/eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
@@ -10,14 +10,6 @@ use sea_orm::{
 };
 use std::collections::BTreeMap;
 
-#[cfg(feature = "test-bins")]
-pub async fn _insert_test_data(
-    db_client: &DatabaseConnection,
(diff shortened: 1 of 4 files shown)


eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -10,6 +10,14 @@ use sea_orm::{
 };
 use std::collections::BTreeMap;
 
+#[cfg(feature = "test-bins")]
+pub async fn _insert_test_data(
```
leviska, line 14: do you need this?

rimrakhimov: I'm not sure that we need it in this PR. But I still added it to make it explicit which function we may use to insert data to the database when testing

sevenzing (approved):

PR #250, opened 2022-12-13
Add full and any match

sevenzing:
Implement full search logic and add `find_contract` function that will try to perform "full search" and then "partial search".

In [last PR](https://github.com/blockscout/blockscout-rs/pull/235) there wasn't source files in response of searching function, this PR adds receiving files


eth-bytecode-db/eth-bytecode-db/src/search/full_match.rs
```diff
@@ -0,0 +1,59 @@
...
+pub async fn find_full_match_contract<C>(
+    db: &C,
+    remote: &BytecodeRemote,
+) -> Result<Vec<MatchContract>, anyhow::Error>
+where
+    C: ConnectionTrait,
+{
+    let candidates = find_source_candidates(db, remote).await?;
+    let mut matches = vec![];
+    for candidate in candidates.into_iter() {
```
rimrakhimov, line 15: In general, in case of "full" match only one source code must be valid. If that is not true, then it may indicate some error in our program execution (most probably, that we incorrectly filling the database on verification).

I suggest to add a `tracing::error` in case if `candidates` contains more than one item, so that we would log such situations and can check them manually.

eth-bytecode-db/eth-bytecode-db/Cargo.toml
```diff
@@ -20,6 +20,8 @@ sea-orm = { version = "0.10", features = [
     "runtime-tokio-rustls",
     "macros",
 ] }
+sea-query = {verions = "0.10"}
```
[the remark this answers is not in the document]
sevenzing: right, moreover - I don't need this library, I just forgot to delete it :/

eth-bytecode-db/eth-bytecode-db/src/search/full_match.rs
sevenzing: good point

rimrakhimov (approved):

PR #251, opened 2022-12-14
update new blocks from blockscout

leviska:
User story: #185 
Added implementation of new blocks updater


stats/stats/src/read.rs
```diff
@@ -0,0 +1,212 @@
...
+    #[error("database error {0}")]
+    DB(#[from] DbErr),
+    #[error("chart {0} not found")]
+    NotFound(String),
+}
+
+#[derive(FromQueryResult)]
+struct CounterData {
+    name: String,
+    date: NaiveDateTime,
```
rimrakhimov, line 18: Why here is `NaiveDateTime` used while in the `chart_data_int` entity it is a `Date` type?

stats/stats/src/update/new_blocks.rs
```diff
@@ -0,0 +1,389 @@
...
+
+#[derive(Default, Debug)]
+pub struct Updater {}
+
+#[async_trait]
+impl super::UpdaterTrait for Updater {
+    async fn update(
+        &self,
+        db: &DatabaseConnection,
+        blockscout: &DatabaseConnection,
```
rimrakhimov, line 36: Nit: maybe name that `blockscout_db` to make it more explicit

stats/stats/src/update/new_blocks.rs
```diff
@@ -0,0 +1,389 @@
...
+                (id.id, data)
+            }
+        };
+
+        let data = data.into_iter().map(|row| chart_data_int::ActiveModel {
+            id: Default::default(),
+            chart_id: Set(id),
+            date: Set(row.day),
+            value: Set(row.count as i64),
+            created_at: Default::default(),
```
rimrakhimov, line 97: You can make it the following way:

```suggestion
            chart_id: Set(id),
            date: Set(row.day),
            value: Set(row.count as i64),
            ..Default::default()
```

rimrakhimov (approved):

stats/stats/src/update/new_blocks.rs
leviska: the type kinda hints that, no?..
`blockscout: Db`

stats/stats/src/update/new_blocks.rs
leviska: yeah, I know, but in this case I wanted to provide all fields explicitly, so if someone adds new field, for example, he would need to look at this line and check, that this field should/shouldn't be default

also imo it's more readable, that you can see which fields are default and which are set

for tests that shouldn't be a problem, but for main logic I think it's a good idea

stats/stats/src/update/new_blocks.rs
rimrakhimov: Personally, I prefer `blockscout_db` as it seems to be more clear that it is database connection without even looking on the function arguments. But it is probably just a question of preferences

sevenzing (approved):

PR #253, opened 2022-12-15
changed counters to hashmap and added more values

leviska:
Changed return type for counters to have hashmap instead of specific fields, so we'll be able to return new charts without changing the structure
Also added a little more mock data

### DIFF
--- a/stats/stats/src/lib.rs
+++ b/stats/stats/src/lib.rs
@@ -48,7 +48,10 @@ pub async fn get_counters(db: &DatabaseConnection) -> Result<Counters, ReadError
         .map(|data| (data.name, (data.date, data.value as u64)))
         .collect();
     let counters = Counters {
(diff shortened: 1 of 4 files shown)


stats/stats/src/lib.rs
```diff
@@ -48,7 +48,10 @@ pub async fn get_counters(db: &DatabaseConnection) -> Result<Counters, ReadError
         .map(|data| (data.name, (data.date, data.value as u64)))
         .collect();
     let counters = Counters {
-        total_blocks_all_time: get_counter_from_data(&data, "total_blocks_all_time")?.to_string(),
+        counters: HashMap::from_iter([(
+            "total_blocks_all_time".into(),
+            get_counter_from_data(&data, "total_blocks_all_time")?.to_string(),
```
rimrakhimov, line 53: Nit: wouldn't it be better to use `camelCase` here as well?

rimrakhimov (approved):

PR #254, opened 2022-12-15
Split Sourcify verification response and local verification response

rimrakhimov:
Makes verification responses returned by Sourcify verification and local compilations not to be the same. That is a part of making verifier to be able to compile contracts got from sourcify metadata verification method.

Currently, we use the same structure for both responses. It creates some problems, as they are not identical and, thus, some optional fields are used instead (e.g. `local_creation_input_parts`, or `local_deployed_bytecode_parts`). We are going to add more fields to Sourcify response and, continuing for those fields to be in both verification cases, seems not to be practical. Splitting response on two different types makes it easier to add and process new exclusive fields.


Hektoreth (approved):

Hektoreth (approved):

Hektoreth (commented):
Thank you for that . But. This is no finish my stres

Hektoreth (approved):

PR #259, opened 2022-12-19
Update eth-bytecode-db proto definition

rimrakhimov:
1) Fix typo in `source_files` field (renames `sources_files` to `source_files`).
2) Remove verification via metadata, as it is not implemented for now, and is not going to be implemented in its existing form in the nearest future
3) Update VerifyResponse status to be enum


rimrakhimov:
Closed due to #263

PR #260, opened 2022-12-19
Add compiler versions endpoints

rimrakhimov:
Adds `compilers/solidity/versions` and `compiler/vyper/versions` endpoints in proto definition. Those endpoints should return compiler versions which client may use for the verification. Internally, should just proxy GET requests to the `smart-contract-verifier`


rimrakhimov:
Closed due to #263

PR #261, opened 2022-12-19
Renamed the blockscout-db entities with prefix

leviska:
Should fix the problem with cargo chef https://github.com/LukeMathWalker/cargo-chef/issues/176


libs/blockscout-db/Cargo.toml
```diff
@@ -6,5 +6,5 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-migration = { path = "./migration" }
-entity = { path = "./entity" }
+blockscout-db-migration = { path = "./blockscout-db-migration" }
+blockscout-db-entity = { path = "./blockscout-db-entity" }
```
rimrakhimov, line 10: I suggest to left paths the same, and update only the crate name:
```suggestion
blockscout-db-migration = { path = "./migration" }
blockscout-db-entity = { path = "./entity" }
```

That way we will avoid problems with the same crate names, while will not create an unnecessary prefixes

sevenzing: Agree, I think we should only change crate name

rimrakhimov (approved):

PR #262, opened 2022-12-20
Added total blocks counter

leviska:
Added implementation of updating total blocks counter

User story: #185

### DIFF
--- a/stats/stats/src/update/total_blocks.rs
+++ b/stats/stats/src/update/total_blocks.rs
@@ -1,0 +1,280 @@
+use async_trait::async_trait;
+use blockscout_db::entity::blocks;
+use chrono::NaiveDateTime;
+use entity::{chart_data_int, charts};
+use sea_orm::{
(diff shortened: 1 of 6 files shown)


stats/stats/src/update/total_blocks.rs
```diff
@@ -0,0 +1,276 @@
...
+    async fn update(
+        &self,
+        db: &DatabaseConnection,
+        blockscout: &DatabaseConnection,
+    ) -> Result<(), UpdateError> {
+        let name = "totalBlocksAllTime";
+        let data = blocks::Entity::find()
+            .column(blocks::Column::Number)
+            .column(blocks::Column::Timestamp)
+            .order_by_desc(blocks::Column::Timestamp)
```
sevenzing, line 37: I think it's better to order by block number. 

We are absolutely sure that blocks have to be sorted by block number, but ordering by timestamp is more tricky and maybe in some blockchains it's not always true.

sevenzing (approved):

rimrakhimov (approved):

PR #268, opened 2022-12-21
Connect reading part to server

leviska:
Connect reading implementation from db into the server 
User story: #185


stats/stats/src/update/mod.rs
```diff
@@ -12,6 +13,8 @@ pub trait UpdaterTrait {
         db: &DatabaseConnection,
         blockscout: &DatabaseConnection,
     ) -> Result<(), UpdateError>;
+
+    fn name(&self) -> String;
```
rimrakhimov, line 17: I believe it would be better to return reference and convert to string only where it is required

```suggestion
    fn name(&self) -> &str;
```

stats/stats/src/update/new_blocks.rs
```diff
@@ -249,7 +252,9 @@ mod tests {
         mock_blockscout(&blockscout).await;
 
         Updater::default().update(&db, &blockscout).await.unwrap();
-        let data = get_chart_int(&db, "newBlocksPerDay").await.unwrap();
+        let data = get_chart_int(&db, "newBlocksPerDay", None, None)
```
rimrakhimov, line 255: Nit: I suggest to assign `"newBlocksPerDay"` value to a const variable and just reuse it for all tests. So, if some renaming would be required, we can change only one variable value

sevenzing: Agree

stats/stats/src/read.rs
```diff
@@ -43,6 +43,7 @@ pub async fn get_counters(db: &DatabaseConnection) -> Result<Counters, ReadError
     ))
     .all(db)
     .await?;
+    dbg!(&data);
```
sevenzing, line 46: accidentally?

stats/stats/src/update/new_blocks.rs
leviska: I made "name" function for trait for that reason, but I'd like to replace it in later PR

stats/stats/src/read.rs
```diff
@@ -63,36 +67,45 @@ struct DateValue {
...
+    let data_request = if let Some(from) = from {
+        data_request.filter(chart_data_int::Column::Date.gte(from))
+    } else {
+        data_request
+    };
+    let data_request = if let Some(to) = to {
+        data_request.filter(chart_data_int::Column::Date.lte(to))
+    } else {
+        data_request
+    };
```
rimrakhimov, line 104: Nit:
```suggestion
    let mut data_request = chart_data_int::Entity::find()
        .column(chart_data_int::Column::Date)
        .column(chart_data_int::Column::Value)
        .filter(chart_data_int::Column::ChartId.eq(id.id))
        .order_by_asc(chart_data_int::Column::Date);

    if let Some(from) = from {
        data_request = data_request.filter(chart_data_int::Column::Date.gte(from))
    };
    if let Some(to) = to {
        data_request = data_request.filter(chart_data_int::Column::Date.lte(to))
    };
```

rimrakhimov (approved):

PR #273, opened 2022-12-23
eth-bytecode-db extractors: initialization

rimrakhimov:
Starts blockscout extractor implementation. Initializes crate for extractors. Initialize initial migrations, according to [schema](https://dbdiagram.io/d/639e555199cb1f3b55a20a10).

The database migrations were intentionally implemented through the sea-orm framework to being possible to apply them to any database backend supported by sea-orm. That should allow to run the extractor locally with some local in-memory database (e.g. Sqlite)

One noticeable difference with other services design is that migrations were extracted into a root `eth-bytecode-db` directory instead of being implemented inside `extractors` crate. The reason is that in that case there would be 2 crates with the same name (`migration`) in one workspace: one for migrations in `eth-bytecode-db` and one for migrations in `extractors`. To solve the problem I suggest to extract `migration` and `entity` crates to root service directory.


bitcoinfinancier (approved):

sevenzing:
seems outdated

PR #275, opened 2022-12-23
Add launcher to stats

sevenzing:
Change stats server launching according to new crate `blockscout-service-launcher`


stats/stats-server/src/service.rs
```diff
@@ -9,6 +9,7 @@ use stats_proto::blockscout::stats::v1::{
 use std::{collections::HashMap, str::FromStr};
 use tonic::{Request, Response, Status};
 
+#[derive(Clone)]
```
leviska, line 12: why?

leviska (approved):

stats/stats-server/src/service.rs
sevenzing: http_server requires Clone:
https://github.com/blockscout/blockscout-rs/blob/eec747566c75696c6ce7ab3656e751a31a3a4be5/libs/blockscout-service-launcher/src/launch.rs#L67

without Clone actix gives compile time error:
image

PR #281, opened 2022-12-25
Add all counters mock

sevenzing:


stats/stats-server/tests/counters.rs
```diff
@@ -0,0 +1,75 @@
+use reqwest_middleware::{ClientBuilder, ClientWithMiddleware};
+use reqwest_retry::{policies::ExponentialBackoff, RetryTransientMiddleware};
+use stats::tests::{init_db::init_db_all, mock_blockscout::fill_mock_blockscout_data};
+use stats_server::{stats, Settings};
+use std::collections::HashSet;
+
+fn client() -> ClientWithMiddleware {
```
leviska, line 7: Why do you need retry in tests?

stats/stats/src/counters_list.rs
```diff
@@ -0,0 +1,19 @@
+pub static TOTAL_BLOCKS: &str = "totalBlocks";
```
rimrakhimov, line 1: I suggest to move `counters_list.rs` inside `counters` module

rimrakhimov (commented):
I suggest to move `counters_list.rs` inside `chart/counters` module. Also let's rename `chart` on `charts` to be more consistent with `lines` and `counters` names

stats/stats/src/chart/counters/total_tokens.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn name(&self) -> &str {
+        counters_list::TOTAL_TOKENS
+    }
+
+    async fn create(&self, db: &DatabaseConnection) -> Result<(), DbErr> {
+        crate::chart::create_chart(
+            db,
+            self.name().into(),
+            ChartType::Counter,
+            ChartValueType::Double,
```
rimrakhimov, line 21: It should be an Int

stats/stats/src/chart/counters/total_tokens.rs
```diff
@@ -0,0 +1,44 @@
...
+    async fn update(
+        &self,
+        db: &DatabaseConnection,
+        blockscout: &DatabaseConnection,
+    ) -> Result<(), UpdateError> {
+        let chart_id = crate::chart::find_chart(db, self.name())
+            .await?
+            .ok_or_else(|| UpdateError::NotFound(self.name().into()))?;
+        // TODO: remove mock
+        insert_double_data(
```
rimrakhimov, line 35: It should be an Int

stats/stats/src/chart/mock.rs
```diff
@@ -1,3 +1,4 @@
+use crate::counters_list;
```
rimrakhimov, line 1: do we need `chart/mock.rs`? It seems that those functions are not used anywhere

stats/stats/src/read.rs
```diff
@@ -20,44 +22,63 @@ pub enum ReadError {
...
+        Data::Double => "chart_data_double",
+    };
     let data = CounterData::find_by_statement(Statement::from_string(
         DbBackend::Postgres,
-        r#"
-        SELECT distinct on (charts.id) charts.name, data.date, data.value 
-            FROM "chart_data_int" "data"
+        format!(
+            r#"
+        SELECT distinct on (charts.id) charts.name, data.date, data.value::text
```
rimrakhimov, line 55: Nit:
```suggestion
            SELECT distinct on (charts.id) charts.name, data.date, data.value::text
```

stats/stats-server/tests/counters.rs
sevenzing: Yes, as we did in smart-contract-verifier-http. Server may not be started yet

stats/stats/src/chart/counters/total_tokens.rs
sevenzing: To be honest I didn't understand meaning of total tokens. Is it number of smart contracts of total value of all tokens?

stats/stats/src/chart/mock.rs
sevenzing: For now, we didn't. But this function will be useful for mocking lines chart

stats/stats/src/chart/counters/total_tokens.rs
rimrakhimov: Hmm, I've thought that it is the total number of token contracts, but now I'm also not sure. Let ask about that.

rimrakhimov (approved):

PR #288, opened 2023-01-02
Disable metrics initialization when metrics should be disabled

rimrakhimov:
Currently metrics are enabled and wrapping middleware to http server is added even if they should be disabled.


sevenzing (approved):

sevenzing (approved):

libs/blockscout-service-launcher/src/launch.rs
```diff
@@ -63,20 +70,31 @@ where
...
-        App::new()
-            .wrap(metrics.clone())
-            .configure(configure_router(&http))
-    })
-    .bind(addr)
-    .unwrap_or_else(|_| panic!("failed to bind server"));
 
-    server.run()
+    if let Some(metrics) = metrics {
+        HttpServer::new(move || {
```
leviska, line 84: make if inside `move ||` lambda

rimrakhimov: You cannot do that. `App::new().wrap(metrics.clone()).configure(configure_router(&http))` and `App::new().configure(configure_router(&http))` are of different types and result in different http servers

PR #289, opened 2023-01-02
eth-bytecode-db - initialize database service

rimrakhimov:


eth-bytecode-db/eth-bytecode-db-server/tests/solidity_standard_json.rs
```diff
@@ -40,7 +40,7 @@ fn service() -> MockSolidityVerifierService {
 
 #[rstest]
 #[tokio::test]
-#[timeout(std::time::Duration::from_secs(30))]
+#[timeout(std::time::Duration::from_secs(60))]
```
leviska, line 43: is there a reason to add explicit timeout? you can enable it for all tests https://stackoverflow.com/questions/73131115/unit-test-timeout-in-rust

eth-bytecode-db/eth-bytecode-db-server/tests/verification_test_helpers/test_input_data.rs
```diff
@@ -1,15 +1,47 @@
...
+            .as_ref()?
+            .local_deployed_bytecode_parts
+            .iter()
+            .flat_map(|part| DisplayBytes::from_str(&part.data).unwrap().to_vec())
+            .collect::<Vec<_>>();
+        Some(DisplayBytes::from(bytes).to_string())
+    }
+}
+
+pub fn input_data_1(source_type: SourceType, match_type: MatchType) -> TestInputData {
```
leviska, line 39: may be rename the function to add some clarification about the input data?

leviska (approved):

eth-bytecode-db/eth-bytecode-db-server/tests/solidity_standard_json.rs
rimrakhimov: That was included for cases when eth-bytecode-db server could not be initialized successfully (e.g. the port number is already taken). In that case the test would not fail, but just run for some big period of time (not sure, but may be even indefinitely). Explicit timeouts make sure that those tests would run for a limited amount of time. 

I checked the link you'd shared, but I still believe explicit timeouts are better in that case, as they does not require additional arguments to be passed when running tests. Moreover, it allows to set different timeouts for different tests

PR #292, opened 2023-01-07
eth-bytecode-db - list compiler versions implementation

rimrakhimov:
Implements solidity and vyper `list_compiler_versions` requests


eth-bytecode-db/eth-bytecode-db-server/tests/compiler_versions.rs
```diff
@@ -0,0 +1,113 @@
...
+use rstest::rstest;
+use smart_contract_verifier_proto::blockscout::smart_contract_verifier::v2 as smart_contract_verifier_v2;
+use tonic::Response;
+
+async fn test_versions(
+    route: &str,
+    verifier: impl VerifierService<smart_contract_verifier_v2::ListCompilerVersionsResponse>,
+    verifier_response: smart_contract_verifier_v2::ListCompilerVersionsResponse,
+) {
+    let db_url = "sqlite::memory:";
```
sevenzing, line 20: Just to understand: 
eth_bytecode_db can start with sqlite databae?

rimrakhimov: It cannot in general. We cannot run migrations on sqlite. But for `list_compiler_versions` we do not need to initialize it, and thus sqlite is ok

leviska: imo it's best to stick to single database (postgres)
we already have tests with it, so it isn't much harder to do

eth-bytecode-db/eth-bytecode-db-server/tests/compiler_versions.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    assert_eq!(
+        verifier_response.compiler_versions, response.compiler_versions,
+        "Compiler versions mismatch"
+    );
+}
+
+#[rstest]
+#[tokio::test]
+#[timeout(std::time::Duration::from_secs(60))]
```
leviska, line 55: is there a reason to add explicit timeout here? you can just run all tests with assert https://stackoverflow.com/questions/73131115/unit-test-timeout-in-rust

leviska (approved):

eth-bytecode-db/eth-bytecode-db-server/tests/compiler_versions.rs
rimrakhimov: @leviska Why, considering that `sqlite::memory` will not put any additional burden on the users?

leviska: because they are different, something can work in one and not work in other
it also gives additional burden on developers, they might think "can this work on sqlite or do I use full postgres" without much reason to do so

PR #296, opened 2023-01-11
Add kv_storage and full/partial chart update

sevenzing:
If blockscout is indexing, we have to perform full update, therefore `full: bool` parametr for update function was introduced.

we will undestand indexing process comparing min_block in blockscout with saved min_block in our database (we will store it using `kv_storage` table)


stats/stats/src/blockscout_indexing.rs
```diff
@@ -0,0 +1,66 @@
...
+) -> Result<bool, DbErr> {
+    let min_block_blockscout = get_min_block_blockscout(blockscout).await?;
+    let min_block_saved = get_min_block_saved(db).await?;
+    tracing::info!(
+        min_block_blockscout = min_block_blockscout,
+        min_block_saved = min_block_saved,
+        "checking min block in blockscout database"
+    );
+    if min_block_blockscout < min_block_saved {
+        set_min_block_saved(db, min_block_blockscout).await?;
```
leviska, line 20: we should update saved value after we updated all the charts, or this could happen:

1. we see, that blockscout is indexing, and want to update
2. first we update the min block in our kv
3. then the service crashes for some reason
4. we restart the service and see, that blockscout doesn't have anything new
5. but in reality we didn't updated the last blocks, so we have wrong state

stats/stats/src/blockscout_indexing.rs
```diff
@@ -16,12 +16,15 @@ pub async fn is_blockscout_indexing(
...
-        Ok(false)
-    }
+    Ok(min_block_blockscout < min_block_saved)
+}
+
+pub async fn save_indexing_info(
+    blockscout: &DatabaseConnection,
+    db: &DatabaseConnection,
+) -> Result<(), DbErr> {
+    let min_block_blockscout = get_min_block_blockscout(blockscout).await?;
```
leviska, line 26: that's not ideal too
we should write here the value we've read

or:
1. we've read 1000 as min blockscout block
2. while we were updating, blockscout indexed also 500-1000 blocks. some of our updates updated that blocks, some don't
3. we've written 500 as min block, and then blockscout stopped indexing
4. some of the charts don't have 500-1000 blocks updated

leviska (approved):

PR #298, opened 2023-01-12
Add json logs

sevenzing:
Add option to launch application with json log format


libs/blockscout-service-launcher/src/settings.rs
```diff
@@ -58,17 +58,28 @@ impl Default for MetricsSettings {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum TracingFormat {
```
leviska, line 63: I think it's better to call this log format

leviska (approved):

libs/blockscout-service-launcher/src/settings.rs
sevenzing: this enum in setting `Tracing`, it will be used by `tracing.rs` file only, also we set Json format for `tracing_sunscriber`, so I think it's better to leave it as TracingFormat

PR #304, opened 2023-01-17
removed chart data type and made everything string

leviska:
Removed specific value types and make everything string, because in reading we return string values, so there is no need in this separation


rimrakhimov (approved):

sevenzing (approved):

stats/stats/justfile
```diff
@@ -17,4 +17,4 @@ generate-entities database_url="postgres://postgres:admin@localhost:5432/stats":
     DATABASE_URL={{database_url}} sea-orm-cli generate entity --lib -o entity/src
     cargo fmt --all -- --config imports_granularity=Crate
 
-restart-generate-entities: restart-postgres migrate-up generate-entities stop-postgres
+# restart-generate-entities: restart-postgres migrate-up generate-entities stop-postgres
```
sevenzing, line 20: why?

leviska: there is no "restart postgres" in this file, I've moved it into stats root dir

PR #311, opened 2023-01-19
Refactor stats charts

leviska:
Refactor stats charts, to make it easier to write new by including common functions and asking to write only the data logic

### DIFF
--- a/stats/stats/src/charts/counters/average_block_time.rs
+++ b/stats/stats/src/charts/counters/average_block_time.rs
@@ -1,5 +1,8 @@
 use crate::{
-    charts::insert::{insert_data, DateValue, DateValueDouble},
+    charts::{
+        insert::{DateValue, DateValueDouble},
+        ChartFullUpdater,
+    },
     UpdateError,
 };
 use async_trait::async_trait;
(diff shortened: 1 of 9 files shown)


stats/stats/src/charts/counters/average_block_time.rs
```diff
@@ -33,9 +40,9 @@ impl AverageBlockTime {
         .one(blockscout)
         .await?
         .map(DateValue::from)
-        .ok_or_else(|| DbErr::Custom("internal error: query returned nothing".into()))?;
+        .ok_or_else(|| UpdateError::Internal("queryw returned nothing".into()))?;
```
sevenzing, line 43: queryw -> query

stats/stats/src/charts/mod.rs
```diff
@@ -70,3 +74,72 @@ pub async fn create_chart(
...
+    async fn get_values(
+        &self,
+        blockscout: &DatabaseConnection,
+    ) -> Result<Vec<DateValue>, UpdateError>;
+
+    async fn update_with_values(
+        &self,
+        db: &DatabaseConnection,
+        blockscout: &DatabaseConnection,
+        _full: bool,
```
sevenzing, line 89: why does ChartFullUpdater need _full: bool? just to match signature of ChartUpdater?

leviska: yes, it will be easier to replace one to another, just change the trait, don't change the update function
both is fine imo, but here I kinda prioritized easier to refactor/change

sevenzing (approved):

PR #316, opened 2023-01-23
More descriptive errors

sevenzing:
Split UpdateError::DB into blockscout and stats db errors


stats/stats/src/charts/mod.rs
```diff
@@ -2,19 +2,20 @@ pub mod counters;
...
 
-use crate::metrics;
-
 #[derive(Error, Debug)]
 pub enum UpdateError {
-    #[error("database error {0}")]
-    DB(#[from] DbErr),
+    #[error("blockscout database error: {0}")]
+    BlockscoutDB(DbErr),
+    #[error("local database error: {0}")]
```
leviska, line 17: May be better "stats database error"

PR #317, opened 2023-01-23
Fix general metrics namespace and labels

leviska:
Fix actix metrics, so they will have the same name across all services with constant label pointing to specific service. It helps making one generic dashboard with parameter

### DIFF
--- a/libs/blockscout-service-launcher/src/metrics.rs
+++ b/libs/blockscout-service-launcher/src/metrics.rs
@@ -11,7 +11,7 @@ pub struct Metrics {
 impl Metrics {
     pub fn new(service_name: &str, endpoint: &str) -> Self {
         let registry = prometheus::default_registry();
-        let const_labels = HashMap::from([("service".into(), service_name.into())]);
+        let const_labels = HashMap::from([("rust_service".into(), service_name.into())]);
         let metrics_middleware = PrometheusMetricsBuilder::new("rust_microservices")
             .registry(registry.clone())
             .endpoint(endpoint)
--- a/libs/blockscout-service-launcher/src/metrics.rs
+++ b/libs/blockscout-service-launcher/src/metrics.rs
@@ -1,6 +1,6 @@
 use actix_web::{App, HttpServer};
 use actix_web_prom::{PrometheusMetrics, PrometheusMetricsBuilder};
-use std::net::SocketAddr;
+use std::{collections::HashMap, net::SocketAddr};
 
 #[derive(Clone)]
 pub struct Metrics {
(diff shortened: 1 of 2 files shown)


libs/blockscout-service-launcher/src/metrics.rs
```diff
@@ -11,12 +11,13 @@ pub struct Metrics {
 impl Metrics {
     pub fn new(service_name: &str, endpoint: &str) -> Self {
         let registry = prometheus::default_registry();
-        let metrics_middleware =
-            PrometheusMetricsBuilder::new(&(service_name.to_owned() + "_metrics"))
-                .registry(registry.clone())
-                .endpoint(endpoint)
-                .build()
-                .unwrap();
+        let const_labels = HashMap::from([("rust_service".into(), service_name.into())]);
```
rimrakhimov, line 14: Nit: maybe just:

```suggestion
        let const_labels = HashMap::from([("service".into(), service_name.into())]);
```
?

rimrakhimov (approved):

sevenzing (approved):

libs/blockscout-service-launcher/src/metrics.rs
```diff
@@ -11,12 +11,13 @@ pub struct Metrics {
...
+        let metrics_middleware = PrometheusMetricsBuilder::new("rust_microservices")
```
rimrakhimov, line 15: Nit:
```suggestion
        let metrics_middleware = PrometheusMetricsBuilder::new("rust_microservices_metrics")
```
?

rimrakhimov (approved):

libs/blockscout-service-launcher/src/metrics.rs
leviska: Prometheus will automatically add "service" label equal to kubernetes label, so it's better to have different name

libs/blockscout-service-launcher/src/metrics.rs
leviska: It's already a metric, don't see a reason why we need to add it
What else could be with this prefix? rust_microservices_???

libs/blockscout-service-launcher/src/metrics.rs
rimrakhimov: Get it. Maybe `service_name` in that case? I just don't like that we use `rust_` prefix in both the metric's name and the label

libs/blockscout-service-launcher/src/metrics.rs
rimrakhimov: Make sense! Just thought about that from the historic perspective, as we used `metrics` suffix before

PR #318, opened 2023-01-23
Add accountsGrowth line chart

sevenzing:
Add accounts growth chart


stats/stats/src/charts/lines/accounts_growth.rs
```diff
@@ -0,0 +1,86 @@
...
+    #[ignore = "needs database to run"]
+    async fn update_accounts_growth() {
+        let chart = AccountsGrowth::default();
+        simple_test_chart(
+            "update_accounts_growth",
+            chart,
+            vec![
+                ("2022-11-09", "1"),
+                ("2022-11-10", "3"),
+                ("2022-11-11", "5"),
```
leviska, line 81: do we have transactions from old accounts? or are they all from new account? if so, please add a transaction or two from old accounts

leviska (approved):

stats/stats/src/charts/lines/accounts_growth.rs
sevenzing: yes, we do have transactions from old account 0x0202020...
image

PR #321, opened 2023-01-25
Add new transactions and fix tests

sevenzing:
Add new transactions chart


stats/stats/src/charts/lines/new_txns.rs
```diff
@@ -0,0 +1,100 @@
...
+        blockscout: &DatabaseConnection,
+        last_row: Option<NaiveDate>,
+    ) -> Result<Vec<DateValue>, UpdateError> {
+        let stmnt = match last_row {
+            Some(row) => Statement::from_sql_and_values(
+                DbBackend::Postgres,
+                r#"
+                SELECT 
+                    date(b.timestamp) as date, 
+                    COUNT(*)::TEXT as value
```
leviska, line 26: does this count work due to group by?
i.e. it first makes group by, and then counts items in each group?

leviska (approved):

stats/stats/src/charts/lines/new_txns.rs
sevenzing: yes, it will work, the same as with new_blocks chart

PR #324, opened 2023-01-26
Added individual chart update schedule

leviska:


stats/config/charts.toml
```diff
@@ -1,6 +1,7 @@
 [[counters]]
 id = "totalBlocks"
 title = "Total Blocks"
+update_schedule = "0 0 0,3,6,9,12,15,18,21 * * * *"
```
sevenzing, line 4: `"0 0 */3 * * * *"`

stats/stats-server/src/charts.rs
```diff
@@ -38,8 +29,30 @@ fn new_hashset_check_duplicates<T: Hash + Eq, I: IntoIterator<Item = T>>(
...
+}
+
 impl Charts {
     pub fn new(config: Config) -> Result<Self, anyhow::Error> {
+        let ValidatedConfig {
+            charts,
+            counters_filter,
+            lines_filter,
+        } = Self::validate_config(&config)?;
+        let settings = Self::new_settings(&config);
```
sevenzing, line 45: maybe move `new_settings` logic into validate_config function, so that you don't need to split ValidatedConfig and Settings and also you don't need to assume that config is valid

stats/stats-server/src/update_service.rs
```diff
@@ -34,52 +34,72 @@ impl UpdateService {
...
+                chart.name(),
+                sleep_duration
+            );
             tokio::time::sleep(sleep_duration).await;
-            self.update().await;
+            self.update(chart.clone()).await;
+        }
+    }
+
+    pub fn force_update_all(self: Arc<Self>) {
```
sevenzing, line 80: don't we need to await for all futures here?

stats/stats-server/src/update_service.rs
```diff
@@ -34,52 +34,72 @@ impl UpdateService {
...
+        for chart in self.charts.charts.iter() {
+            {
+                let this = self.clone();
+                let chart = chart.clone();
+                tokio::spawn(async move { this.update(chart).await });
+            }
+        }
+    }
+
+    pub fn run(self: Arc<Self>, default_schedule: Schedule) {
```
sevenzing, line 90: NIT: I think it's better to move all public function to the top

stats/stats-server/src/update_service.rs
```diff
@@ -34,52 +34,72 @@ impl UpdateService {
...
+        for chart in self.charts.charts.iter() {
+            let settings = self.charts.settings.get(chart.name()).unwrap();
```
sevenzing, line 92: `.expect("chart settings should be valid")`

stats/stats-server/src/charts.rs
leviska: When I was developing this, I made everything in one function `new`. Then I decided to split it into two functions, validation and then some additional computations, like new settings. There is nothing wrong with assumption that config should be valid imo 
I don't think that putting everything in one big function is better, but idk

sevenzing (approved):

PR #327, opened 2023-01-31
Min blockscout block for each chart

leviska:
Storing min blockscout block for each chart, instead of once, so we could update charts correctly independently


stats/stats/src/charts/mod.rs
```diff
@@ -128,23 +152,33 @@ pub trait ChartUpdater: Chart {
...
                 .one(db)
                 .await
                 .map_err(UpdateError::StatsDB)?
+                .filter(|row| {
+                    if let Some(block) = row.min_blockscout_block {
+                        block == min_blockscout_block
+                    } else {
+                        false
+                    }
+                })
```
sevenzing, line 181: can't we filter on database query level?

stats/stats/src/charts/mod.rs
```diff
@@ -128,23 +152,33 @@ pub trait ChartUpdater: Chart {
...
         blockscout: &DatabaseConnection,
-        full: bool,
+        force_full: bool,
     ) -> Result<(), UpdateError> {
         let chart_id = crate::charts::find_chart(db, self.name())
             .await
             .map_err(UpdateError::StatsDB)?
             .ok_or_else(|| UpdateError::NotFound(self.name().into()))?;
-        let last_row = if full {
+        let min_blockscout_block = get_min_block_blockscout(blockscout)
```
sevenzing, line 161: I would add log here, smth like tracing::info(min_blockscout_block = min_blockscout_block, "extracted min blockscout block for chart {}", self.name())

stats/stats-server/src/server.rs
```diff
@@ -72,7 +72,7 @@ pub async fn stats(settings: Settings) -> Result<(), anyhow::Error> {
 
     tokio::spawn(async move {
         if settings.update_on_start {
-            update_service.clone().force_update_all().await;
+            update_service.clone().force_update_all(false).await;
```
sevenzing, line 75: Maybe move this `false` variable to config value with default `false`?

stats/stats/src/charts/mod.rs
leviska: We want to check if the last row equals to this. Filtering in sql might give us a row which is equal, but old
In current implementation this shouldn't work, because all rows must have the same min block, but imo filtering in sql is more bug prone

sevenzing (approved):

PR #329, opened 2023-01-31
Add new native coin transfers

sevenzing:


stats/stats/src/charts/lines/new_native_coins_transfers.rs
```diff
@@ -0,0 +1,105 @@
...
+        blockscout: &DatabaseConnection,
+        last_row: Option<NaiveDate>,
+    ) -> Result<Vec<DateValue>, UpdateError> {
+        let stmnt = match last_row {
+            Some(row) => Statement::from_sql_and_values(
+                DbBackend::Postgres,
+                r#"
+                SELECT 
+                    DATE(b.timestamp) as date,
+                    COUNT(*)::TEXT as value
```
leviska, line 26: shouldn't this calculate the sum of the coin values, not the number of transactions? 
i.e. not that "5 txs with coin transfers happened", but "132 eth was transferred in a day"

sevenzing: It should be count of transactions

leviska (approved):

Issue #328, opened 2023-01-31
Add patcher for filename in visualizer


sevenzing:
Blockscout can store filenames a little wrong, devs will fix it ASAP, but for old contracts, we have to maintain these filenames. 

Example:

![Image](https://user-images.githubusercontent.com/41516657/215746057-3f2e4ef4-3ab3-404a-85f1-dcdcd263761b.png)

`@openzeppelin/...`
`/_openzeppelin/...`

Suggested rules:
1. path should not have starting slash `/`
2. all `@openzeppelin` in paths have to be replaced with `_openzeppelin` if there is at least one _openzeppelin.

PR #330, opened 2023-02-01
Add patch for filename paths

sevenzing:
#328


rimrakhimov:
Is the following situation possible with input?

```
"src/@hello/kitty/a.sol": "content",
"src/_hello/kitty/b.sol": "content",
```

leviska (approved):

sevenzing:
> Is the following situation possible with input?
> 
> ```
> "src/@hello/kitty/a.sol": "content",
> "src/_hello/kitty/b.sol": "content",
> ```

I would say it's not possible, because in solidity you always import it like `import "@openzeppelin/contracts/...";`

I also didn't come up with an idea of how to handle not only first component of path,  however, If you suggest a way how to handle this type of import, I would discuss it.

rimrakhimov (approved):

visualizer/visualizer-server/src/solidity/types.rs
```diff
@@ -35,12 +46,37 @@ fn output_mask(field_mask: Option<FieldMask>) -> Result<OutputMask, anyhow::Erro
...
+            (path, content)
+        })
+        .collect()
+}
+
+// Should be the same as in
+// https://github.com/ethereum/sourcify/blob/d0882a5d6158d0f56e121835d79860034f072cd8/services/verification/src/services/Injector.ts#L907
+fn sanitize_path(path: &str) -> String {
+    let path = REGEX_ONLY_CHARS.replace_all(path, "_");
+    let path = REGEX_NO_DOTS.replace_all(path.as_ref(), "_");
```
rimrakhimov, line 70: Nit: what is the reason behind implementing regex strings as LazyStatic?

rimrakhimov (approved):

visualizer/visualizer-server/src/solidity/types.rs
sevenzing: to be sure that they are compiled when the service is started

PR #331, opened 2023-02-02
Total accounts and shared cache

leviska:


stats/stats/src/charts/chart.rs
```diff
@@ -0,0 +1,72 @@
...
+        force_full: bool,
+    ) -> Result<(), UpdateError>;
+}
+
+#[derive(Debug, FromQueryResult)]
+struct ChartID {
+    id: i32,
+}
+
+pub async fn find_chart(db: &DatabaseConnection, name: &str) -> Result<Option<i32>, DbErr> {
```
sevenzing, line 40: btw, can we make `find_chart()` is part of Chart trait with default impl, and call it `id()` for example? It seems for logical to me

stats/stats/src/charts/counters/total_accounts.rs
```diff
@@ -0,0 +1,69 @@
...
+impl ChartFullUpdater for TotalAccounts {
+    async fn get_values(
+        &self,
+        blockscout: &DatabaseConnection,
+    ) -> Result<Vec<DateValue>, UpdateError> {
+        let mut cache = self.cache.lock().await;
+        let data = cache
+            .get_or_update(async move { AccountsGrowth::read_values(blockscout).await })
+            .await?;
+        let data = data.into_iter().rev().take(1).collect();
```
sevenzing, line 33: I'm not sure that last value is the latest, maybe get max by date?

rimrakhimov (approved):

stats/stats/src/charts/chart.rs
leviska: we can do that, but may be in another pr? don't see a reason to do it here

sevenzing (approved):

Issue #307, opened 2022-12-22
Fail to verify SC via upload std json: OverflowKnownLength


SaigyoujiYuyuko233:
When I upload a 2MB+ json file for sc verification, there is an error ``JSON payload (4152195 bytes) is larger than allowed (limit: 2097152 bytes).``

With backend log:
```
smart-contract-verifier    | 2022-12-22T05:32:47.117254Z  WARN HTTP request{http.method=POST
http.route=/api/v1/solidity/verify/standard-json http.flavor=1.1
http.scheme=http http.host=host.docker.internal:8043
http.client_ip=172.19.0.1 http.user_agent=hackney/1.18.1 
http.target=/api/v1/solidity/verify/standard-json 
otel.name=HTTP POST /api/v1/solidity/verify/standard-json 
otel.kind="server" 
request_id=c69c1e33-ae5e-4870-8ffd-15ce3cdacd47 
exception.message=JSON payload (4152211 bytes) is larger than allowed (limit: 2097152 bytes).
exception.details=OverflowKnownLength { length: 4152211, limit: 2097152 } 
http.status_code=413 
otel.status_code="OK"}:
tracing_actix_web::middleware: Error encountered while processing the incoming HTTP request: OverflowKnownLength { length: 4152211, limit: 2097152 }
```

### Environment

* Deployment type (Manual/Docker/Docker-compose): Docker-compose
* Elixir & Erlang/OTP versions (`elixir -version`):
```
Erlang/OTP 24 [erts-12.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit:no-native-stack]
Elixir 1.13.1 (compiled with Erlang/OTP 24)
```
* Node JS version (`node -v`): v16.13.1
* Operating System: Fedora 36 w/ kernel 6.0.12-200.fc36.x86_64
* Blockscout Version/branch/commit: [5.0.0-prerelease-f218a75f](https://hub.docker.com/layers/blockscout/blockscout/5.0.0-prerelease-f218a75f/images/sha256-ac6e87c2ee7ecffff114161f43c7da1a166a9d7535b84b2e7a85535c30ade494?context=explore)
* Archive node type && version (Erigon/Geth/Nethermind/Ganache/?): Ganache

### Steps to reproduce
1. prepare a json file that is more than 2MB
2. go to smart contract verification page and choose ``via Standard Input Json``
3. upload the json file and click "verify & publish"

### Expected behaviour
[...]

PR #332, opened 2023-02-03
Add max-body-size to server settings

sevenzing:
This PR starts fix error with limited max-body-size: https://github.com/blockscout/blockscout-rs/issues/307. It updates common library for starting blockscout servers. Next PR will update lib version in smart-contract-verifier

### DIFF
--- a/libs/blockscout-service-launcher/Cargo.toml
+++ b/libs/blockscout-service-launcher/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-service-launcher"
-version = "0.5.2"
+version = "0.6.0"
 description = "Allows to launch blazingly fast blockscout rust services"
 license = "MIT"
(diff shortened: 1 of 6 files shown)


libs/blockscout-service-launcher/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-service-launcher"
-version = "0.5.1"
+version = "0.5.2"
```
rimrakhimov, line 3:
```suggestion
version = "0.6.0"
```

I would say that it is rather major update. Addition of new public field may potentially be breaking in case users accessed those fields directly (see [`stats-server`](https://github.com/blockscout/blockscout-rs/blob/main/stats/stats-server/src/settings.rs#L53))

rimrakhimov (approved):

Issue #323, opened 2023-01-26
stats: exclude current day from reading line charts


leviska:
When reading line charts, exclude the last day, because it can be partially computed and have unexpected to user value

PR #336, opened 2023-02-03
Remove last data point from lines chart

leviska:
impl #323

### DIFF
--- a/stats/stats-server/src/read_service.rs
+++ b/stats/stats-server/src/read_service.rs
@@ -73,7 +73,7 @@ impl StatsService for ReadService {
             .from
             .and_then(|date| NaiveDate::from_str(&date).ok());
         let to = request.to.and_then(|date| NaiveDate::from_str(&date).ok());
-        let data = stats::get_chart_data(&self.db, &request.name, from, to)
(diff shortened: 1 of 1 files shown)


rimrakhimov (approved):

stats/stats-server/src/read_service.rs
```diff
@@ -82,6 +82,9 @@ impl StatsService for ReadService {
                 value: point.value,
             })
             .collect();
+
+        // remove last data point, because it can be partially updated
+        data.pop();
```
sevenzing, line 87: Maybe pop only if date is current day? Like add filter data.filter(|v| v.date < now().date)

In that case we are sure that last value is partially updated, and if blockchain didn't produce something for one day, we will return prev value

leviska: When we update, our invariant is "last value can be partially updated" (that's why we always update it)
Here, you want to specify, that last value is partial only if it's equal to current day, which is untrue (for example, what if some chart updates once in two days? or service was down for a day?)

> if blockchain didn't produce something for one day
you mean if blockscout stop indexing? then, if it was stopped in the middle of the day, we still have partial updated value
if you mean that no i.e. transactions were produced, then we should fill the date with zero, and not skip it

sevenzing (approved):

PR #340, opened 2023-02-06
Launcher - update default settings. Bump to v0.6.0

rimrakhimov:


sevenzing (approved):

libs/blockscout-service-launcher/src/settings.rs
```diff
@@ -37,7 +37,7 @@ impl Default for GrpcServerSettings {
     fn default() -> Self {
         Self {
             enabled: true,
```
leviska, line 39: may be set this to false for default?

leviska (approved):

libs/blockscout-service-launcher/src/settings.rs
rimrakhimov: We may. Don't you think it will be too confusing?

rimrakhimov: Okay, maybe not. I agree

Issue #197, opened 2022-11-16
Transactions stats


rimrakhimov:
Epic: https://github.com/blockscout/blockscout-rs/issues/184

Transactions stats

Need to create an endpoint to provide transaction related statistics

- [x] Average transaction fee - the daily average amount in ~USD~ Tokens spent per transaction
https://github.com/blockscout/blockscout-rs/pull/361
- [x] Completed transactions - all transactions excluding  pending, dropped, replaced transactions
https://github.com/blockscout/blockscout-rs/pull/319
- [x] Total transactions - all transactions including pending, dropped, replaced transactions
https://github.com/blockscout/blockscout-rs/pull/312
- [x] Transaction fees - amount of tokens paid as fees per day
https://github.com/blockscout/blockscout-rs/pull/320
- [x] Daily transactions - new transactions number per day
https://github.com/blockscout/blockscout-rs/pull/321
- [x] Transactions growth - total transactions number per day
https://github.com/blockscout/blockscout-rs/pull/342

PR #342, opened 2023-02-06
Add txns growth

sevenzing:
https://github.com/blockscout/blockscout-rs/issues/197


stats/stats/src/charts/lines/txns_growth.rs
```diff
@@ -0,0 +1,130 @@
...
+            .await
+            .map_err(UpdateError::BlockscoutDB)?;
+        let last_row = get_last_row(self, chart_id, min_blockscout_block, db, full).await?;
+        let mut data = {
+            let mut cache = self.cache.lock().await;
+            cache
+                .get_or_update(async move { NewTxns::read_values(blockscout, last_row).await })
+                .await?
+        };
+        data.sort();
```
leviska, line 59: `sort_unstable`

stats/stats/src/charts/lines/txns_growth.rs
```diff
@@ -0,0 +1,130 @@
...
+        let min_data = data.first().map(|v| v.date);
+        if let Some(to) = min_data {
```
leviska, line 61: let's remove ident level by 
```rust
let to = match min_data {
  Some(to) => to,
  None => {
    tracing::warn!("new txns returned empty array");
    return Ok(());
  }
};
// the rest of the code
```

```

File: stats/stats/src/charts/lines/txns_growth.rs [hunk 3]
```diff
@@ -0,0 +1,130 @@
...
+                Some(last_data) => last_data
+                    .value
+                    .parse::<i64>()
+                    .map_err(|e| UpdateError::Internal(e.to_string()))?,
+                None => {
+                    if last_row.is_some() {
+                        tracing::warn!(
+                            chart_name = self.name(),
+                            last_row = ?last_row,
+                            "last_row is Some, but chart doesn't have any data until this row"
```
leviska, line 82: 1. this isn't something to warn about, because this is a regular thing, i.e. on clear db it's absolutely normal to have no data before
2. but I can see how this can be useful if something goes wrong, but I really think we should log "what we will do because condition", and not just "some condition": for example, log here "calculating growth from 0, because no old data was found". Ask yourself, can you interpret the log without looking at the code? Your message doesn't give any context: ok, chart doesn't have any data, so what? When in my message it's clear what we will do and why

stats/stats/src/charts/updater.rs
```diff
@@ -159,3 +104,74 @@ pub trait ChartUpdater: Chart {
...
+    force_full: bool,
+) -> Result<Option<NaiveDate>, UpdateError>
+where
+    C: Chart + ?Sized,
+{
+    let last_row: Option<NaiveDate> = if force_full {
+        tracing::info!(
+            min_blockscout_block = min_blockscout_block,
+            chart = chart.name(),
+            "last_row is None due to force override"
```
leviska, line 122: I think old message was clearer for same reason I described in previous message

stats/stats/src/tests/simple_test.rs
```diff
@@ -1,15 +1,27 @@
...
+
     chart.update(&db, &blockscout, true).await.unwrap();
-    let data = get_chart_data(&db, chart.name(), None, None).await.unwrap();
+    _simple_test_chart(&db, &chart, &expected).await;
+
+    chart.update(&db, &blockscout, false).await.unwrap();
+    _simple_test_chart(&db, &chart, &expected).await;
+}
+
+async fn _simple_test_chart(
```
leviska, line 19: `get_chart_and_assert`

stats/stats/src/tests/simple_test.rs
```diff
@@ -18,17 +30,24 @@ pub async fn simple_test_chart(test_name: &str, chart: impl Chart, expected: Vec
...
+
     counter.update(&db, &blockscout, true).await.unwrap();
-    let data = get_counters(&db).await.unwrap();
+    _simple_test_counter(&db, &counter, expected).await;
+
+    counter.update(&db, &blockscout, false).await.unwrap();
+    _simple_test_counter(&db, &counter, expected).await;
+}
+
+async fn _simple_test_counter(db: &DatabaseConnection, counter: &impl Chart, expected: &str) {
```
leviska, line 49: `get_counter_and_assert_eq`

stats/stats/src/charts/lines/txns_growth.rs
```diff
@@ -0,0 +1,130 @@
...
+            .map_err(UpdateError::StatsDB)?
+            .ok_or_else(|| UpdateError::NotFound(self.name().into()))?;
+        let min_blockscout_block = get_min_block_blockscout(blockscout)
+            .await
+            .map_err(UpdateError::BlockscoutDB)?;
+        let last_row = get_last_row(self, chart_id, min_blockscout_block, db, full).await?;
+        let mut data = {
+            let mut cache = self.cache.lock().await;
+            cache
+                .get_or_update(async move { NewTxns::read_values(blockscout, last_row).await })
```
leviska, line 56: we pass here `last_row` for growth chart, but we should provide it for new txns chart
this can lead to improper updating for new txns, for example if new txns will be updated only once two days, and growth will be updated once every day, then new txns could miss one day

"but something something we probably will never have such config" - yes, but I think we should at least make everything correct but slow, if the config is wrong

I think the easiest solution is to provide here `None` instead of last row, and schedule new txns update before txns growth

stats/stats/src/charts/lines/txns_growth.rs
```diff
@@ -0,0 +1,130 @@
...
+                .order_by_desc(chart_data::Column::Date)
+                .into_model()
+                .one(db)
+                .await
+                .map_err(UpdateError::StatsDB)?;
+            let mut starting_sum = match last_data {
+                Some(last_data) => last_data
+                    .value
+                    .parse::<i64>()
+                    .map_err(|e| UpdateError::Internal(e.to_string()))?,
```
leviska, line 76: `.map_err(UpdateError::Internal)`

stats/stats/src/charts/lines/txns_growth.rs
sevenzing: Agree, sending None and schedule before is ok for me

leviska (approved):

PR #348, opened 2023-02-09
Add ChartDependentUpdate and remove cache usage

sevenzing:
Move common logic for dependent charts to `trait ChartDependentUpdate`

### DIFF
--- a/stats/stats/src/lib.rs
+++ b/stats/stats/src/lib.rs
@@ -7,5 +7,5 @@ pub mod tests;
 pub use entity;
 pub use migration;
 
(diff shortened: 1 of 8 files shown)


stats/stats/src/lib.rs
```diff
@@ -8,4 +8,4 @@ pub use entity;
 pub use migration;
 
 pub use charts::{cache, counters, lines, Chart, UpdateError};
```
leviska, line 10: add DateValue?

leviska (approved):

PR #350, opened 2023-02-10
Add units for line charts

sevenzing:
Add `units` field for line chart, now chart response is
```
{
    "chart": [],
    "units: "string"
}
```


stats/stats-proto/proto/stats.proto
```diff
@@ -35,7 +35,10 @@ message Point {
   string value = 2;
 }
 
-message LineChart { repeated Point chart = 1; }
```
leviska, line 38: wait why do you return it in the data handle, not in the info handle? 
we have `LineChartInfo`, which contains all the meta information, I think it should go there

leviska (approved):

PR #352, opened 2023-02-10
Bump launcher to v0.6.0

rimrakhimov:


sevenzing (approved):

eth-bytecode-db/eth-bytecode-db-server/src/server.rs
```diff
@@ -94,8 +94,13 @@ pub async fn run(settings: Settings) -> Result<(), anyhow::Error> {
         service_name: "eth_bytecode_db".to_owned(),
         server: settings.server,
         metrics: settings.metrics,
-        tracing: settings.tracing,
-        jaeger: settings.jaeger,
     };
+
+    blockscout_service_launcher::init_logs(
```
leviska, line 99: I suggest init logs as early as possible (preferably first line in main)
For example, what if `db_connection` fails? It will not log anything and just panic

rimrakhimov: Great idea
Fixed that

PR #354, opened 2023-02-13
Make totalTxns dependent + add updater module

sevenzing:
+ made Total Transaction counter calculated from New Transactions line
+ moved common code in `parse_and_sum` function
+ added updater module (from updater.rs)


stats/stats/src/charts/updater/full.rs
```diff
@@ -0,0 +1,43 @@
+use crate::{
+    charts::{
+        find_chart,
+        insert::{insert_data_many, DateValue},
+    },
+    metrics, Chart, UpdateError,
+};
+use async_trait::async_trait;
+
```
leviska, line 9: remove new line

stats/stats/src/charts/counters/total_txns.rs
```diff
@@ -68,7 +67,7 @@ mod tests {
     #[tokio::test]
     #[ignore = "needs database to run"]
     async fn update_total_txns() {
-        let counter = TotalTxns::default();
-        simple_test_counter("update_total_txns", counter, "17").await;
+        let counter = TotalTxns::new(Default::default());
+        simple_test_counter("update_total_txns", counter, "16").await;
```
leviska, line 71: why did it change?

leviska (approved):

stats/stats/src/charts/counters/total_txns.rs
sevenzing: It changed because we used to count like `count(transactions)`, now we apply filter that `transaction.block.consensus = true`

PR #381, opened 2023-02-27
Handle the same sources

rimrakhimov:
Before, if the same verification request sent more than once, all of them would create a new entry in the database. Then requesting for the sources for the given bytecode returned all those entries in the result, even though they all have been identical.

The PR fixes that problem, and makes sure that only distinct sources are stored into the database

### DIFF
--- a/eth-bytecode-db/eth-bytecode-db-server/tests/verification_test_helpers/mod.rs
+++ b/eth-bytecode-db/eth-bytecode-db-server/tests/verification_test_helpers/mod.rs
@@ -253,7 +253,7 @@ pub mod test_cases {
         Service: VerifierService<smart_contract_verifier_v2::VerifyResponse>,
         Request: Serialize,
     {
-        let db = init_db(test_suite_name, "test_verify_then_search").await;
+        let db = init_db(test_suite_name, "test_verify_same_source_twice").await;
 
         let test_data = test_input_data::basic(source_type, MatchType::Full);
         let creation_input = test_data.creation_input().unwrap();
--- a/eth-bytecode-db/eth-bytecode-db-server/tests/verification_test_helpers/mod.rs
+++ b/eth-bytecode-db/eth-bytecode-db-server/tests/verification_test_helpers/mod.rs
@@ -242,4 +242,93 @@ pub mod test_cases {
             "Sources returned on verification and search differ"
         );
     }
+
+    pub async fn test_verify_same_source_twice<Service, Request>(
+        test_suite_name: &str,
+        service: Service,
+        route: &str,
+        verification_request: Request,
+        source_type: SourceType,
+    ) where
+        Service: VerifierService<smart_contract_verifier_v2::VerifyResponse>,
+        Request: Serialize,
+    {
+        let db = init_db(test_suite_name, "test_verify_then_search").await;
+
+        let test_data = test_input_data::basic(source_type, MatchType::Full);
+        let creation_input = test_data.creation_input().unwrap();
+
+        let db_url = db.db_url();
+        let verifier_addr = init_verifier_server(service, test_data.verifier_response).await;
+
+        let eth_bytecode_db_base = init_eth_bytecode_db_server(db_url, verifier_addr).await;
+
+        let response = reqwest::Client::new()
+            .post(eth_bytecode_db_base.join(route).unwrap())
+            .json(&verification_request)
+            .send()
+            .await
+            .expect("Failed to send verification request");
+
+        let verification_response: eth_bytecode_db_v2::VerifyResponse = response
+            .json()
+            .await
+            .expect("Verification response deserialization failed");
+
+        let response_2 = reqwest::Client::new()
+            .post(eth_bytecode_db_base.join(route).unwrap())
+            .json(&verification_request)
+            .send()
+            .await
+            .expect("Failed to send verification request");
+
+        let verification_response_2: eth_bytecode_db_v2::VerifyResponse = response_2
+            .json()
+            .await
+            .expect("Verification response deserialization failed");
+
+        assert_eq!(
+            verification_response, verification_response_2,
+            "Verification responses are different"
+        );
+
+        let creation_input_search_response: eth_bytecode_db_v2::SearchSourcesResponse = {
+            let request = {
+                eth_bytecode_db_v2::SearchSourcesRequest {
+                    bytecode: creation_input,
+                    bytecode_type: eth_bytecode_db_v2::BytecodeType::CreationInput.into(),
+                }
+            };
+
+            let response = reqwest::Client::new()
+                .post(eth_bytecode_db_base.join(DB_SEARCH_ROUTE).unwrap())
+                .json(&request)
+                .send()
+                .await
+                .expect("Failed to send creation input search request");
+            // Assert that status code is success
+            if !response.status().is_success() {
+                let status = response.status();
+                let message = response.text().await.expect("Read body as text");
+                panic!(
+                    "Creation input search: invalid status code (success expected). Status: {status}. Message: {message}"
+                )
+            }
+            response
+                .json()
+                .await
+                .expect("Creation input search response deserialization failed")
+        };
+
+        assert_eq!(
+            1,
+            creation_input_search_response.sources.len(),
+            "Invalid number of sources returned"
+        );
+        assert_eq!(
+            verification_response.source.unwrap(),
+            creation_input_search_response.sources[0],
+            "Sources returned on verification and search differ"
+        );
+    }
 }
--- a/eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
+++ b/eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
@@ -6,7 +6,8 @@ use entity::{
 };
 use sea_orm::{
     entity::prelude::ColumnTrait, sea_query::OnConflict, ActiveModelTrait, ActiveValue::Set,
-    DatabaseConnection, DatabaseTransaction, EntityTrait, QueryFilter, TransactionTrait,
+    ConnectionTrait, DatabaseBackend, DatabaseConnection, DatabaseTransaction, DbErr, EntityTrait,
+    QueryFilter, Statement, TransactionTrait,
 };
 use std::collections::BTreeMap;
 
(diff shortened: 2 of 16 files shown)


eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -131,21 +111,116 @@ async fn insert_source_files(
...
+    let file_ids = {
+        let mut file_ids = file_models.iter().map(|file| file.id).collect::<Vec<_>>();
+        file_ids.sort();
+        file_ids.dedup();
+        file_ids
+    };
+    // Results in `SELECT md5(array_agg(x)::text) FROM (VALUES (1), (2), (3)) t(x).
+    // Without t() `array_agg(x)::text` results in {(1),(2),(3)} instead of {1,2,3}.
+    let file_ids_hash_query = format!(
+        "SELECT md5(array_agg(x)::text)::uuid FROM (VALUES {}) t(x)",
```
leviska, line 139: You should use something like Statement::from_str_and_values instead of format

eth-bytecode-db/eth-bytecode-db-server/tests/verification_test_helpers/mod.rs
```diff
@@ -242,4 +242,93 @@ pub mod test_cases {
...
+        test_suite_name: &str,
+        service: Service,
+        route: &str,
+        verification_request: Request,
+        source_type: SourceType,
+    ) where
+        Service: VerifierService<smart_contract_verifier_v2::VerifyResponse>,
+        Request: Serialize,
+    {
+        let db = init_db(test_suite_name, "test_verify_same_source_twice").await;
```
sevenzing, line 256: not sure, maybe not just "test_verify_same_source_twice" but 

`test_suite_name + "_test_verify_same_source_twice"`

just in case two tests will run at the same time

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
leviska: Wait what why do you use sql to calculate hash(?)?

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -131,21 +111,116 @@ async fn insert_source_files(
...
+}
+
+async fn insert_source_details(
+    txn: &DatabaseTransaction,
+    source: types::Source,
+    file_models: &[files::Model],
+) -> Result<(sources::Model, bool), anyhow::Error> {
+    let abi = match source.abi {
+        None => None,
+        Some(abi) => serde_json::from_str(&abi).context("deserialize abi")?,
```
leviska, line 127: That's a .map

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -131,21 +111,116 @@ async fn insert_source_files(
...
+    };
+
+    let file_ids = {
+        let mut file_ids = file_models.iter().map(|file| file.id).collect::<Vec<_>>();
```
leviska, line 131: Collect to hashset/btreeset?

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -131,21 +111,116 @@ async fn insert_source_files(
...
+    // Without t() `array_agg(x)::text` results in {(1),(2),(3)} instead of {1,2,3}.
+    let file_ids_hash_query = format!(
+        "SELECT md5(array_agg(x)::text)::uuid FROM (VALUES {}) t(x)",
+        file_ids
+            .iter()
+            .map(|i| format!("({i})"))
+            .collect::<Vec<_>>()
+            .join(",")
+    );
+    let file_ids_hash: sea_orm::prelude::Uuid = txn
```
leviska, line 146: Just include seaorm::prelude::*

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -131,21 +111,116 @@ async fn insert_source_files(
...
+        .context("calculate hash of file ids")?;
+
+    let (source, inserted) = {
+        let compiler_settings: sea_orm::prelude::Json =
+            serde_json::from_str(&source.compiler_settings)
+                .context("deserialize compiler settings")?;
+
+        let db_source = sources::Entity::find()
+            .filter(sources::Column::CompilerVersion.eq(source.compiler_version.clone()))
+            .filter(sources::Column::CompilerSettings.eq(compiler_settings.clone()))
```
leviska, line 167: Didn't you calculate hashes specifically to remove this comparison? Or does it compare hashes?

leviska: May be I got something mixed up

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -131,21 +111,116 @@ async fn insert_source_files(
...
+            .filter(sources::Column::ContractName.eq(source.contract_name.clone()))
+            .filter(sources::Column::FileIdsHash.eq(file_ids_hash))
+            .one(txn)
+            .await
+            .context("select from \"sources\" by \"compiler_version\", \"compiler_settings\", \"file_name\", \"contract_name\", and \"file_ids_hash\"")?;
 
+        match db_source {
+            Some(source) => (source, false),
+            None => {
+                let source = sources::ActiveModel {
```
leviska, line 178: Wait wait wait
Why do you select first? Is it to check if this exists? This doesn't guarantee you nothing, becausd between select and insert anything can happen
You should make an insert and have some logic ON CONFLICT

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: We cannot apply error propagation (`?`) in the `.map` closure

rimrakhimov: Didn't know about `.transpose()`. Will refactor it, thanks

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: The hash is calculated to enforce uniqueness property, because the index data can be of a limited size only. Here we use that data instead, as we can compare the whole data without any problem. I'm not sure about performance, but it is just much easier to do using `sea-orm` framework. In case of any performance issues, I believe we may update that filtering

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: It may be a problem, I agree. I suggest to fix that in a separate PR, though, as we used that approach before this PR. I suggest to merge those changes and then to fix database data insertion

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: Don't really like that approach

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: Updated to use `Statement::from_str_and_values` though

sevenzing (approved):

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -78,39 +89,12 @@ pub(crate) async fn insert_verified_contract_data(
...
-}
-
-async fn insert_source_files(
+async fn insert_files(
     txn: &DatabaseTransaction,
-    source_model: &sources::Model,
-    source_files: BTreeMap<String, String>,
-) -> Result<(), anyhow::Error> {
-    for (name, content) in source_files {
+    files: BTreeMap<String, String>,
```
leviska, line 94: why btreemap, not hashmap?

rimrakhimov: Because the order is important in that case

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -78,39 +89,12 @@ pub(crate) async fn insert_verified_contract_data(
...
+) -> Result<Vec<files::Model>, anyhow::Error> {
+    let mut result = Vec::new();
+    for (name, content) in files {
```
leviska, line 97: files.into_iter()....?
you can collect iterator of Results into Result:
```
let result: Result<Vec<_>, _> = files.into_iter().map(|x| foo(x)?).collect();
let data = result?;
```

leviska: ah, async >.<

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: Because we use `BTreeMap` for maps everywhere for the service

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -78,39 +89,12 @@ pub(crate) async fn insert_verified_contract_data(
...
         let file = {
             let file = files::Entity::find()
```
leviska, line 99: may be (later?) make this concurent

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -131,21 +115,112 @@ async fn insert_source_files(
...
+        .transpose()?;
+
+    let file_ids: BTreeSet<_> = file_models.iter().map(|file| file.id).collect();
 
+    // Results in `SELECT md5({1,2,3}::text)
+    let file_ids_hash_query = Statement::from_sql_and_values(
+        DatabaseBackend::Postgres,
+        "SELECT md5($1::text)::uuid",
+        [sea_orm::Value::Array(
+            ArrayType::BigInt,
```
leviska, line 141: can't you make `file_ids.into_iter().map(..).collect::<Vec<_>>().into()`?
https://docs.rs/sea-orm/latest/sea_orm/enum.Value.html#impl-From%3CVec%3CT%2C%20Global%3E%3E-for-Value

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -131,21 +115,112 @@ async fn insert_source_files(
...
+    file_models: &[files::Model],
+) -> Result<(), anyhow::Error> {
+    for file in file_models {
         let source_file = source_files::ActiveModel {
             source_id: Set(source_model.id),
             file_id: Set(file.id),
             ..Default::default()
         };
-        source_files::Entity::insert(source_file)
+        let result = source_files::Entity::insert(source_file)
```
leviska, line 212: why not 
```
let values = file_models.into_iter().map(|x| soure_files::ActiveModel ...);
source_files::Entity::insert_many(values)
```
something with error?

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: Need to think about that, but for now it seems as a nice idea

PR #383, opened 2023-03-01
eth-bytecode-db - fix insert then select

rimrakhimov:


eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -15,6 +15,47 @@ use sea_orm::{
 };
 use std::collections::{BTreeMap, BTreeSet};
 
+macro_rules! insert_then_select {
+    ( $txn:expr, $entity_module:ident, $active_model:expr, [ $( ($column:ident, $value:expr) ),+ $(,)? ] ) => {
```
leviska, line 19: I think? this can be done without macro
https://docs.rs/sea-orm/latest/sea_orm/entity/trait.EntityTrait.html
but it will probably require to redo this whole thing, so it's okay for now

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
```diff
@@ -95,26 +136,14 @@ async fn insert_files(
...
-                .insert(txn)
-                .await
-                .context("insert into \"files\"")?,
-            }
+        let active_model = files::ActiveModel {
+            name: Set(name.clone()),
+            content: Set(content.clone()),
+            ..Default::default()
         };
+        let (file, _inserted) =
```
leviska, line 144: Nit: let (file, _) ?

leviska (approved):

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: Would it be possible to use filters in that case?

eth-bytecode-db/eth-bytecode-db/src/verification/db.rs
rimrakhimov: Prefer the version with direct indication what the second argument returns

PR #391, opened 2023-03-08
Initialalize blockscout-auth common crate

sevenzing:
This pr added a new crate called "blockscout-auth" in libs directory. This crate responsible for authenticate blockscout user requests from rust services

### DIFF
--- a/libs/blockscout-auth/src/lib.rs
+++ b/libs/blockscout-auth/src/lib.rs
@@ -1,11 +1,9 @@
 use cookie::Cookie;
-use serde::Serialize;
 use std::collections::HashMap;
 use thiserror::Error;
-use tonic::{codegen::http::header::COOKIE, metadata::MetadataMap, Request};
+use tonic::{codegen::http::header::COOKIE, metadata::MetadataMap};
 
 const JWT_TOKEN_NAME: &str = "_explorer_key";
-const CSRF_TOKEN_NAME: &str = "_csrf_token";
 
 #[derive(Debug)]
 pub struct AuthSuccess {
@@ -14,24 +12,23 @@ pub struct AuthSuccess {
 
 #[derive(Error, Debug, PartialEq, Eq)]
 pub enum Error {
-    #[error("invalid data: {0}")]
-    InvalidData(String),
+    #[error("invalid jwt token: {0}")]
+    InvalidJwt(String),
+    #[error("invalid cstf token: {0}")]
+    InvalidCsrf(String),
     #[error("user is unauthorized: {0}")]
     Unauthorized(String),
     #[error("blockscout invalid response: {0}")]
-    BlockscoutApiError(String),
+    BlockscoutApi(String),
 }
 
-pub async fn auth_from_tonic<T: Serialize>(
-    request: Request<T>,
+pub async fn auth_from_tonic(
+    metadata: &MetadataMap,
+    csrf_token: Option<&str>,
     blockscout_host: &str,
 ) -> Result<AuthSuccess, Error> {
-    let jwt = extract_jwt(request.metadata())?;
-    let csrf_token = serde_json::to_value(request.into_inner())
-        .map_err(|e| Error::InvalidData(format!("invalid request payload: {e}")))?
-        .get(CSRF_TOKEN_NAME)
-        .and_then(|token| token.as_str().map(|s| s.to_string()));
-    auth_from_tokens(jwt.as_ref(), csrf_token.as_deref(), blockscout_host).await
+    let jwt = extract_jwt(metadata)?;
+    auth_from_tokens(jwt.as_ref(), csrf_token, blockscout_host).await
 }
 
 pub async fn auth_from_tokens(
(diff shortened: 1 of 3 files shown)


libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+
+#[derive(Debug)]
+pub struct AuthSuccess {
+    pub user_id: String,
+}
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum Error {
+    #[error("invalid data: {0}")]
+    InvalidData(String),
```
rimrakhimov, line 18: I suggest to create two errors instead `InvalidJwtToken` and `InvalidCsrfToken`, so that using crates could differentiate between those cases if would be needed.

rimrakhimov, line 18: I suggest to create two errors instead `InvalidJwtToken` and `InvalidCsrfToken`, so that using crates could differentiate between those cases if would be needed.

libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+    #[error("user is unauthorized: {0}")]
+    Unauthorized(String),
```
rimrakhimov, line 20:
```suggestion
    Unauthorized(String),
```
```suggestion
    #[error("user is unauthenticated: {0}")]
    Unauthenticated(String),
```

libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+        .get(COOKIE.as_str())
+        .map(|value| value.to_str())
+        .ok_or_else(|| Error::InvalidData("no cookies were provided".to_string()))?
+        .map_err(|e| Error::InvalidData(format!("invalid cookie format: {e}")))?;
+    let cookies = Cookie::split_parse_encoded(cookies_raw)
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|e| Error::InvalidData(format!("cannot parse cookie: {e}")))?
+        .into_iter()
+        .map(|c| (c.name().to_string(), c))
+        .collect();
```
rimrakhimov, line 70: You can collect decrease number of places where collect occurs to one:
```suggestion
        .map(|val| {
            val.map(|c| (c.name().to_string(), c))
                .map_err(|e| Error::InvalidData(format!("cannot parse cookie: {e}")))
        })
        .collect::<Result<_, _>>()?;
```

libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+        .ok_or_else(|| Error::InvalidData(format!("'{JWT_TOKEN_NAME}' not found in request")))?;
+    Ok(token.to_string())
+}
+
+fn get_cookies(metadata: &MetadataMap) -> Result<HashMap<String, Cookie>, Error> {
+    let cookies_raw = metadata
+        .get(COOKIE.as_str())
+        .map(|value| value.to_str())
+        .ok_or_else(|| Error::InvalidData("no cookies were provided".to_string()))?
+        .map_err(|e| Error::InvalidData(format!("invalid cookie format: {e}")))?;
```
rimrakhimov, line 64: Nit: may be a little bit clearer:

```suggestion
    let cookies_raw = metadata
        .get(COOKIE.as_str())
        .ok_or_else(|| Error::InvalidData("no cookies were provided".to_string()))?
        .to_str()
        .map_err(|e| Error::InvalidData(format!("invalid cookie format: {e}")))?;
```

libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+
+pub async fn auth_from_tonic<T: Serialize>(
+    request: Request<T>,
+    blockscout_host: &str,
+) -> Result<AuthSuccess, Error> {
+    let jwt = extract_jwt(request.metadata())?;
+    let csrf_token = serde_json::to_value(request.into_inner())
+        .map_err(|e| Error::InvalidData(format!("invalid request payload: {e}")))?
+        .get(CSRF_TOKEN_NAME)
+        .and_then(|token| token.as_str().map(|s| s.to_string()));
```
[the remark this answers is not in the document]
sevenzing: no, csrf token located in body

libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+pub struct AuthSuccess {
+    pub user_id: String,
+}
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum Error {
+    #[error("invalid jwt token: {0}")]
+    InvalidJwt(String),
+    #[error("invalid cstf token: {0}")]
+    InvalidCsrf(String),
```
rimrakhimov, line 18: Nit: given that csrf is a type of attack, it's logically quite strange to have `InvalidCsrf`
```suggestion
    InvalidCsrfToken(String),
```

libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[derive(Debug)]
+pub struct AuthSuccess {
+    pub user_id: String,
+}
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum Error {
+    #[error("invalid jwt token: {0}")]
+    InvalidJwt(String),
```
rimrakhimov, line 16: Nit: just for consistency
```suggestion
    InvalidJwtToken(String),
```

libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[error("invalid cstf token: {0}")]
```
rimrakhimov, line 17: typo:
```suggestion
    #[error("invalid csrf token: {0}")]
```

libs/blockscout-auth/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    jwt: &str,
+    csrf_token: Option<&str>,
+    _blockscout_host: &str,
+) -> Result<AuthSuccess, Error> {
+    // TODO: replace with actual blockscout api call
+    let mut jwt = jwt.to_string();
+    if let Some(cstf_token) = csrf_token {
+        jwt.push_str(cstf_token)
+    }
+    Ok(AuthSuccess { user_id: jwt })
```
rimrakhimov, line 44: Nit: just some easier option
```suggestion
    let user_id = format!("{jwt}{}", csrf_token.unwrap_or_default());
    Ok(AuthSuccess { user_id } )
```

rimrakhimov (approved):

PR #397, opened 2023-03-15
Add blockscout-auth main logic

sevenzing:
https://github.com/blockscout/blockscout-admin/issues/48 : Update authentication library to make actual requests to blockscout


libs/blockscout-auth/src/lib.rs
```diff
@@ -1,184 +1,5 @@
...
-            .expect_err("success response from request without csrf_token");
-
-        let request = Request::new(GetBody {});
-        auth_from_metadata(request.metadata(), true, "")
-            .await
-            .expect_err("success response for empty request");
-    }
-}
+pub use auth::{auth_from_metadata, auth_from_tokens, AuthSuccess, Error};
+pub use mock::{init_mocked_blockscout_auth_service, MockUser};
```
rimrakhimov, line 5: What making the public is for?

libs/blockscout-auth/src/auth.rs
```diff
@@ -0,0 +1,347 @@
...
+        .build()
+        .map_err(|e| Error::InternalError(e.to_string()))?;
+    let mut url = blockscout_host
+        .join("/api/account/v1/authenticate")
+        .expect("should be valid url");
+    url.set_query(
+        blockscout_api_key
+            .map(|api_key| format!("{API_KEY_NAME}={api_key}"))
+            .as_deref(),
+    );
```
rimrakhimov, line 100: Maybe send api_key in the header instead?

libs/blockscout-auth/src/auth.rs
```diff
@@ -0,0 +1,347 @@
...
+
+    let response = if csrf_token.is_some() {
+        client.post(url)
+    } else {
+        client.get(url)
+    }
+    .send()
+    .await
+    .map_err(|_| Error::BlockscoutApi("failed to connect".to_string()))?;
```
rimrakhimov, line 109: Nit: may be a little clearer as avoid building a new client from builder which may potentially result in error
```suggestion
    let headers = build_http_headers(jwt, csrf_token)?;
    let client = reqwest::Client::new();
    let response = if csrf_token.is_some() {
        client.post(url)
    } else {
        client.get(url)
    }
    .headers(headers)
    .send()
    .await
    .map_err(|_| Error::BlockscoutApi("failed to connect".to_string()))?;
```

libs/blockscout-auth/src/auth.rs
```diff
@@ -0,0 +1,347 @@
...
+#[serde(untagged)]
+enum BlockscoutAuthResponse {
+    Success(BlockscoutAuthSuccessResponse),
+    Error(BlockscoutAuthErrorResponse),
+}
+
+#[derive(Debug, Clone, Deserialize)]
+struct BlockscoutAuthErrorResponse {
+    message: String,
+}
```
rimrakhimov, line 31: Wouldn't the request result in non 200 status code in that case?

libs/blockscout-auth/src/auth.rs
```diff
@@ -0,0 +1,347 @@
...
+    };
+    auth_from_tokens(&jwt, csrf_token, blockscout_host, blockscout_api_key).await
+}
+
+pub async fn auth_from_tokens(
+    jwt: &str,
+    csrf_token: Option<&str>,
+    blockscout_host: &Url,
+    blockscout_api_key: Option<&str>,
+) -> Result<AuthSuccess, Error> {
```
rimrakhimov, line 87: Why not to return `BlockscoutAuthSuccessResponse` (just renamed to `AuthSuccess`)?

libs/blockscout-auth/src/lib.rs
sevenzing: I will need to run mock server in admin-rs

libs/blockscout-auth/src/auth.rs
```diff
@@ -0,0 +1,347 @@
...
+    .map_err(|_| Error::BlockscoutApi("failed to connect".to_string()))?;
+
+    let status = response.status();
+    let response_raw = response
+        .text()
+        .await
+        .map_err(|e| Error::BlockscoutApi(e.to_string()))?;
+
+    match status {
+        StatusCode::OK | StatusCode::UNAUTHORIZED => {
```
rimrakhimov, line 118: It may be better to process those two cases separately. If I get it right that `BlockscoutAuthResponse::Error` is returned in the case of `StatusCode::UNAUTHORIZED` and `BlockscoutAuthResponse::Success` in case of `StatusCode::OK`

libs/blockscout-auth/src/auth.rs
sevenzing: the problem here is that blockscout returns `id: i64`, but in admin-rs we always need String type, so I decided to introduce new struct AuthSuccess to separate blockscout's response and library's response

libs/blockscout-auth/src/auth.rs
sevenzing: Yeah, ok. I did it for the first time, then I thought it would be better to do this way because of less-copy paste

libs/blockscout-auth/src/auth.rs
sevenzing: If blockscout returned 200 with {"message": ""} then yes, otherwise no

libs/blockscout-auth/src/auth.rs
rimrakhimov: I suggest for the lib to return the same output as blockscout does (including `id` as `i64` if that is the case). I believe it should be the responsibility of the client application to decide what fields are needed and how they should be interpreted. It is better not to make those a library concerns

libs/blockscout-auth/src/auth.rs
sevenzing: yeah, also your way is better because less indent level, will fix it

libs/blockscout-auth/src/auth.rs
sevenzing: ok

libs/blockscout-auth/src/auth.rs
rimrakhimov: Get it; probably we need to discuss that with them

rimrakhimov:
@sevenzing Also, I suggest to create a separate Client which would be initialized with blockscout url and will initialize a reqwest Client in the constructor. That was we will not have to pass the same arguments for every authentication request, and could reuse the same reqwest::Client, which further may be replaced by some more advanced version (e.g. reqwest-retry)

libs/blockscout-auth/src/auth.rs
rimrakhimov: And what is the case when 200 with {"message": ""} is returned?

sevenzing:
@rimrakhimov I would prefer to leave it as it is, since do I not like idea of writing another client for blockscout-auth. So my suggestion is to merge this code and find the correct way to deal with it later

rimrakhimov (approved):

PR #398, opened 2023-03-15
Drop last point config variable

### LINKED ISSUE #374: stats: Exclude last point for each chart
Add a setting that will allow to exclude last point for each chart

leviska:
Closes #374

### DIFF
--- a/stats/stats-server/src/read_service.rs
+++ b/stats/stats-server/src/read_service.rs
@@ -83,8 +83,14 @@ impl StatsService for ReadService {
             })
             .collect();
 
-        // remove last data point, because it can be partially updated
-        data.pop();
+        let settings =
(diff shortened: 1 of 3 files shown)


stats/stats-server/src/read_service.rs
```diff
@@ -83,8 +83,14 @@ impl StatsService for ReadService {
             })
             .collect();
 
-        // remove last data point, because it can be partially updated
-        data.pop();
+        let settings =
+            self.charts.settings.get(&request.name).ok_or_else(|| {
+                tonic::Status::not_found(format!("chart {} not found", request.name))
+            })?;
```
sevenzing, line 89: maybe move this code to line#66?  because it looks like double check the same thing

sevenzing (approved):

PR #407, opened 2023-03-27
Initial blockscout client

sevenzing:


libs/blockscout-client/src/lib.rs
```diff
@@ -0,0 +1,577 @@
+
+#[macro_use]
+extern crate serde;
+
+pub mod address {
+    include!("./address.rs");
```
leviska, line 6: why include and not just `mod address`?

leviska: ah, it's autogenerated?

Issue #416, opened 2023-04-09
Weaknesses in the `Metadata` Identification Method


Hellobloc:
### Background
As mentioned in the documentation with blockscout, your approach to distinguish `Main Part` and `Metadata Part` is to validate the metadata part by compiling the `contract` and `compiler_modified` separately to look for different content. 

This is clever because it allows contracts containing multiple metadata to be validated, and it could in theory even support validation of the following contracts.
```solidity 
pragma solidity ^0.8.0;
contract Deploy2 {
    address public immutable a;

    bytes32 public ptr;
    constructor(address d) public {
        a = address(d);//
    }
}
contract Deploy3 {

    bytes32 public ptr;//

}
contract A {
    bytes public constant code = type(Deploy2).creationCode;
    bytes public constant code2 = type(Deploy3).creationCode;
    bytes1 public immutable flag;
    bytes1 public immutable flag2;
    constructor() public {
        flag = code[code.length-20]&0xaa;
        flag2 = code2[code2.length-20]&0xaa;
    }
}
```

But only if `compiler_modified` does not trigger changes to other non-metadata parts.
Your design for `compiler_modified` is as follows, i.e., it causes changes to the metadata by adding a `Library` that does not seem to be used.
https://github.com/blockscout/blockscout-rs/blob/2cdb3c2c8e532c3d8ac3718cf01a3462d36636c1/smart-contract-verifier/smart-contract-verifier/src/verifier/contract_verifier.rs#L100-L121
### Risk Description
However, this seems to cause the `Library` address to be marked as `metadata` in some extreme cases, which may pose some risks, such as the `library` address and other adjacent bytecodes provided in the source code not corresponding to the `real bytecode`.

## Exploit
The user can make the `library` address also at risk of being marked as `metadata` by adding a contract with `SOME_TEXT_USED_AS_FILE_NAME` as the file name in the contract source code and defining a `library` named `L_` in that file.
The details are as follows：
```javascript
var solc = require('solc');

var input = {
          language: 'Solidity',
          settings: {
            libraries: {
              'SOME_TEXT_USED_AS_FILE_NAME': {
                L_: '0x4200000000000000000000000000000000000001'
              }
            },
            outputSelection: {
              '*': {
                '*': ['evm.bytecode']
              }
            }
          },
          sources: {
            'SOME_TEXT_USED_AS_FILE_NAME': {
              content: 'library L_ { function f() public returns (uint) { return 7; } }'
            },
            'a.sol': {
[...]

PR #420, opened 2023-04-12
Fix metadata extraction

rimrakhimov:
Fixes the issue with adding a library to generate compiler outputs with updated metadata well explained in https://github.com/blockscout/blockscout-rs/issues/416. 

With the previous approach we've created a non-existent library to be passed in the compiler input to generate an alternative metadata hash. The problem was that to get the non-existent library name we've used only the first letters of all existing libraries. That allowed some specifically crafted input to trick the verifier to consider Main bytecode part as Metadata.

This PR fixes that by adding a space to the end of all source files instead of modifying libraries input field.

We are really greateful for @Hellobloc for pointing this problem out!


Hellobloc:
Nice job!

PR #434, opened 2023-04-17
Fix drop last point

sevenzing:


stats/stats/src/charts/insert.rs
```diff
@@ -97,6 +97,11 @@ impl DateValue {
             self
         }
     }
+
+    pub fn can_be_partial(&self) -> bool {
```
rimrakhimov, line 101: Nit: imo, `can_be_partial` does not represent what function does. My suggestion is to name it `is_partial`

rimrakhimov (approved):

PR #480, opened 2023-05-30
Add async update

sevenzing:


stats/stats-server/src/settings.rs
```diff
@@ -32,7 +32,7 @@ pub struct Settings {
     pub blockscout_db_url: String,
     #[serde_as(as = "DisplayFromStr")]
     pub default_schedule: Schedule,
-    pub force_update_on_start: Option<bool>, // None = no update
```
rimrakhimov, line 35: That is a breaking change to remove the configuration option and include it into other struct. So, I insist on leaving it as it is for now, and include it into the group when we will be working on v2.0.0

rimrakhimov (approved):

PR #499, opened 2023-06-07
verifier: make use of custom Compiler Input

rimrakhimov:
Define a custom CompilerInput for Vyper verification that includes interfaces, and valid optimization settings. Update multi-part verifiaction to use that compiler input


smart-contract-verifier/smart-contract-verifier/src/vyper/artifacts.rs
```diff
@@ -0,0 +1,295 @@
...
+pub enum Interface {
+    Vyper(ethers_solc::artifacts::Source),
+    Abi(interfaces::AbiSource),
+    ContractTypes(interfaces::ContractTypesSource),
+}
+
+impl Interface {
+    pub fn try_new(path: &Path, content: String) -> Result<Self, anyhow::Error> {
+        // Adapted from: https://github.com/vyperlang/vyper/blob/v0.3.9/vyper/cli/vyper_compile.py#L221
+        if path.extension() == Some(std::ffi::OsStr::new("json")) {
```
sevenzing, line 100: hm, just note that if file has name `.json` then path.extensions() will be an empty string, should we care about it?

rimrakhimov: I believe, no. Because in that case that file could not be imported properly anyway

smart-contract-verifier/smart-contract-verifier/src/vyper/artifacts.rs
```diff
@@ -0,0 +1,295 @@
...
+            let content: serde_json::Value = serde_json::from_str(&content)
+                .map_err(|err| anyhow::anyhow!("couldn't parse content as a json value: {err}"))?;
```
sevenzing, line 102: `couldn't parse content of interface as a json value: {err}`

smart-contract-verifier/smart-contract-verifier/src/vyper/artifacts.rs
```diff
@@ -0,0 +1,295 @@
...
+#[serde(untagged)]
+pub enum Interface {
+    Vyper(ethers_solc::artifacts::Source),
+    Abi(interfaces::AbiSource),
+    ContractTypes(interfaces::ContractTypesSource),
+}
+
+impl Interface {
+    pub fn try_new(path: &Path, content: String) -> Result<Self, anyhow::Error> {
+        // Adapted from: https://github.com/vyperlang/vyper/blob/v0.3.9/vyper/cli/vyper_compile.py#L221
```
sevenzing, line 99: Adapted from: https://github.com/vyperlang/vyper/blob/v0.3.9/vyper/cli/vyper_compile.py#L217-L246

sevenzing (approved):

PR #533, opened 2023-06-17
Fix typo

nedondev:
I change from a `usefule` to a `useful` in README.md. I don't know why the last line was difference in git diff.


rimrakhimov:
@nedondev Thank you for your PR. Merged it into the main

Issue #546, opened 2023-06-23
Add convenient way to configure charts


sevenzing:
For now charts configuration is not convenient using .toml file. 

Also disabling charts is very inconvenient, add way to disable charts using ENV

PR: https://github.com/blockscout/blockscout-rs/pull/547

PR #547, opened 2023-06-23
Add json+env chart configuration

sevenzing:
solves #546 

Added JSON option for chart configuration

Also we added ENV support for configuring charts, for example, disabling `totalBlocks` and `activeAccounts` can be done with
```bash
STATS_CHARTS__LINES__ACCOUNTS__CHARTS__ACTIVE_ACCOUNTS__ENABLED=false
#                    ^section name     ^line name
STATS_CHARTS__COUNTERS__TOTAL_BLOCKS__ENABLED=false
#                       ^counter name
```

### DIFF
--- a/stats/stats-server/src/config/json_config.rs
+++ b/stats/stats-server/src/config/json_config.rs
@@ -1,0 +1,87 @@
+use super::{
+    chart_info::{CounterInfo, LineChartInfo},
+    toml_config,
+};
+use convert_case::{Case, Casing};
+use serde::Deserialize;
+use std::collections::BTreeMap;
+
+#[derive(Default, Debug, Clone, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct LineChartSection {
+    pub title: String,
+    pub charts: BTreeMap<String, LineChartInfo>,
+}
+
+#[derive(Default, Debug, Clone, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct LineCharts {
+    pub sections: BTreeMap<String, LineChartSection>,
+}
+
+#[derive(Default, Debug, Clone, Deserialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct Config {
+    pub counters: BTreeMap<String, CounterInfo>,
+    pub lines: LineCharts,
+}
+
+impl From<Config> for toml_config::Config {
+    fn from(value: Config) -> Self {
+        Self {
+            counters: value
+                .counters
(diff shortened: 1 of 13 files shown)


stats/stats-server/src/config/read.rs
```diff
@@ -0,0 +1,26 @@
+use super::{json_config, toml_config};
+use std::path::PathBuf;
+
+pub fn read_charts_config(path: PathBuf) -> Result<toml_config::Config, anyhow::Error> {
+    let extension = path.extension();
+    if extension == Some(std::ffi::OsStr::new("json")) {
+        let json_config: json_config::Config = config::Config::builder()
+            .add_source(config::File::from(path))
+            .add_source(
+                config::Environment::with_prefix("STATS_CFG")
```
rimrakhimov, line 10:
```suggestion
                config::Environment::with_prefix("STATS_CHARTS_CONFIG")
```

Or just `STATS_CHARTS`.

`STATS_CFG` does not seem to indicate what that prefix does

stats/stats-server/src/config/read.rs
```diff
@@ -0,0 +1,26 @@
+use super::{json_config, toml_config};
+use std::path::PathBuf;
+
+pub fn read_charts_config(path: PathBuf) -> Result<toml_config::Config, anyhow::Error> {
```
rimrakhimov, line 4: Nit:
```suggestion
pub fn read_charts_config(path: &Path) -> Result<toml_config::Config, anyhow::Error> {
```

stats/stats-server/src/config/json_config.rs
```diff
@@ -0,0 +1,80 @@
...
+                .map(toml_config::LineChartInfo::from)
+                .collect(),
+        }
+    }
+}
+
+impl From<(String, LineChartInfo)> for toml_config::LineChartInfo {
+    fn from((id, info): (String, LineChartInfo)) -> Self {
+        Self {
+            id: id.to_case(Case::Camel),
```
rimrakhimov, line 64: I suggest to add a from statement as well, just to make sure the library recognizes the initial cases correctly and do not recognize any other pattern

```suggestion
            id: id.from_case(Case::Snake).to_case(Case::Camel),
```

rimrakhimov (approved):

stats/stats-server/src/config/read.rs
sevenzing: I decided to choose the shortest prefix. `STATS_CHARTS` is fine

Issue #564, opened 2023-07-17
stats bug: verified contracts growth


sevenzing:
https://eth-goerli.blockscout.com/stats
![Image](https://github.com/blockscout/blockscout-rs/assets/41516657/d7a195e5-36f7-4c80-88b0-b36ac2923f10)

sevenzing:
Released in staging:

https://blockscout-main.k8s-dev.blockscout.com/stats

PR #567, opened 2023-07-18
Fix fill missing point bug

sevenzing:
solves https://github.com/blockscout/blockscout-rs/issues/564

There is a problem with filling zeros after data is retrieved: if `from` is less than min date of extracted data, then "growth" chart is filled with zeros from the left (see screenshots in issue)

Solution is 
+ move "fill missing point" logic to core crate
+ select data with extra points -> fill data correctly -> remove unrelated points


rimrakhimov (commented):
Lttm

PR #575, opened 2023-08-08
sourcify lib: verify from etherscan

rimrakhimov:
Add a new endpoint: `verify_from_sourcify` to sourcify client. Updates the error handling mechanisms


libs/sourcify/src/client.rs
```diff
@@ -94,12 +92,50 @@ impl Client {
...
+        #[serde(rename_all = "camelCase")]
+        struct Request<'a> {
+            chain_id: &'a str,
+            address: String,
+        }
+
+        let request = Request {
+            chain_id,
+            address: contract_address.to_string(),
+        };
```
sevenzing, line 114: I think this way is okay, but I just wonder why not simply use serde_json::json
```
let request = serde_json::json!({
    "chainId": chain_id,
    "address": contract_address.
});
```

sevenzing (approved):

libs/sourcify/src/client.rs
rimrakhimov: I believe it may done both ways, the separate structure was just the first which came up in my head

sevenzing:
maybe rename this crate to `sourcify-client` ?

rimrakhimov:
> maybe rename this crate to sourcify-client ?

We may discuss that in a separate thread I believe

PR #587, opened 2023-08-14
Add filter for zero timestamp for all charts

sevenzing:
https://github.com/blockscout/blockscout-rs/issues/588


rimrakhimov (commented):
Shouldn't we also add that condition to the `native_coin_holders_growth`, `native_coin_supply`, and `new_contracts` charts?

rimrakhimov (approved):

Issue #578, opened 2023-08-10
Extend compiler output returned from verifier service


rimrakhimov:

rimrakhimov:
https://github.com/blockscout/blockscout-rs/pull/601

rimrakhimov:
https://github.com/blockscout/blockscout-rs/pull/605

PR #601, opened 2023-08-31
verifier: extended compiler output

rimrakhimov:
#578


smart-contract-verifier/smart-contract-verifier/src/verifier/all_metadata_extracting_verifier.rs
```diff
@@ -105,6 +107,212 @@ impl<T: Source> Verifier<T> {
...
+                        mod raw_types {
+                            use serde::Deserialize;
+                            use std::collections::BTreeMap;
+
+                            #[derive(Clone, Debug, Deserialize, PartialEq)]
+                            pub struct CompilerOutput {
+                                #[serde(default)]
+                                pub contracts: BTreeMap<String, BTreeMap<String, Contract>>,
+                            }
+
```
sevenzing, line 119: I suggest to move logic of extraction artifacts to separate file, or to the bottom of this file.
in my opinion, `verify` function should not be that big and it should show more business logic and less implementation of parsing

sevenzing (approved):

PR #614, opened 2023-09-19
Verifier alliance database implementation

rimrakhimov:


eth-bytecode-db/eth-bytecode-db/src/verification/handlers/mod.rs
```diff
@@ -132,11 +177,49 @@ async fn process_verify_response(
...
+        Ok(())
+    };
+
     let _ = process_database_insertion()
         .await
         .map_err(|err: anyhow::Error| {
             tracing::error!("Error while inserting contract data into database: {err:#}")
         });
 
+    let _ = process_alliance_database_insertion()
```
sevenzing, line 216: do we want to make `process_database` and `process_alliance_database` concurrenly?

however I see that this function`process_verify_response` is veryyy long and complex function now, so maybe it can be left without complications :)

rimrakhimov: Updated :+1:

sevenzing (approved):

PR #616, opened 2023-09-26
Fix native coin symbol for some charts in config

sevenzing:


rimrakhimov:
Do I get it right, that those descriptions have not contained the native coin symbol before? Why do you want to change them? And why haven't you changed them for `new_native_coin_holders` and `native_coin_supply` in that case?

rimrakhimov (approved):

PR #620, opened 2023-10-03
eth-bytecode-db: update server to support verifier alliance database

rimrakhimov:


eth-bytecode-db/eth-bytecode-db-server/src/settings.rs
```diff
@@ -79,6 +85,19 @@ impl Default for SourcifySettings {
     }
 }
 
+#[derive(Debug, Default, Clone, Deserialize, PartialEq, Eq)]
+#[serde(default, deny_unknown_fields)]
+pub struct VerifierAllianceDatabaseSettings {
+    pub enabled: bool,
+    pub url: String,
```
sevenzing, line 92: why not `url::Url` to add config validation ?)

sevenzing (approved):

eth-bytecode-db/eth-bytecode-db-server/src/settings.rs
rimrakhimov: Actually, because that is how `sea-orm` handles connection url

PR #626, opened 2023-10-09
Refactor visualizer service

### LINKED ISSUE #623: Update visualizer service structure and initialization to standard form
The standard development approach to blockscout microservices is to define a proto definition and http transcoding in a separate `service_name-proto` crate, have the business logic in a separate `service_name` crate, and connect business logic with proto definitions through `service_name-server` crate which have the settings and servers startup logic inside. The servers startup logic has been mostly extracted into [`blockscout-service-launcher`](https://github.com/blockscout/blockscout-rs/tree/main/libs/blockscout-service-launcher/) library. 

Visualizer service was developed before those practices had been accepted. Here we want to refactor the service to have the same structure as others.

The issue has two parts:
1. Extract proto definition and building of tonic and actix services into a separate crate (`visualizer-proto`)
2. Update the server initialization to use the `blockscout-service-launcher` library, as now this logic is hardcoded inside the server crate  itself

lok52:
Closes #623 

Updates:
1. Extract proto definitions in a separate crate
2. Use `blockscout-service-launcher` for server initialisation


visualizer/visualizer-server/src/server.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    let visualizer = Arc::new(SolidityVisualizerService::default());
+    let health = Arc::new(HealthService::default());
+
+    let router = Router { visualizer, health };
+    let grpc_router = router.grpc_router();
+    let http_router = router;
+
+    let launch_settings = LaunchSettings {
+        service_name: "visualizer".to_owned(),
```
rimrakhimov, line 47: Nit: we probably can reuse the `SERVICE_NAME` const here

rimrakhimov (approved):

visualizer/visualizer-server/src/server.rs
lok52: Fixed

PR #627, opened 2023-10-10
Add launcher test utils

lok52:
Add test utilities for launcher service:
1. Generic `TestDbGuard`
2. Server start-up utilities


libs/blockscout-service-launcher/src/test/server.rs
```diff
@@ -0,0 +1,76 @@
...
+        if let Ok(_response) = client
+            .get(health_endpoint.clone())
+            .query(&[("service", health_check_service)])
+            .send()
+            .await
+        {
+            break;
+        }
+    }
+}
```
sevenzing, line 37: what if `run().await` failed? this loop wont stop actually, right?
I suggest to use https://docs.rs/reqwest-retry to make requests with total max duration of 10 seconds

sevenzing (approved):

rimrakhimov:
I suggest to make it two separate modules available under different set of features (`test-database` and `test-server`). For some cases (see `eth-bytecode-db//tests`) we may need to initialize only database without the server, and in that case it may make sense to use only one of the modules.

Also, maybe it would be possible to drop the database when the test fails? In my experience, it may be become a mess sometimes when you have lots of test cases persisting in the database from all kind of tests. I believe it may still be useful in some cases (but is not in most), so if we could set it explicitly on initialization if the database should be dropped (but drop it by default) would be really great. I'm not sure if that is possible, though

libs/blockscout-service-launcher/Cargo.toml
```diff
@@ -83,3 +85,12 @@ database-0_10 = [
     "dep:sea-orm-0_10",
     "dep:sea-orm-migration-0_10",
 ]
+
+test-server = [
+    "dep:reqwest",
```
rimrakhimov, line 90: Probably missed "launcher" dependency?

libs/blockscout-service-launcher/src/test_server/server.rs
```diff
@@ -0,0 +1,90 @@
...
+    tokio::spawn(async move { run().await });
+
+    let client = reqwest::Client::new();
+    let health_endpoint = base.join("health").unwrap();
+
+    let wait_health_check = async {
+        loop {
+            if let Ok(_response) = client
+                .get(health_endpoint.clone())
+                .query(&[("service", health_check_service)])
```
rimrakhimov, line 39: I'm not sure, but the actual service name probably may be omitted. If I'm not mistaken, the most important is that the request should contain `service=` query inside. But it may be left as it is without any concrete service names. We would ignore the name anyway

libs/blockscout-service-launcher/src/test_database/mod.rs
```diff
@@ -0,0 +1,3 @@
+mod database;
+
+pub use database::TestDbGuard;
```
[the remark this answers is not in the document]
lok52: I guess we can move the modules into a separate files like `test_database.rs` and export them in `lib.rs`

libs/blockscout-service-launcher/src/test/database.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::sync::Arc;
+
+#[derive(Clone, Debug)]
+pub struct TestDbGuard {
+    conn_with_db: Arc<DatabaseConnection>,
+    db_url: String,
+}
+
+impl TestDbGuard {
+    pub async fn new<Migrator: MigratorTrait>(db_name: &str, db_url: String) -> Self {
```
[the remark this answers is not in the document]
lok52: Fixed

libs/blockscout-service-launcher/src/test_server/server.rs
lok52: Fixed

libs/blockscout-service-launcher/src/test_server.rs
```diff
@@ -0,0 +1,96 @@
...
+        let message = response.text().await.expect("Read body as text");
+        panic!("({annotation})Invalid status code (success expected). Status: {status}. Message: {message}")
+    }
+
+    response
+        .json()
+        .await
+        .unwrap_or_else(|_| panic!("({annotation})Response deserialization failed"))
+}
+pub async fn send_post_request<Response: for<'a> serde::Deserialize<'a>>(
```
rimrakhimov, line 83: I believe it may be useful to have annotated versions of `send_post_request` and `send_get_request` methods. We make use of annotated post request in `eth-bytecode-db`

rimrakhimov (approved):

sevenzing (approved):

libs/blockscout-service-launcher/src/test_server.rs
lok52: Updated

PR #628, opened 2023-10-11
Bens: initial proto def

sevenzing:
Initial pr of BENS (blockscout ens)


blockscout-ens/bens-proto/proto/bens.proto
```diff
@@ -0,0 +1,85 @@
...
+enum DomainSource {
+  /// https://docs.ens.domains/
+  ENS = 0;
+  // /// https://docs.unstoppabledomains.com/
+  //UNSTOPPABLE_DOMAINS = 1;
+}
+
+message Domain {
+  /// Source of domain name. Now could be only ENS, in future we could add UNSTOPPABLE_DOMAINS
+  DomainSource domain_source = 1;
```
rimrakhimov, line 21: Let maybe add this field when another domain will be added? It seems to be non-breaking change after all

blockscout-ens/bens-proto/proto/bens.proto
```diff
@@ -0,0 +1,85 @@
...
+  /// Unique id for the domain, also known as nodehash
+  string id = 2;
+  /// The human readable name, if known. Unknown portions replaced with hash in square brackets (eg, foo.[1234].eth)
+  string name = 3;
+  /// Integer representation of labelhash
+  uint64 token_id = 4;
```
rimrakhimov, line 27: Could you please add an example of what is that

blockscout-ens/bens-proto/proto/bens.proto
```diff
@@ -0,0 +1,85 @@
...
+  /// Optinal. Resolved address of this domain
+  optional string resolved_address = 6;
+  /// Optinal. The account that owns the ERC721 NFT for the domain
+  optional string registrant = 7;
+  /// Optinal. Unix timestamp of expiry date. None means never expires
+  optional uint64 expiry_date = 8;
+  /// Unix timestamp of regisration date
+  uint64 registration_date = 9;
+  /// Map chain -> resolved_address. This map will contain `current_chain_id` -> `resovled_address` if `resovled_address` is not None
+  map<string, string> other_address = 10;
```
rimrakhimov, line 39: What are the use cases we need this map for?

rimrakhimov, line 39: What are the use cases we need this map for?

blockscout-ens/bens-proto/proto/api_config_http.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+http:
+  rules:
+    #################### DomainsExtractor ####################
+    - selector: blockscout.bens.v1.DomainsExtractor.SearchAddress
+      post: /api/v1/domains/search-address
+      body: "*"
+
+    - selector: blockscout.bens.v1.DomainsExtractor.GetDetailedDomain
+      post: /api/v1/domains/get-detailed
```
rimrakhimov, line 12: I believe it would be better to make it as get request with more resty paths. E.g:

```suggestion
      get: /api/v1/{chain_id}/detailed-domains/{name}
```

And also name the response structure just as `DetailedDomain` instead of `GetDetailedDomainResponse`

blockscout-ens/bens-proto/proto/api_config_http.yaml
```diff
@@ -0,0 +1,18 @@
+type: google.api.Service
+config_version: 3
+
+http:
+  rules:
+    #################### DomainsExtractor ####################
+    - selector: blockscout.bens.v1.DomainsExtractor.SearchAddress
+      post: /api/v1/domains/search-address
```
rimrakhimov, line 8: Lets also make it more in line with our other service apis: 
```suggestion
      post: /api/v1/domains/address:search
```

blockscout-ens/bens-proto/proto/bens.proto
sevenzing: https://etherscan.io/name-lookup-search?id=levvv.eth

blockscout-ens/bens-proto/proto/bens.proto
sevenzing: you can bind any address of any chain to ENS domain:
here is example of only ethereum address
image

blockscout-ens/bens-proto/proto/api_config_http.yaml
sevenzing: domain name can contains emoji and non ascii symbols, so I thought that maybe it is better to put it into the body. surely you can encode emoji into url, we can do it this way if you want

blockscout-ens/bens-proto/proto/api_config_http.yaml
sevenzing: doesn't it sound like you searching for address, not searching domains using address?

blockscout-ens/bens-proto/proto/bens.proto
sevenzing: duplicate probably

sevenzing: image

blockscout-ens/bens-proto/proto/api_config_http.yaml
sevenzing: maybe `/api/v1/{chain_id}/domains:reverse-search`

rimrakhimov (approved):

rimrakhimov (approved):

PR #638, opened 2023-10-19
Bens-logic: initial

sevenzing:
this pr adds basic logic of bens service:
+ getting domain by name
+ searching domain by resolved or owned address

I decided to use `sqlx` engine for database, because sea-orm doesnt support `INT4RANGE` postgres types and doesnt have option to change schema name of table in runtime

also I added several tests using sqlx::test (very convenient!)


blockscout-ens/bens-logic/src/subgraphs_reader/reader.rs
```diff
@@ -0,0 +1,237 @@
...
+
+async fn find_domain(
+    pool: &PgPool,
+    schema: &str,
+    id: &str,
+) -> Result<Option<Domain>, SubgraphReadError> {
+    let maybe_domain = sqlx::query_as(&format!(
+        r#"
+        SELECT
+        DISTINCT ON (id) *
```
rimrakhimov, line 88: Why do we need DISTINCT ON here, if id is unique anyway (we provide a concrete id in WHERE clause)?

blockscout-ens/bens-logic/src/subgraphs_reader/reader.rs
```diff
@@ -0,0 +1,237 @@
...
+            WHERE 
+                (
+                    owner = $1
+                    OR wrapped_owner = $1
+                )
+                AND name IS NOT NULL
+                AND to_timestamp(expiry_date) > now()
+            ORDER BY
+                id,
+                block_range DESC
```
rimrakhimov, line 154: Q: Just interested, what `block_range` is used for in "domain" table?

rimrakhimov (approved):

blockscout-ens/bens-logic/src/subgraphs_reader/reader.rs
sevenzing: ID is not unique field in this db, however I changed distinct on (id) + order by id to graph-node way:
https://github.com/graphprotocol/graph-node/blob/19fd41bb48511f889dc94f5d82e16cd492f29da1/store/postgres/src/block_range.rs#L26

blockscout-ens/bens-logic/src/subgraphs_reader/reader.rs
sevenzing: graph-node stores history of every entity using block_range. range `[1234,)` (right bound is null) means that this is current state of enitity

blockscout-ens/Cargo.toml
```diff
@@ -1,5 +1,6 @@
 [workspace]
 
 members = [
+    "bens-logic",
```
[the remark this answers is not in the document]
sevenzing: I would like to leave `logic` in crate name, I think it provides more info on what it is

blockscout-ens/bens-logic/src/subgraphs_reader/reader.rs
```diff
@@ -0,0 +1,233 @@
...
+        SELECT *
+        FROM {schema}.domain
+        WHERE
+            id = $1 
+            AND name IS NOT NULL
+            AND (
+                expiry_date is null
+                OR to_timestamp(expiry_date) > now()
+            )
+            AND block_range @> 2147483647
```
rimrakhimov, line 96: Nit: it make sense to add a comment here

rimrakhimov (approved):

blockscout-ens/bens-logic/src/subgraphs_reader/reader.rs
sevenzing: Moved common part to single variable and added comment for that var

PR #642, opened 2023-10-23
Fix for get-domain

sevenzing:
We have to return expired domains only in get-domain handle, but not during searching for resolved addresses


rimrakhimov (approved):

blockscout-ens/bens-logic/src/entity/subgraph/domain.rs
```diff
@@ -10,11 +12,12 @@ pub struct Domain {
...
+    pub ttl: Option<chrono::DateTime<Utc>>,
     pub is_migrated: bool,
-    pub created_at: sqlx::types::BigDecimal,
+    pub created_at: chrono::DateTime<Utc>,
     pub owner: String,
     pub registrant: Option<String>,
     pub wrapped_owner: Option<String>,
-    pub expiry_date: Option<sqlx::types::BigDecimal>,
+    pub expiry_date: Option<chrono::DateTime<Utc>>,
+    pub is_expired: bool,
```
rimrakhimov, line 22: Where that field is used?

sevenzing: I added this field for frontend, just in case

PR #646, opened 2023-10-24
Add cors settings

sevenzing:
Added actix-cors crate to handle cors headers in requests. Disabling cors middleware by default

Note that wildcarding in allowed_origin is not supported, so `http://*.blockscout.com` is not a valid origin

### DIFF
--- a/libs/blockscout-service-launcher/Cargo.toml
+++ b/libs/blockscout-service-launcher/Cargo.toml
@@ -12,7 +12,7 @@ edition = "2021"
 
 [dependencies]
 actix-web = { version = "4", optional = true }
-actix-cors = "0.6.4"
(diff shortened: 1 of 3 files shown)


libs/blockscout-service-launcher/Cargo.toml
```diff
@@ -12,6 +12,7 @@ edition = "2021"
 
 [dependencies]
 actix-web = { version = "4", optional = true }
+actix-cors = "0.6.4"
```
rimrakhimov, line 15:
```suggestion
actix-cors = { version = "0.6.4", optional = true }
```

rimrakhimov (commented):
Could you please check on some service with several concurrent integration tests that setting up the Metrics middleware won't break anything in running tests

PR #648, opened 2023-10-24
Stats: bump blockscout-service-launcher

sevenzing:
Updates lib version in stats


stats/stats-server/src/settings.rs
```diff
@@ -37,6 +37,7 @@ impl Default for Settings {
                     enabled: true,
                     addr: SocketAddr::from_str("0.0.0.0:8050").unwrap(),
                     max_body_size: 2 * 1024 * 1024, // 2 Mb - default Actix value
+                    ..Default::default()
```
rimrakhimov, line 40: Nit: I believe it would be better to state the cors config explicitly (i.e. `cors: Default::default`)

rimrakhimov (approved):

PR #649, opened 2023-10-25
Bens: history of domain

sevenzing:
Add history of domain handling. 

+ frontend needs timestamp and sender of transaction, however subgraph database have only transaction hash. for now I decided to call blockscout API for additional transaction data, in future we will cache/index this info somehow
+ SQL for extracting domain events is complicated, and it requires templating in any way (schema name) so I decided to codegen using tera (powerful template engine) -- see `history.sql`


blockscout-ens/bens-logic/src/subgraphs_reader/sql/transaction_history.rs
```diff
@@ -0,0 +1,82 @@
...
+            ],
+            "registration_event_tables": [
+                "name_registered",
+                "name_renewed",
+                "name_transferred"
+            ]
+        }))
+        .expect("failed to load history context")
+    };
+}
```
rimrakhimov, line 63: Wouldn't those values be different depending on the chain we are processing?

sevenzing: For now, we have only 2 subgraphs, and their schema is identical, so I decided to use common values for them

sevenzing: If we will face problem of different schemas for events, we will change this code, it won't be hard

PR #652, opened 2023-10-30
eth-bytecode-db-extractors - extract job queue into a separate crate

rimrakhimov:


eth-bytecode-db-extractors/blockscout-entity/src/seaql_migrations.rs
```diff
@@ -0,0 +1,16 @@
...
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub version: String,
+    pub applied_at: i64,
+}
+
+#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
+pub enum Relation {}
+
+impl ActiveModelBehavior for ActiveModel {}
```
sevenzing, line 16: Maybe remove useless codegen?

eth-bytecode-db-extractors/job-queue/src/migration.rs
```diff
@@ -0,0 +1,133 @@
...
+        from_statements(
+            manager,
+            &[
+                create_extension_pgcrypto,
+                create_function_job_queue_set_modified_at,
+                create_enum_job_status,
+                create_table_job_queue,
+                create_index_job_queue_status,
+                create_trigger_job_queue_set_modified_at,
+                create_function_insert_job,
```
sevenzing, line 76: tbh, I don't like the idea that you declare 7 variables and then copypaste them here. if you miss some variable or write it twice, compiler wont tell you and it will be hard to spot the mistake. 

Maybe write them inside `&[]` and add `-- sql comment` if you need some kind of names for your sql statements

sevenzing: same problem with `fn down()`

sevenzing: oh, I see that you use the same approach in `blockscout-migration`... okay, then we can leave it as it is for consistency :)

rimrakhimov: The problem here is that you cannot put a single string with several sql statements into `Statement::from_string`. Usually we are using `from_sql` method for that cases that internally split all statements based on ';'. But in functions we use ";" internally and thus, the approach with splitting cannot be applied here. That is why we need a separate statements

PR #656, opened 2023-11-01
Change bens proto

sevenzing:
I changed API:
+ GetDomain and ListDomainEvents now accept `name` instead of `id`
+ Removed DomainView enum
+ Splitted Domain into Domain and DetailedDomain, such that user will not decide between Full and Basic info
+ Change output of ResolveAddress and ListDomain -- single field called `items` to support future pagination
+ Added pagination field to include total_records field
+ Added sorting fields
+ Added comments for every RPC function to show where frontend should use it
+ Change type of `address` fields to Address with single value `hash`. in future we may add some additional information about address, like `is_contract` etc

### DIFF
--- a/blockscout-ens/bens-proto/proto/bens.proto
+++ b/blockscout-ens/bens-proto/proto/bens.proto
@@ -17,10 +17,6 @@ service DomainsExtractor {
   rpc BatchResolveAddressNames(BatchResolveAddressNamesRequest) returns (BatchResolveAddressNamesResponse) {}
 }
 
-message Address {
-  string hash = 1;
-}
-
 message Domain {
   // Unique id for the domain, also known as nodehash
   string id = 1;
@@ -69,10 +65,13 @@ message DomainEvent {
   optional string action = 4;
 }
 
-enum OrderDirection {
-    NOT_SET = 0;
-    ASC = 1;
-    DESC = 2;
+message Address {
+  string hash = 1;
+}
+
+enum Order {
+    ASC = 0;
+    DESC = 1;
 }
 
 message Pagination {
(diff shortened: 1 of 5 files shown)


blockscout-ens/bens-proto/proto/bens.proto
```diff
@@ -40,101 +62,99 @@ message DomainEvent {
...
 }
 
-message ResolveAddressResponse {
-  // List of domains that resolved to requested address
+message LookupAddressResponse {
+  // List of domains that resolved to or owned by requested address
   // Sorted by relevance, so first address could be displayed as main resolved address
-  repeated Domain resolved = 1;
-  // List of domains owned by requested address
-  repeated Domain owned = 2;
```
rimrakhimov, line 128: We we decided to combine it into one field?

rimrakhimov (commented):
Could you please move `Address`, `OrderDirection`, and `Pagination` in the following order `Address`, `Pagination`, `OrderDirection` (see https://google.aip.dev/191#file-layout)

blockscout-ens/bens-proto/proto/bens.proto
sevenzing: yes, single `items` field is more appropriate here

sevenzing: for instance owned can be just copy of resolved (If I own domain, I can easily set resolved_address to my addr). therefore we need to remove duplicates. frontend can distinguish between resolved and owned by simply filtering `const resolved = items.filter((item) => item.resolved_address == address)`. also single `items` field is appropriate way to do pagination

blockscout-ens/bens-proto/proto/bens.proto
```diff
@@ -40,101 +62,99 @@ message DomainEvent {
...
+  Address from_address = 3;
   // Optinal. Action name
   optional string action = 4;
 }
 
+enum OrderDirection {
+    NOT_SET = 0;
+    ASC = 1;
+    DESC = 2;
+}
```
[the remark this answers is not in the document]
sevenzing: renamed to Order and removed NOT_SET

rimrakhimov (approved):

PR #662, opened 2023-11-02
eth-bytecode-db: update values in the database if they have already existed

rimrakhimov:
The smart-contract-verifier may contains some bugs inside, which may cause it return some invalid data (e.g., #661). But already verified contracts will not change if we try to re-verify them after bugs are fixed, as for now the service have just ignored sources which are already in the database.

This PR makes the verification data store to be updated if the data for the existing unique constraints is inserted. That would allow to update the contracts inside the database by simple re-verification of them.

### DIFF
--- a/eth-bytecode-db/eth-bytecode-db/src/verification/handlers/mod.rs
+++ b/eth-bytecode-db/eth-bytecode-db/src/verification/handlers/mod.rs
@@ -178,7 +178,6 @@ async fn process_verify_response(
     let (process_eth_bytecode_db_result, process_alliance_db_result) =
         futures::future::join(process_eth_bytecode_db_future, process_alliance_db_future).await;
     let _ = process_eth_bytecode_db_result.map_err(|err: anyhow::Error| {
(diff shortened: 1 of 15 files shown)


eth-bytecode-db/eth-bytecode-db/src/verification/handlers/mod.rs
```diff
@@ -178,13 +178,15 @@ async fn process_verify_response(
...
     let _ = process_eth_bytecode_db_result.map_err(|err: anyhow::Error| {
+        println!("Error while inserting contract data into database: {err:#}");
         tracing::error!(
             ?eth_bytecode_db_action_contract_address,
             ?eth_bytecode_db_action_chain_id,
             "Error while inserting contract data into database: {err:#}"
         )
     });
     let _ = process_alliance_db_result.map_err(|err: anyhow::Error| {
+        println!("Error while inserting contract data into verifier alliance database: {err:#}");
```
sevenzing, line 189: forget to remove prints?

sevenzing (approved):

PR #664, opened 2023-11-03
Bens: add server

sevenzing:
Added server implementation over `bens-logic` crate


rimrakhimov (commented):
Would tests be added in later PRs?

lok52 (approved):

PR #677, opened 2023-11-13
verifier: API template for additional actions added into verification process

rimrakhimov:
An API enhancement suggestion that allows to support additional post-verification actions


sevenzing:
guess we need serde(default) for `post_actions` to be added in `build.rs`

lok52 (approved):

PR #680, opened 2023-11-14
Add materialized view for address name resolving

sevenzing:
I added materialized view for quick batch ENS name resolving. I decided to write view updating inside server (thanks [tokio-cron-scheduler](https://github.com/mvniekerk/tokio-cron-scheduler/tree/main)) rather than use `pg_cron`.

+ the feature can be enabled/disabled using `BENS__SUBGRAPH__CACHE_ENABLED=true/false` variable.
+ sheduler will refresh view according to `BENS__SUBGRAPH__REFRESH_CACHE_SCHEDULE` (6-items extended crontab format)

### DIFF
--- a/blockscout-ens/bens-logic/src/subgraphs_reader/sql/address_names.rs
+++ b/blockscout-ens/bens-logic/src/subgraphs_reader/sql/address_names.rs
@@ -2,6 +2,8 @@ use crate::subgraphs_reader::SubgraphReadError;
 use sqlx::{postgres::PgPool, Executor};
 use tracing::instrument;
 
+use super::{DOMAIN_DEFAULT_WHERE_CLAUSE, DOMAIN_NOT_EXPIRED_WHERE_CLAUSE};
+
 #[instrument(
     name = "create_address_names_view",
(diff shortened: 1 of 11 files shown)


blockscout-ens/bens-logic/src/subgraphs_reader/sql/address_names.rs
```diff
@@ -0,0 +1,87 @@
...
+        from {schema}.domain
+        where
+            resolved_address IS NOT NULL
+            AND label_name IS NOT NULL
+            AND name NOT LIKE '%[%'
+            AND (
+                expiry_date is null
+                OR to_timestamp(expiry_date) > now()
+            )
+            AND block_range @> 2147483647
```
rimrakhimov, line 33: It is better to reuse `DOMAIN_NOT_EXPIRED_WHERE_CLAUSE` and `DOMAIN_DEFAULT_WHERE_CLAUSE` constants to make it consistent between the view and the actual  query

rimrakhimov (approved):

PR #712, opened 2023-12-20
[LAUNCHER] Ignore HTTP incoming and processed log messages for '/health' endpoint

rimrakhimov:


libs/blockscout-service-launcher/src/launcher/span_builder.rs
```diff
@@ -2,21 +2,48 @@ use actix_web::{
...
     Lazy::new(|| Mutex::new(HashMap::new()));
 
+static SKIP_HTTP_TRACE_PATHS: Lazy<RwLock<HashSet<String>>> =
+    Lazy::new(|| RwLock::new(HashSet::new()));
+
 #[derive(Default)]
 pub struct CompactRootSpanBuilder;
 
+impl CompactRootSpanBuilder {
+    pub fn init_skip_http_trace_paths<S: Into<String>>(
```
sevenzing, line 26: why write this function inside CompactRootSpanBuilder? 👀

sevenzing (approved):

libs/blockscout-service-launcher/src/launcher/span_builder.rs
rimrakhimov: Mostly to indicate that it is `CompactRootSpanBuilder` that uses those paths to be skipped

Issue #723, opened 2023-12-26
Bens: add pagination and change API


sevenzing:
Frontend team asked to add pagination for `/api/v1/:chain_id/domains:lookup`, add feature to return all domains if `name=null` and change method from POST to GET

- [x] Change api in bens
https://github.com/blockscout/blockscout-rs/pull/722
- [x] Fix blockscout integration made in https://github.com/blockscout/blockscout/pull/8972
https://github.com/blockscout/blockscout/pull/9062
- [x] Add pagination and full domain search
https://github.com/blockscout/blockscout-rs/pull/724

PR #724, opened 2023-12-26
[BENS] Pagination and list of all domains

sevenzing:
https://github.com/blockscout/blockscout-rs/issues/723


blockscout-ens/bens-logic/src/subgraphs_reader/pagination/paginator.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub next_page_token: Option<String>,
+}
+
+impl<I> PaginatedList<I> {
+    pub fn empty() -> Self {
+        Self {
+            items: vec![],
+            next_page_token: None,
+        }
+    }
```
rimrakhimov, line 30: Nit: maybe just derive a `Default` trait over the struct?

rimrakhimov (approved):

blockscout-ens/bens-logic/src/subgraphs_reader/pagination/paginator.rs
sevenzing: Then we need to define Default for Vec<I> (for example default for Vec<Domain> cannot be derived since Domain doesnt implement Default)
image

sevenzing: ¯\_(ツ)_/¯

PR #725, opened 2023-12-26
Smart-contract-verifier http client

### LINKED ISSUE #709: [ETH_BYTECODE_DB] http client to connect to smart-contract-verifier
We would like to use HTTP requests to `smart-contract-verifier` service instead of grpc connection we are using now to proxy verification requests. 

Currently used grpc connection imposes some limitations: 1) we cannot load balance requests to the verification service; 2) implemented tracing schema with `request_id` depends on the request arriving to the HTTP actix server. Both limitations may be worked around but that would require quite an effort on both Rust and Devops teams. So, for now the easier approach would be to start using HTTP instead of GRPC as a communication means between two services.

Subtasks:
- [ ] Implement a basic `smart-contract-verifier` HTTP client
- [ ] Update `eth-bytecode-db` to use HTTP client instead of GRPC.

rimrakhimov:
Closes #709


eth-bytecode-db/eth-bytecode-db-server/src/settings.rs
```diff
@@ -48,7 +48,15 @@ pub struct DatabaseSettings {
 #[serde(deny_unknown_fields)]
 pub struct VerifierSettings {
     #[serde_as(as = "DisplayFromStr")]
-    pub uri: tonic::transport::Uri,
+    pub http_url: url::Url,
+    #[serde(default = "default_verifier_max_retries")]
+    pub max_retries: u32,
+    #[serde(default)]
+    pub validate_url: bool,
```
sevenzing, line 55: validate url means that you check that provided url is indeed a URL. But you are sending health request, so I suggest to call it`check_health`

sevenzing (approved):

eth-bytecode-db/eth-bytecode-db-server/src/settings.rs
rimrakhimov: Updated to `probe_url`

PR #746, opened 2024-01-17
[ETH_BYTECODE_DB] Support for genesis contracts in verifier alliance db

rimrakhimov:
Apply the following rule for the contracts added in genesis or hardcoded during fork, as those do not have valid transaction hash

> in the case of a "genesis" contract, the transaction_hash should be set
        to keccak256(creation_code_hash || runtime_code_hash). this is because the transaction_hash
        needs to differ to distinguish between two versions of the same genesis contract, and so
        it needs to embed inside it the only feature that changes.

### DIFF
--- a/eth-bytecode-db/eth-bytecode-db/src/verification/handlers/mod.rs
+++ b/eth-bytecode-db/eth-bytecode-db/src/verification/handlers/mod.rs
@@ -375,16 +375,12 @@ async fn process_verifier_alliance_db_action(
                 .context("Converting source into database ready version")?;
 
             // At least one of creation and runtime code should exist to add the contract into the database.
-            if creation_code.is_none() && runtime_code.is_none() {
-                anyhow::bail!("Both creation and runtime codes are nulls")
-            }
-
             let transaction_hash = derive_transaction_hash(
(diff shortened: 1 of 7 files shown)


eth-bytecode-db/eth-bytecode-db/src/verification/handlers/mod.rs
```diff
@@ -338,33 +379,27 @@ async fn process_verifier_alliance_db_action(
...
 
-            // TODO: make transaction_hash input argument optional, and calculate it
-            //       as `keccak256(creation_code || runtime_code)` in that case
-            let transaction_hash = transaction_hash.to_vec();
+            let transaction_hash = derive_transaction_hash(
+                transaction_hash.clone(),
+                creation_code.clone(),
+                runtime_code.clone(),
+            )
+            .expect("At least one of creation or runtime codes must exist");
```
sevenzing, line 387: I would remove this expect and bail on L379 and add `ok_or_else(|| anyhow::anyhow("Both creation and runtime codes are nulls"))?` to derive_transaction_hash

sevenzing (approved):

PR #753, opened 2024-01-21
Unify protoc dependencies between services

rimrakhimov:


docker/base.Dockerfile
```diff
@@ -0,0 +1,17 @@
+ARG CARGO_CHEF_VERSION=0.1.62-rust-1.75-buster
+FROM lukemathwalker/cargo-chef:${CARGO_CHEF_VERSION} as chef
+
+ARG PROTOC_VERSION=25.2
+ARG PROTOC_GEN_OPENAPIV2_VERSION=2.19.0
+
+RUN apt-get update && apt-get install -y curl wget unzip
+RUN wget https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-linux-x86_64.zip -O ./protoc.zip \
```
sevenzing, line 8: Add CASE to handle amd platform, you can find example in BENS or admin services

.github/workflows/services-base.yml
```diff
@@ -0,0 +1,123 @@
...
+          CONTENT=$(cat $DOCKERFILE_PATH)
+    
+          nl=$'\n'
+    
+          if [[ $CONTENT =~ ARG\ CARGO_CHEF_VERSION=([^$nl]+) ]]; then
+            CARGO_CHEF_TAG=${BASH_REMATCH[1]}
+          else
+            echo "Failed to extract CARGO_CHEF_TAG from Dockerfile"
+            exit 1
+          fi
```
k1rill-fedoseev, line 42: Do you think it's better to define versions in env and pass them down to docker using build args, instead of parsing a Dockerfile like this?

rimrakhimov: I'm not sure, tbh. I like the current approach because all arguments are explicit and located in the same place as their usage. But I'm not insisting on that

sevenzing (approved):

PR #760, opened 2024-01-31
[ETH_BYTECODE_DB] Search into verifier alliance

rimrakhimov:
Adds the endpoints that allow to search inside the verifier alliance database if the one is present


eth-bytecode-db/eth-bytecode-db-proto/proto/v2/eth-bytecode-db.proto
```diff
@@ -223,6 +231,9 @@ message SearchAllSourcesRequest {
   string chain = 3;
   /// The address of the contract being verified
   string address = 4;
+  /// If set and is `false`, will not try to obtain sources from 3rd party services
+  /// (e.g., Sourcify and Etherscan)
+  optional bool only_local = 5;
```
sevenzing, line 236: if `only_local=false` it **will not** go to Sourcify and Etherscan ? maybe it **will** go?

sevenzing: and why `optional`? you need default value to be `true`? in this case let's just reverse meaning of variable to be `search_remote`

rimrakhimov: Nope, the default value is set to be `false`, you can find it in the endpoint handler further. I've added `optional` here merely so to show that users may not specify that value by default. In which case, it will default to `false` and the behaviour will not change from the current one

rimrakhimov: Do you suggest to remove optional, and set it to default to `false` in case if missing in the `build.rs`?

sevenzing: oh, I see. Well, actually I don't know what is better, maybe writing `optional` to show that this field is actually optional is better than add `serde::default` in `build.rs` :)

eth-bytecode-db/eth-bytecode-db-server/src/services/database.rs
```diff
@@ -74,6 +75,27 @@ impl Database for DatabaseService {
...
+            chain_id = request.chain,
+            contract_address = request.address
+        );
+
+        let mut sources = vec![];
+        if let Some(alliance_db_client) = self.client.alliance_db_client.clone() {
+            sources = self
+                .search_alliance_sources(alliance_db_client, &request.chain, &request.address)
+                .await?;
+        }
```
sevenzing, line 94: I'm not sure about this, but maybe we have to return 500 or 404 if `alliance_db_client` is not set..?

rimrakhimov: Updated it to return `unavailable` for those cases, for now

eth-bytecode-db/eth-bytecode-db-server/src/services/database.rs
```diff
@@ -74,6 +75,27 @@ impl Database for DatabaseService {
...
+        let request = request.into_inner();
+        super::trace_request_metadata!(
+            chain_id = request.chain,
+            contract_address = request.address
+        );
+
+        let mut sources = vec![];
+        if let Some(alliance_db_client) = self.client.alliance_db_client.clone() {
+            sources = self
+                .search_alliance_sources(alliance_db_client, &request.chain, &request.address)
```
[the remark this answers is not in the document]
rimrakhimov: Added `_internal` suffix

sevenzing (approved):

PR #775, opened 2024-02-12
Initialize blockscout-client lib (http client)

rimrakhimov:


libs/blockscout-client/src/client.rs
```diff
@@ -0,0 +1,112 @@
...
+    base_url: Url,
+    api_sensitive_endpoints_key: Option<String>,
+    request_client: ClientWithMiddleware,
+}
+
+impl Client {
+    pub fn new(config: config::Config) -> Self {
+        let config = match config::ValidatedConfig::try_from(config) {
+            Ok(config) => config,
+            Err(err) => panic!("Invalid blockscout client configuration: {err}"),
```
sevenzing, line 17: why panic?

libs/blockscout-client/src/client.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    pub(crate) fn build_url_with_query(&self, path: &str, query: Option<&str>) -> Url {
+        let mut url = self.base_url.clone();
+        url.set_path(path);
+        url.set_query(query);
+        url
+    }
+
+    pub(crate) async fn _post_request<U, Request, Response>(
```
sevenzing, line 56: why _ ?

libs/blockscout-client/src/client.rs
```diff
@@ -0,0 +1,112 @@
...
+    ) -> Result<Response>
+    where
+        Response: serde::de::DeserializeOwned,
+    {
+        match response {
+            Ok(response) if response.status().is_success() => {
+                let full = response.bytes().await?;
+
+                let jd = &mut serde_json::Deserializer::from_slice(full.as_ref());
+                serde_path_to_error::deserialize(jd).map_err(|err| Error::Decode(Box::new(err)))
```
sevenzing, line 106: 👍

libs/blockscout-client/src/v2/import.rs
```diff
@@ -0,0 +1,23 @@
...
+        address_hash: ethers_core::types::Address,
+    ) -> Result<types::ImportSmartContractResponse> {
+        let path = format!(
+            "/api/v2/import/smart-contracts/0x{}",
+            hex::encode(address_hash)
+        );
+        let headers = match client.api_sensitive_endpoints_key() {
+            None => vec![],
+            Some(key) => vec![("x-api-key", key)],
+        };
```
sevenzing, line 18: I would suggest to make logic of adding `x-api-key` to be more abstract to easily add this logic for new endpoints, however it's okay to leave it for now

libs/blockscout-client/src/client.rs
rimrakhimov: We do not use it for now and as it is not public it should be used somewhere in order not to cause the compilation warning. However it was just copied from the client implementation of `smart-contract-verifier`, and as we will extend that client further, I believe it makes sense to leave it so it can be easily used when required

libs/blockscout-client/src/v2/import.rs
rimrakhimov: Could you please explain what do you mean by more abstract?

sevenzing: I mean more declarative
Maybe building request like `client.get().with_sensitive_endpoints_key().with_headers(headers).build(url)`

sevenzing: and `with_sensitive_endpoints_key` will handle your `match` for actual key

sevenzing: but we can leave it for next version I guess ;)

sevenzing (approved):

libs/blockscout-client/src/v2/import.rs
rimrakhimov: I see
I agree, that looks nice to have, but I'm also for moving it to the next iteration 👍

PR #783, opened 2024-02-21
Make use of foundry-compilers instead of ethers-solc

rimrakhimov:
`ethers-solc` library has not added support for the "Cancun" evm version. This PR starts using the `foundry-compilers` crate instead of deprecated `ethers-solc` for input related structs. 

Output related structs continues to use `ethers-solc`, as there are still some problems with returning proper abis via `foundry-compilers`


smart-contract-verifier/smart-contract-verifier-server/Cargo.toml
```diff
@@ -10,6 +10,10 @@ smart-contract-verifier = { path = "../smart-contract-verifier" }
 smart-contract-verifier-proto = { path = "../smart-contract-verifier-proto" }
 sig-provider-extension = { path = "../sig-provider-extension", optional = true }
 
+alloy-dyn-abi = "0.6.3"
+alloy-json-abi = "0.6.3"
```
sevenzing, line 14: Not used I guess

sevenzing (approved):

PR #802, opened 2024-03-13
[VERIFIER] Batch solidity contracts verification

rimrakhimov:
Introduces two new endpoints for solidity verifier: `BatchVerifyMultiPart` and `BatchVerifyStandardJson`. They add 2 main features:
1. Effectively verify several bytecodes against the same input files. Before, several separate requests had to be made for such cases, resulting the same sources were recompiled per each request. Using new endpoints allows users to compile the sources once, and compare all bytecodes against those compilation results.
2. Compare both creation and runtime codes at the same request. Usual verification endpoints require users to provide only one bytecode and specify whether that is creation or runtime code


smart-contract-verifier/smart-contract-verifier-proto/proto/v2/smart-contract-verifier.proto
```diff
@@ -250,4 +254,118 @@ message LookupMethodsResponse {
...
+  string compiler_version = 4;
+
+  message Language {
+    enum Language {
+      UNSPECIFIED = 0;
+      SOLIDITY = 1;
+      YUL = 2;
+      VYPER = 3;
+    }
+  }
```
sevenzing, line 340: why `Language.Language` ?

sevenzing: maybe remove `message Language` ?

smart-contract-verifier/smart-contract-verifier-server/src/types/batch_verification.rs
```diff
@@ -0,0 +1,156 @@
...
+    Ok(Response::new(response))
+}
+
+fn process_verification_result(
+    result: BatchVerificationResult,
+) -> Result<proto::ContractVerificationResult, Status> {
+    let verification_result = match result {
+        BatchVerificationResult::Failure(_) => {
+            proto::contract_verification_result::VerificationResult::Failure(
+                proto::ContractVerificationFailure {},
```
sevenzing, line 105: so we did't give any reason why verification failed, why?

smart-contract-verifier/smart-contract-verifier-proto/proto/v2/smart-contract-verifier.proto
rimrakhimov: Without message wrapper there is a conflict between `Language.VYPER` and `Compiler.VYPER` values. Thus, another option is to add `LANGUAGE_` prefix to all values which I personally like less than this approach

smart-contract-verifier/smart-contract-verifier-server/src/types/batch_verification.rs
rimrakhimov: There is no main reason why verification failed. Different contracts resultant from compilation have different reasons. So, if verification failed it basically means that no contract corresponded to the bytecodes we are trying to verify (neither creation nor runtime), and there is no more details we can provide.

I had an idea to return the particular reason for each contract compiled but for now it seemed too much, imo

smart-contract-verifier/smart-contract-verifier/src/batch_verifier/batch_contract_verifier.rs
```diff
@@ -0,0 +1,245 @@
...
+
+        let (does_creation_match, creation_values, creation_transformations) = match &contract
+            .creation_code
+        {
+            Some(contract_code) => {
+                match transformations::process_creation_code(
+                    contract_code,
+                    parsed_contract.creation_code.to_vec(),
+                    &parsed_contract.compilation_artifacts,
+                    serde_json::to_value(parsed_contract.creation_code_artifacts.clone()).unwrap(),
```
sevenzing, line 127: I suggest to replace this logic of converting to json and parsing json with enum :)

sevenzing (approved):

PR #812, opened 2024-03-27
Initialize smart-guessr service

rimrakhimov:


whatsabi/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+            path: '/api/v1/swagger'
+        }))
+        .use(
+            logger({
+                level: "info",
+            })
+        )
+        .get("/api/v1/abi", async ( { request, log, query }) => {
+            log.info(request, "New request");
+            return await processAbi(query.address, query.provider)
```
sevenzing, line 21: maybe add some log to indicate that for `address` we found abi with N amount of entries

.github/workflows/whatsabi.yml
```diff
@@ -0,0 +1,92 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - uses: actions-ecosystem/action-regex-match@v2
+        id: regex
+        with:
+          text: ${{ github.ref }}
+          regex: '^(refs\/tags\/{{project-name}}\/(v\d+\.\d+\.\d+))|(refs\/heads\/(main))$'
```
sevenzing, line 60: `{{project-name}} -> whatsabi`

sevenzing (approved):

whatsabi/src/index.ts
sevenzing: or `address not found`
or `no bytecode for address`
and so on

whatsabi/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+                if (!ethers.isAddress(query.address)) {
+                    return error(400, "Invalid address value")
+                }
+            }
+        })
+        .listen(port);
+}
+
+let app = initApp(port);
+export type App = typeof app;
```
lok52, line 40: Unused?

whatsabi/src/index.ts
```diff
@@ -0,0 +1,68 @@
...
+    `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`
+);
+
+async function processAbi(address: string, provider_url: string) {
+    const provider = ethers.getDefaultProvider(provider_url);
+    const signatureLookup = new whatsabi.loaders.MultiSignatureLookup([
+        new whatsabi.loaders.OpenChainSignatureLookup(),
+        new whatsabi.loaders.SamczunSignatureLookup(),
+        new whatsabi.loaders.FourByteSignatureLookup(),
+    ]);
```
lok52, line 52: Probably could be extracted as a global const to prevent repeated object creation

lok52 (approved):

whatsabi/src/index.ts
rimrakhimov: Btw, in case if address is not a contract, then just empty array will be returned as a result and not an error

PR #816, opened 2024-03-30
[USER_OPS_INDEXER] Support geth traces via debug module

k1rill-fedoseev:
Geth doesn't support `trace_*` JSON RPC methods (all other nodes should support them though). Indexing some rare user operations requires tracing capabilities, so it makes sense to automatically select use either `trace_*` or `debug_*` methods based on the underlying JSON RPC node type.


user-ops-indexer/user-ops-indexer-logic/src/indexer/rpc_utils.rs
```diff
@@ -0,0 +1,158 @@
...
+                to: frame.to.clone().and_then(|to| to.as_address().cloned()),
+                input: frame.input.clone(),
+            });
+        }
+        if let Some(calls) = &frame.calls {
+            if calls.len() > idx {
+                path.push((&frame, idx + 1));
+                path.push((&calls[idx], 0));
+            }
+        }
```
rimrakhimov, line 128: Maybe I'm missing something, but why to make such complex schema for adding all subcalls instead of something like this:
```suggestion
        if let Some(calls) = frame.calls {
            path.extend(calls)
        }
        res.push(CommonCallTrace {
            typ: match frame.typ.as_str() {
                "CALL" => TraceType::Call,
                "CALLCODE" => TraceType::CallCode,
                "STATICCALL" => TraceType::StaticCall,
                "DELEGATECALL" => TraceType::DelegateCall,
                "CREATE" => TraceType::Create,
                _ => TraceType::Other,
            },
            from: frame.from,
            to: frame.to.and_then(|to| to.as_address().cloned()),
            input: frame.input,
        });
```

rimrakhimov (approved):

user-ops-indexer/user-ops-indexer-logic/src/indexer/rpc_utils.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    res
```
rimrakhimov, line 131: Maybe I'm missing something, but why to make such complex schema for adding all subcalls instead of something like this:

```suggestion
    let mut path = Vec::from([root]);
    let mut res = Vec::new();

    while let Some(frame) = path.pop() {
        if let Some(calls) = frame.calls {
            path.extend(calls)
        }
        res.push(CommonCallTrace {
            typ: match frame.typ.as_str() {
                "CALL" => TraceType::Call,
                "CALLCODE" => TraceType::CallCode,
                "STATICCALL" => TraceType::StaticCall,
                "DELEGATECALL" => TraceType::DelegateCall,
                "CREATE" => TraceType::Create,
                _ => TraceType::Other,
            },
            from: frame.from,
            to: frame.to.and_then(|to| to.as_address().cloned()),
            input: frame.input,
        });
    }

    res
```

k1rill-fedoseev: I think such approach will mess up with the order of flattened traces. E.g. in the following example:
```
U calls A
A calls B, C
B calls D, E
C calls F, G
```
We want to produce the following flatted result:
```
U => A
A => B
B => D
B => E
A => C
C => F
C => G
```

You suggestion will produce:

```
U => A
A => B
A => C
B => D
B => E
C => F
C => G
```

user-ops-indexer/user-ops-indexer-logic/src/indexer/rpc_utils.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+fn flatten_geth_trace(root: CallFrame) -> Vec<CommonCallTrace> {
+    let mut path = Vec::from([(&root, 0)]);
+    let mut res = Vec::new();
+
+    while let Some((frame, idx)) = path.pop() {
+        if idx == 0 {
+            res.push(CommonCallTrace {
+                typ: match frame.typ.as_str() {
```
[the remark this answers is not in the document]
k1rill-fedoseev: Yeah, indeed, looks like in Geth it's possible, but not in others. Added a `"CREATE2" => TraceType::Create` line. Thanks!

user-ops-indexer/user-ops-indexer-logic/src/indexer/rpc_utils.rs
rimrakhimov: Get it 👌

rimrakhimov: Btw, just interested why the order is important in our case? And also, wdyt about adding a test case that checks the order as well?

PR #823, opened 2024-04-04
Basic cache implementation

bragov4ik:
https://www.notion.so/blockscout/The-First-Project-df4d82c0c9a445bcac009daf2b1f9581

TODO:
- (more) tests
- self review


basic-cache/basic-cache-proto/proto/v1/basic-cache.proto
```diff
@@ -0,0 +1,47 @@
...
+
+message SourceFile {
+  string name = 1;
+  string content = 2;
+}
+
+message SmartContract {
+  string url = 1
+    [ (convert_options.convert) = {type: "url::Url"} ];
+  repeated SourceFile sources = 2;
```
sevenzing, line 30: why not HashMap to prevent `name` duplicates

bragov4ik:
probably this https://github.com/blockscout/blockscout-rs/pull/824 should fix the build 👀

basic-cache/basic-cache-logic/src/in_memory_cache.rs
```diff
@@ -0,0 +1,68 @@
+use std::{collections::HashMap, hash::Hash, sync::Mutex};
+
+use crate::CacheManager;
+
+#[derive(Debug)]
+pub struct HashMapCache<K, V> {
+    // need interior mutability because usual work with DB is done
+    // via immutable reference
+    inner: Mutex<HashMap<K, V>>,
```
lok52, line 9: Maybe RwLock would be better?

basic-cache/basic-cache-proto/proto/v1/basic-cache.proto
bragov4ik: With `map` sending duplicate filenames is silently handled and one of the files is dropped. It can't be checked the same way as it was with `repeated`, so I can't throw an error during validation.

Won't `repeated` be better in this case?

basic-cache/basic-cache-logic/src/in_memory_cache.rs
bragov4ik: There's no shared ownership of `Mutex`, and the code is single threaded. No difference I guess?

bragov4ik: in general case I see what you mean though :)

PR #825, opened 2024-04-05
Update versions of github actions (+ test fix)

bragov4ik:
There are some warnings in CI that should be fixed by bumping actions versions.
example:
https://github.com/blockscout/blockscout-rs/actions/runs/8560946122/job/23461168368?pr=823#step:8:23

the update should be trivial as the updated actions have barely any options and seem to do elementary stuff

### DIFF
--- a/.github/workflows/bens.yml
+++ b/.github/workflows/bens.yml
@@ -150,7 +150,7 @@ jobs:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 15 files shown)


rimrakhimov (approved):
Lgtm 👍 

Just interested, maybe you know the way to extract common workflow steps into a separate workflow which can be reused further? So, that such changes can be made only in the original workflow file in the future

bragov4ik:
> maybe you know the way to extract common workflow steps into a separate workflow which can be reused further?

AFAIR this is what we need:
https://docs.github.com/en/actions/using-workflows/reusing-workflows

sounds good, we should do it some time 👀

rimrakhimov:
Let's merge this PR, and then check if we can apply reusable workflows to our setup when we have time

PR #831, opened 2024-04-10
[DA_INDEXER] Celestia blobs indexer

AllFi:
This PR introduces a simple indexer for Celestia blobs and a service to fetch blobs without the necessity to know a namespace beforehand. The Celestia indexer runs on top of the Celestia light node.

The name da-indexer was chosen because this service might potentially be extended to support other DA solutions (Avail, Eigen DA, etc.).

There are two potential use cases for this service:
1. This service might be useful for indexing Optimism-like rollups that use Celestia as a DA layer. The L1 transactions for this kind of rollup will include the height and commitment of the blob on Celestia. https://github.com/celestiaorg/optimism/blob/0ce4f69ef4fd113fae6f3ea0c99ea99a1dfc9c67/op-batcher/batcher/driver.go#L510
2. This service can also be used on UI to display the data of the batches in rollup settlement transactions on L1.

I've tested the indexer on the Celestia mainnet, and it seems to be working properly.


da-indexer/da-indexer-logic/src/celestia/indexer.rs
```diff
@@ -0,0 +1,206 @@
...
+struct Job {
+    height: u64,
+}
+
+pub struct Indexer {
+    client: Client,
+    db: Arc<DatabaseConnection>,
+    settings: IndexerSettings,
+
+    last_known_height: RwLock<u64>,
```
rimrakhimov, line 31: It may be better and easier to use `AtomicU64` instead

da-indexer/da-indexer-logic/src/celestia/rpc_client.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// A client using 'http\[s\]' protocol.
+    Http(HttpClient),
+    /// A client using 'ws\[s\]' protocol.
+    Ws(WsClient),
+}
+
+impl Client {
+    /// The maximum request size in the default client in celestia_rpc is not sufficient for some blocks,
+    /// therefore, we need to implement our own client with larger limits.
+    pub async fn new(
```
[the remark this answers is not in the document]
AllFi: Fixed, thank you for pointing it out!

da-indexer/da-indexer-logic/src/celestia/repository/blocks.rs
```diff
@@ -0,0 +1,92 @@
...
+    let mut gaps = Gap::find_by_statement(Statement::from_sql_and_values(
+        db.get_database_backend(),
+        r#"
+            SELECT height + 1 as gap_start, 
+                   next_nr - 1 as gap_end
+            FROM (
+                SELECT height, lead(height) OVER (ORDER BY height) as next_nr
+                FROM celestia_blocks WHERE height <= $1
+            ) nr
+            WHERE nr.height + 1 <> nr.next_nr;"#,
```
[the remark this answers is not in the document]
AllFi: It seems that it doesn't, I've added explicit ordering.

da-indexer/da-indexer-logic/src/celestia/indexer.rs
AllFi: Yes, it looks better this way.

da-indexer/da-indexer-migration/src/m20220101_000001_create_table.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let sql = r#"
+            CREATE TABLE "celestia_blobs" (
+                "id" bytea PRIMARY KEY,
+                "height" bigint NOT NULL,
```
rimrakhimov, line 12: Nit: just interested, why not to create a foreign key reference from `celestia_blobs.height` to `celestia_blocks.height`?

rimrakhimov (approved):

da-indexer/da-indexer-migration/src/m20220101_000001_create_table.rs
AllFi: The main reason why it is missing now is the order of writes in the db. Currently, we insert blobs before inserting the block's metadata. We can't reverse this order within the current implementation since we rely on the blocks table to determine missing data. I guess we might make these inserts atomic if you think that it is necessary.

AllFi: Fixed in [dee226b](https://github.com/blockscout/blockscout-rs/pull/831/commits/dee226bfe4f15405cb3dfb3b0354a2f436705296)

rimrakhimov (approved):

rimrakhimov:
Imo, looks great
Should we merge it?

AllFi:
Hi, @rimrakhimov! Thank you for the review!

The current state of the indexer should be sufficient to cover the first Blockscout-Celenium use case, so I guess we can merge it.

PR #832, opened 2024-04-11
[ETH_BYTECODE_DB] Alliance solidity batch import

rimrakhimov:
Adds two endpoints:
1. `/api/v2/alliance/solidity/multi-part:batch-import`
2. `/api/v2/alliance/solidity/standard-json:batch-import`

The endpoints allow to importing solidity contracts into verifier alliance database in batches. Make use of sc-verifier's batch solidity verify endpoints.

Were initially required to improve the efficiency of proxy-verifier service. So that multiple contracts that could be verified with the same sources, would only require the sources to be compiled once. That should improve the performance of such requests drastically as the compilation is usually the most demanding operation.


eth-bytecode-db/eth-bytecode-db-server/src/services/verifier_base.rs
```diff
@@ -26,6 +26,27 @@ pub fn process_verification_result(
...
+}
+
+pub fn process_batch_import_error(error: Error) -> tonic::Status {
+    match error {
+        Error::Internal(message) => {
+            tracing::info!(details=%message, "Internal error");
+            tonic::Status::internal(message.to_string())
+        }
+        Error::Verifier(message) => {
+            tracing::info!(details=%message, "Internal error");
```
sevenzing, line 43:
```suggestion
            tracing::error!(details=%message, "Internal verifier error");
```

eth-bytecode-db/eth-bytecode-db/src/verification/types.rs
```diff
@@ -309,6 +311,110 @@ impl TryFrom<Source> for DatabaseReadySource {
...
+    }
+
+    Ok(parts)
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn parse_code_parts() {
+        // let code_parts = serde_json
```
sevenzing, line 414: 🤔

sevenzing (approved):

PR #833, opened 2024-04-11
[BENS] Perform to_checksum to response addresses

sevenzing:
during batch resolving blockscout tries to fetch addresses from response map using checksum address, but BENS returns addresses in lower case (from graph-node database). This PR provides checksum address serialization according to chain_id (for rootstock 🥲 )

### DIFF
--- a/blockscout-ens/bens-server/src/conversion/mod.rs
+++ b/blockscout-ens/bens-server/src/conversion/mod.rs
@@ -10,14 +10,6 @@ mod events;
 pub use domain::*;
 pub use events::*;
 
-pub fn order_direction_from_inner(inner: proto::Order) -> Order {
-    match inner {
-        proto::Order::Unspecified => Order::Desc,
-        proto::Order::Asc => Order::Asc,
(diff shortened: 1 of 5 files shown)


blockscout-ens/bens-server/src/conversion/mod.rs
```diff
@@ -24,3 +26,24 @@ pub enum ConversionError {
...
+        hash: checksummed(&address, chain_id),
+    }
+}
+
+fn address_from_str_logic(addr: &str, chain_id: i64) -> Result<proto::Address, ConversionError> {
+    let addr = Address::from_str(addr)
+        .map_err(|e| ConversionError::LogicOutput(format!("invalid address '{addr}': {e}")))?;
+    Ok(proto::Address {
+        hash: checksummed(&addr, chain_id),
+    })
```
bragov4ik, line 48: nitpick: can use `address_from_logic` here (maybe with #[inline])

rimrakhimov (approved):

PR #834, opened 2024-04-11
[SCOUTCLOUD] add CRUD for deployments and start deploy

sevenzing:


scoutcloud/scoutcloud/src/logic/users/auth.rs
```diff
@@ -1,64 +1,144 @@
...
+        } else {
+            Err(AuthError::Unauthorized("no to the instance".to_string()))
+        }
+    }
+
+    pub async fn allowed_to_create_instance<C>(&self, db: &C) -> Result<(), AuthError>
+    where
+        C: ConnectionTrait,
+    {
+        if self.user.is_superuser {
```
rimrakhimov, line 85: Is zero balanced accounts allowed to create instances?

scoutcloud/scoutcloud/src/logic/users/auth.rs
```diff
@@ -1,64 +1,144 @@
...
+            return Ok(());
+        }
+        if self.user.balance.is_sign_negative() {
+            return Err(AuthError::InsufficientBalance);
+        }
+        if Instance::count(db, self).await? >= MAX_INSTANCES_PER_USER {
```
rimrakhimov, line 91: Here we limit the number of instances per token and not per user

scoutcloud/scoutcloud/src/logic/users/auth.rs
sevenzing: for now we give super user full authentication

rimrakhimov (approved):

PR #836, opened 2024-04-17
[STATS] Marking approximate points + general improvements

bragov4ik:
## Changes
Approximate points marking:
- New field for data point (`Point` in proto) indicating whether the data is approximate (whether it should be drawn as such). Currently all those marks are placed on trailing subsequent points.
- The field is omitted for `false` values in json serialization for efficiency. Absence of the field should be treated as `false` (meaning the data is not approximate == exact)

Noticeable changes in stats service:
- Regardless of requested interval, no future points are now returned.
  - Previously they were generated according to settings; e.g. if one were to request data for 10 days forward.
- Improved handling of near-today dates.
  - Previously was mitigated by manual removal of last point. It prevented new updated values from being shown until the end of the day. Now they will be returned as soon as they become available.
- Added limit for maximum requested interval; it was possible to request server to generate data points for all days for ~ 10 000 years, which might be vulnerable to DoS (easy to fix so added just in case). The limit is configurable, default value - 500 years (max request processing time change from ~200ms to ~50ms).

Other:
- Development docker compose for stats service
- Restructure in stats service source code to accomodate the changes
- Readability improvements
- Updated timestamp columns in stats db to consider timezone. Did not cause troubles yet; a preventive measure.
- Pass time as argument instead of calling `..::now()` during update (for better testability)


stats/stats/src/missing_date.rs
```diff
@@ -1,49 +1,44 @@
...
+///
+/// Note that values outside of the range can still affect the filled values.
+pub fn fill_and_filter_chart(
+    data: Vec<DateValue>,
     from: Option<NaiveDate>,
     to: Option<NaiveDate>,
     policy: MissingDatePolicy,
-) -> Result<Vec<DateValue>, DbErr> {
-    let mut data_request = chart_data::Entity::find()
-        .select_only()
```
bragov4ik, line 14: note that I removed this line while transferring the code. seems unnecessary

stats/justfile
```diff
@@ -26,7 +26,7 @@ test *args:
 
 test-with-db *args:
     -just db-port="{{test-db-port}}" db-name="" docker-name="{{docker-name}}-test" start-postgres
-    just db-port="{{test-db-port}}" db-name=""                                    test {{args}}
+    -just db-port="{{test-db-port}}" db-name=""                                    test {{args}}
```
sevenzing, line 29: so you want to remove test table if test failed? 
as an example tests in `sqlx` remove test data only if test succeed:
https://docs.rs/sqlx/latest/sqlx/attr.test.html#automatic-test-database-management-requires-migrate-feature

```
Test databases are automatically cleaned up as tests succeed, but failed tests will leave their databases in-place to facilitate debugging. Note that to simplify the implementation, panics are always considered to be failures, even for #[should_panic] tests.
```

stats/stats-proto/src/tests.rs
```diff
@@ -0,0 +1,39 @@
...
+
+const PRECISE_POINT_2: &str = r#"
+{
+    "date": "2024-03-14",
+    "value": "188542399"
+}
+"#;
+
+#[test]
+fn is_approximate_serialization() {
```
sevenzing, line 21: it looks like you are testing `serde_json` implementation. I think it's a little unnecessary, but you can leave it if you want

stats/stats/src/charts/chart.rs
```diff
@@ -42,43 +45,35 @@ pub trait Chart: Sync {
...
+    ///
+    /// Usually set to 1 for line charts. Currently they have resolution of
+    /// 1 day. Also, data for portion of the (last) day has to be recalculated
+    /// on the next day.
+    ///
+    /// I.e. for number of blocks per day, stats for current day (0) are
+    /// not complete because blocks will be produced till the end of the day.
+    ///    |===|=  |
+    /// day -1   0
+    fn approximate_until_updated(&self) -> u64 {
```
sevenzing, line 65: I didnt get why name of this variable `approximate_until_updated`. maybe something like `approximate_last_points` or `approximate_last_values_count` of `approximate_last_points_number`

stats/stats/src/charts/db_interaction/read.rs
```diff
@@ -84,29 +149,36 @@ async fn get_chart(
...
-    } else {
-        data_request
-    };
-    let data_request = if let Some(to) = to {
-        data_request.filter(chart_data::Column::Date.lte(to))
-    } else {
-        data_request
+    if let Some(from) = from {
+        let custom_where = Expr::cust_with_values::<sea_orm::sea_query::Value, _>(
+            "date >= (SELECT COALESCE(MAX(date), '1900-01-01'::date) FROM chart_data WHERE chart_id = $1 AND date <= $2)",
```
sevenzing, line 160: could you please add comment here, I didnt get why do we need this sql here

stats/stats/src/charts/counters/mock.rs
```diff
@@ -25,7 +29,7 @@ impl ChartFullUpdater for MockCounter {
         _blockscout: &DatabaseConnection,
     ) -> Result<Vec<DateValue>, UpdateError> {
         let item = DateValue {
-            date: chrono::offset::Local::now().date_naive(),
+            date: NaiveDate::parse_from_str("2022-11-12", "%Y-%m-%d").unwrap(),
```
sevenzing, line 32: [parse_from_rfc3339](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.parse_from_rfc3339) ?

stats/stats/src/charts/db_interaction/read.rs
```diff
@@ -126,11 +198,25 @@ mod tests {
             charts::ActiveModel {
                 name: Set(TotalBlocks::default().name().to_string()),
                 chart_type: Set(ChartType::Counter),
+                last_updated_at: Set(Some(
+                    DateTime::parse_from_str(
```
sevenzing, line 202: [parse_from_rfc3339](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.parse_from_rfc3339) ?

stats/stats-proto/src/tests.rs
bragov4ik: more like integration between `actix_prost` and `serde_json` in this case)
yeah, the intention is to check if the field is serialized conditionally as expected; maybe it's possible to do it in some serializer-agnostic way (but rn it's simple and it works)

stats/stats/src/charts/chart.rs
bragov4ik: I didn't think about trimming future points while naming it; good catch ty

stats/stats/src/charts/db_interaction/read.rs
```diff
@@ -57,24 +62,84 @@ pub async fn get_counters(
     Ok(counters)
 }
 
+/// Mark corresponding data points as approximate.
+///
+/// Approximate are:
+/// - points after `last_updated_at` date
+/// - `mark_trailing_count` dates starting from `last_updated_at` and moving back
```
sevenzing, line 69: `mark_trailing_count` -> `mark_before_updated`

sevenzing (approved):

PR #837, opened 2024-04-17
[VERIFIER] Mark as partial matches for contracts without metadata hash

rimrakhimov:
Currently for contracts where metadata hash is absent in bytecode all matches are full. But in general we cannot verify the originality of sources for such contracts. That creates situations when the contract has been verified with not original bytecode but cannot be re-verified because the match is full. That PR makes the verifier to always return partial match for such contracts.

Currently yul and vyper contracts have no metadata hash included in the bytecode at all. For solidity contracts, it is possible to remove that data from the bytecode via setting either `"bytecodeHash": "none"` or `"appendCBOR": false`

### DIFF
--- a/smart-contract-verifier/smart-contract-verifier-server/tests/integration/transformations.rs
+++ b/smart-contract-verifier/smart-contract-verifier-server/tests/integration/transformations.rs
@@ -46,9 +46,10 @@ async fn libraries_manually_linked() {
     test_success_multi_part_and_standard_json("libraries_manually_linked").await
 }
 
-// TODO: no auxdata is parsed right now when `metadataHash` is "none"
 #[tokio::test]
 async fn metadata_hash_absent() {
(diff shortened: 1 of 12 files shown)


smart-contract-verifier/smart-contract-verifier-server/tests/integration/transformations.rs
```diff
@@ -47,10 +47,23 @@ async fn libraries_manually_linked() {
 }
 
 // TODO: no auxdata is parsed right now when `metadataHash` is "none"
```
bragov4ik, line 49: Is it still relevant?

bragov4ik: I mean "todo" part

sevenzing (approved):

bragov4ik (approved):

PR #846, opened 2024-04-24
[SCOUTCLOUD] Add jobs to handle starting and stopping instances

sevenzing:
Add fang worker to handle starting and stopping deployments. add balance checker to watch for unpaid deployments


scoutcloud/scoutcloud/src/logic/deploy/handlers/update_status.rs
```diff
@@ -1,41 +1,42 @@
...
+    let result = handle_instance_action(db, github, runner, instance, action, user_token).await?;
     Ok(result)
 }
 
-async fn handle_instance_action<C>(
-    db: &C,
-    github: &GithubClient,
+async fn handle_instance_action(
+    db: &DatabaseConnection,
+    _github: &GithubClient,
```
lok52, line 34: Why do we need it?

lok52 (approved):

scoutcloud/scoutcloud/src/logic/jobs/balance.rs
```diff
@@ -0,0 +1,272 @@
...
+        let unpaid_list = UnpaidDeployment::all(&tx).await.map_err(DeployError::Db)?;
+        if unpaid_list.is_empty() {
+            return Ok(());
+        }
+        tracing::info!(unpaid = ?unpaid_list, "found {} unpaid deployments. start processing", unpaid_list.len());
+        for unpaid in unpaid_list {
+            if unpaid.creator_can_pay(&tx).await.map_err(DeployError::Db)? {
+                unpaid.mark_as_paid(&tx).await.map_err(DeployError::Db)?;
+            } else {
+                tracing::warn!(
```
bragov4ik, line 55: Shouldn't it be info, since the service works as intended?

scoutcloud/scoutcloud/src/logic/jobs/balance.rs
```diff
@@ -0,0 +1,272 @@
...
+    creator_id: i32,
+}
+
+impl UnpaidDeployment {
+    pub fn hours(&self) -> i32 {
+        self.total_used_hours - self.total_paid_hours
+    }
+
+    pub fn expense_amount(&self) -> Decimal {
+        self.cost_per_hour.mul(Decimal::new(self.hours() as i64, 0))
```
bragov4ik, line 96: nit: use `saturating_mul`; it does not panic
(it seems that it's not possible in the case, thus nit c:)

scoutcloud/scoutcloud-migration/src/m20240208_092748_create_triggers.rs
```diff
@@ -49,15 +49,27 @@ impl MigrationTrait for Migration {
...
+                    UPDATE deployments
+                    SET total_cost = total_cost + NEW.expense_amount
+                    WHERE id = NEW.deployment_id;
                 ELSIF TG_OP = 'UPDATE' THEN
                     UPDATE users
                     SET balance = balance - (NEW.expense_amount - OLD.expense_amount)
                     WHERE id = NEW.user_id;
+
+                    UPDATE deployments
+                    SET total_cost = total_cost + (NEW.expense_amount - OLD.expense_amount)
```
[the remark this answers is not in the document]
sevenzing: in order to change server size you have to redeploy your instance, so it wont be possible.
also we calculate expenses every hour, not every 10 days

scoutcloud/scoutcloud-migration/src/m20240208_092748_create_triggers.rs
```diff
@@ -49,15 +49,27 @@ impl MigrationTrait for Migration {
                     UPDATE users
                     SET balance = balance - NEW.expense_amount
                     WHERE id = NEW.user_id;
+
```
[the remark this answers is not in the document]
sevenzing: well function calls are part of transaction they are called from, so I don't think it's possible to start new transaction here. if you know how to do it let me know

scoutcloud/scoutcloud/src/logic/jobs/balance.rs
sevenzing: well, if we have situation when `cost_per_hour * price` should panic, I want my code to panic. not saturating the result and pretending that everything is ok

scoutcloud/scoutcloud/src/logic/deploy/handlers/update_status.rs
sevenzing: so called "legacy"

scoutcloud/scoutcloud-migration/src/m20240208_092748_create_triggers.rs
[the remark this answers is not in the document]
bragov4ik: aaahhh ok, [triggers execute within the transaction that caused it](https://arc.net/l/quote/xtatlxjo)
I see, didn't know it 🙃

Issue #841, opened 2024-04-19
stats: optimism total contracts bug


yvaskov:
blockscout api:

![Image](https://github.com/blockscout/blockscout-rs/assets/41516657/cf437dc7-55ba-44cc-98d7-f62115e39cd5)

stats api:

![Image](https://github.com/blockscout/blockscout-rs/assets/41516657/bab08524-cbd0-4f44-94a9-76ddd4f4b541)

## reason
optimism cannot index all internal transactions and `total_contracts` uses `new_contracts` chart (like cache)

## solution

write `total_contracts` as SQL not dependent counter

bragov4ik:
> Total verified contracts affected as well

![Image](https://github.com/blockscout/blockscout-rs/assets/8144358/53796565-dd50-4ea8-91e9-b408a8d2c3ca)

![Image](https://github.com/blockscout/blockscout-rs/assets/8144358/e88cc5cd-2f6f-41ed-9797-086f2c4b019b)

@yvaskov seems to be in sync, isn't it?

yvaskov:
@bragov4ik yeah, it caught up.

bragov4ik:
We discussed the problem and possible solutions, so I'll summarize it here.
There are 2 main ways to solve the problem (and 1 kinda funny):
1. Calculate total contracts independently from `new_contracts` using `addresses` table.
**+**: quick fix, works, seems fast to query
**-**: `new_contracts` remains inaccurate, query complexity increases cumulatively, additional load on db (since additional query is performed)
2. Fix `new_contracts` (probably using `updated_at` field)
**+**: total contracts is consistent with blockscout's (to some degree), sql query results are getting reused
**-**: complex logic (will take time to implement)
3. Request the counter value from blockscout (semantically the same as 1.)
**+**: `new_contracts` remains inaccurate, full consistency with blockscout data, reusing sql results
**-**: likely to be inconsistent with `new_contracts` (once it's fixed), kinda defers a responsibility of stats service

I think it makes sense to first implement 1. and later focus on 2.
(we do not utilize 3. at the end 🙃 )

bragov4ik:
btw about consistency of calculation methods:

There are two methods: via transaction lists and via `addresses`. I ran queries to check differences between results of them (on optimism sepolia; earlier today; queries are listed below):
- contracts found with `addresses` but not with transactions: **2069**
- contracts found with transactions but not with `addresses`: **0**

Thus, `addresses` seem to be a more accurate method.

[...]

PR #847, opened 2024-04-25
[STATS] Change TotalContracts counting method

bragov4ik:
Closing #841


bragov4ik:
Query seems to work when lauching locally

PR #851, opened 2024-04-26
[VERIFIER, ETH_BYTECODE_DB] Support for vyper blueprint contracts

### LINKED ISSUE #843: Support for blueprint (EIP-5202) contracts
Blueprint contracts are defined in [EIP-5202](https://eips.ethereum.org/EIPS/eip-5202) and supported by Vyper compilers via `-f blueprint_bytecode` CLI option.

- [ ] [VERIFIER, ETH_BYTECODE_DB] Add support for blueprint contracts verification
- [ ] [ETH_BYTECODE_DB] Add support for blueprint contracts lookup
- [ ] [VERIFIER_ALLIANCE] Research how it can be added into verifier alliance database

rimrakhimov:
Closes #843


libs/verification-common/src/blueprint_contracts.rs
```diff
@@ -0,0 +1,101 @@
...
+
+// ERC5202_PREFIX = b"\xFE\x71\x00"  # default prefix from ERC-5202
+// def blueprint_bytecode(self) -> bytes:
+//     blueprint_bytecode = ERC5202_PREFIX + self.bytecode
+//
+//     # the length of the deployed code in bytes
+//     len_bytes = len(blueprint_bytecode).to_bytes(2, "big")
+//     deploy_bytecode = b"\x61" + len_bytes + b"\x3d\x81\x60\x0a\x3d\x39\xf3"
+//
+//     return deploy_bytecode + blueprint_bytecode
```
sevenzing, line 11: needed?

libs/verification-common/src/blueprint_contracts.rs
```diff
@@ -0,0 +1,101 @@
...
+
+pub fn from_runtime_code(code: Bytes) -> Option<Bytes> {
+    let prefix = [0xfe, 0x71, 0x00];
```
sevenzing, line 14: are you sure you need 0x00 at the end?

from eip: https://eips.ethereum.org/EIPS/eip-5202#reference-implementation

image

libs/verification-common/src/blueprint_contracts.rs
```diff
@@ -0,0 +1,101 @@
...
+        .then(|| code.slice(prefix.len()..))
+}
+
+pub fn from_creation_code(code: Bytes) -> Option<Bytes> {
+    if code.len() < 10 {
+        return None;
+    }
+
+    let deploy_bytecode_prefix = [
+        0x61, code[1], code[2], 0x3d, 0x81, 0x60, 0x0a, 0x3d, 0x39, 0xf3,
```
sevenzing, line 25: so code[1] + code[2] is length of bytecode, right? could you please add comment describing that

libs/verification-common/src/blueprint_contracts.rs
```diff
@@ -0,0 +1,101 @@
...
+        let code = from_hex("fe71003461002457602061009a5f395f5160015560015f5561005f61002860003961005f6000f35b5f80fd5f3560e01c60026001821660011b61005b01601e395f51565b63158ef93e81186100535734610057575f5460405260206040f3610053565b633fa4f245811861005357346100575760015460405260206040f35b5f5ffd5b5f80fd0018003784185f810400a16576797065728300030a0013");
+        let expected_parsed_code = from_hex("3461002457602061009a5f395f5160015560015f5561005f61002860003961005f6000f35b5f80fd5f3560e01c60026001821660011b61005b01601e395f51565b63158ef93e81186100535734610057575f5460405260206040f3610053565b633fa4f245811861005357346100575760015460405260206040f35b5f5ffd5b5f80fd0018003784185f810400a16576797065728300030a0013");
+        assert_eq!(
+            Some(expected_parsed_code),
+            from_runtime_code(code),
+            "length encoded as 1 byte"
+        );
+
+        // A bytecode with length encoded as more than 1 byte
+        // https://etherscan.io/address/0x79D584d2D49eC8CE8Ea379d69364b700bd35874D#code
```
sevenzing, line 59: https://eth.blockscout.com/address/0x79D584d2D49eC8CE8Ea379d69364b700bd35874D?tab=contract

😄

rimrakhimov: It is not and cannot be verified until that PR is merged :(

libs/verification-common/src/blueprint_contracts.rs
rimrakhimov: Took it from vyper implementation for now, which does not have any data supported and, thus, always have the third byte as 0x00

Probably you are right and it make sense to implement it more generically according to ERC standard

eth-bytecode-db/eth-bytecode-db-proto/proto/v2/eth-bytecode-db.proto
```diff
@@ -101,6 +101,9 @@ message Source {
 
   /// Info about the runtime code (sourcemaps, linkreferences, immutables) encoded as a json.
   optional string deployed_bytecode_artifacts = 12;
+
+  /// If true, the verified contract was a ERC-5202 compatible blueprint contract
+  bool is_blueprint = 13;
```
k1rill-fedoseev, line 106: If we gonna have multiple similar customizations, maybe it's better to introduce a single field with such "tags". Can be a string/enum or a list of strings/enums. Wdyt?

rimrakhimov: I like the idea of having a list of enums in case we have similar customizations in the future. However, I'm not sure it is a good thing to add right now. For now I see no more tags which could potentially be added in the future, and it may be that the schema with list of strings/enums will not fit when such tags appear. 

E.g., I can imagine the situation where the contract may be tagged as a proxy with some additional info that should be returned along with the tag

So, imo it may be better to implement the simplest approach until an actual use case appear

libs/verification-common/src/blueprint_contracts.rs
rimrakhimov: Fixed

sevenzing (approved):

PR #855, opened 2024-05-01
[LIBS] env-collector

sevenzing:


libs/blockscout-service-launcher/src/env_collector.rs
```diff
@@ -0,0 +1,505 @@
...
+fn from_key_to_json_path(key: &str, service_prefix: &str) -> String {
+    key.trim_start_matches(&format!("{service_prefix}__"))
+        .to_lowercase()
+        .replace("__", ".")
+        .to_string()
+}
+
+fn serialize_env_vars_to_md_table(vars: Envs) -> String {
+    let mut result = r#"
+| Variable | Is required | Example value | Comment |
```
k1rill-fedoseev, line 312: Do you think it make sense to stick to the same format as in Blockscout docs - https://github.com/blockscout/docs/blob/master/for-developers/information-and-settings/env-variables.md?
```suggestion
| Variable | Required | Description | Default value |
```

sevenzing: sure, will fix it

PR #873, opened 2024-05-12
[ETH_BYTECODE_DB] fix: look for blueprint contracts

rimrakhimov:
### Motivation

Blueprint contracts with constructor arguments in the abi could not be found using eth-bytecode-db because the service required those arguments to exist in the provided contract creation code. But for blueprint contracts those arguments are empty.

### DIFF
--- a/eth-bytecode-db/eth-bytecode-db/src/search/types.rs
+++ b/eth-bytecode-db/eth-bytecode-db/src/search/types.rs
@@ -1,8 +1,48 @@
 use bytes::Bytes;
-use entity::sea_orm_active_enums::BytecodeType;
+use entity::sea_orm_active_enums;
+use std::fmt::{Display, Formatter};
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct BytecodeRemote {
     pub bytecode_type: BytecodeType,
     pub data: Bytes,
+}
+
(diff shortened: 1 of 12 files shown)


eth-bytecode-db/eth-bytecode-db/src/search/types.rs
```diff
@@ -1,8 +1,48 @@
...
+        match value {
+            BytecodeType::CreationCode | BytecodeType::CreationCodeWithoutConstructor => {
+                sea_orm_active_enums::BytecodeType::CreationInput
+            }
+            BytecodeType::RuntimeCode => sea_orm_active_enums::BytecodeType::DeployedBytecode,
+        }
+    }
+}
+
+impl Display for BytecodeType {
```
sevenzing, line 38: nit: in scoutcloud I decided to use https://docs.rs/serde_plain/latest/serde_plain/macro.derive_display_from_serialize.html
maybe you also want to use it :)

sevenzing (approved):

eth-bytecode-db/eth-bytecode-db/src/search/types.rs
rimrakhimov: Thanks! I will make use of it if such one more such enum arises or this one is changed frequently. But for now, I prefer leave it as it is, so not to add one more dependency just to save a couple of lines already written

PR #885, opened 2024-05-18
ZNSConnect integration

AayushCodes:


blockscout-ens/graph-node/subgraphs/zns-subgraph/src/zns-registry.ts
```diff
@@ -0,0 +1,102 @@
...
+  Domain,
+  Account,
+  Transfer,
+  Registration,
+  ExpiryExtended,
+} from "../generated/schema";
+import { crypto, Bytes } from "@graphprotocol/graph-ts";
+
+export function handleMintedDomain(event: MintedDomainEvent): void {
+  let domain = new Domain(event.params.tokenId.toString());
```
sevenzing, line 17: ID of domain should be in ERC-137 format: https://github.com/ethereum/ercs/blob/master/ERCS/erc-137.md

but tokenId is just autoincremented integer

blockscout-ens/graph-node/subgraphs/zns-subgraph/src/zns-registry.ts
```diff
@@ -0,0 +1,102 @@
...
+  domain.name = event.params.domainName;
+  domain.createdAt = event.block.timestamp;
+  domain.expiryDate = event.params.expiry;
+  domain.labelName = event.params.domainName;
```
sevenzing, line 21: `label` should be different from `name` 💯 

`name` -- `vitalik.eth`
`label` -- `vitalik`

https://docs.ens.domains/terminology#label

blockscout-ens/graph-node/subgraphs/zns-subgraph/src/zns-registry.ts
```diff
@@ -0,0 +1,102 @@
...
+    expiryExtended = new ExpiryExtended(expiryExtendedId);
+    expiryExtended.domain = domain.id;
+    expiryExtended.blockNumber = event.block.number.toI32();
+    expiryExtended.transactionID = event.transaction.hash;
+    expiryExtended.expiryDate = event.params.expiry;
+    expiryExtended.save();
+  }
+}
+
+export function handleTransfer(event: TransferEvent): void {
```
sevenzing, line 83: Am I correct that in your protocol `resolvedAddress` is the same as `owner`

if yes, this function has to update `resolvedAddress` field also

blockscout-ens/graph-node/subgraphs/zns-subgraph/src/zns-registry.ts
AayushCodes: Ok so we can do that, rather than the token ID, the id for each domain can be the namehash of the domain. Would you like it to be the namehash of "zns.poly" or just "zns"?

blockscout-ens/graph-node/subgraphs/zns-subgraph/src/zns-registry.ts
AayushCodes: Ok we'll make sure that is done

blockscout-ens/graph-node/subgraphs/zns-subgraph/src/zns-registry.ts
AayushCodes: Got it

blockscout-ens/graph-node/subgraphs/zns-subgraph/src/zns-registry.ts
sevenzing: namehash should be from the whole domain. Note that it's not just hash from string `"zns.poly"`. the algorithm described here: https://github.com/ethereum/ercs/blob/master/ERCS/erc-137.md#namehash-algorithm

for example subgraph for rootstock uses this function: https://github.com/blockscout/blockscout-rs/blob/ea31f1c5e0f88a586252fff9aa6fd29c6252fd80/blockscout-ens/graph-node/subgraphs/rns-subgraph/src/utils.ts#L104-L119

blockscout-ens/graph-node/subgraphs/zns-subgraph/src/zns-registry.ts
```diff
@@ -81,9 +70,15 @@ export function handleRenewedDomain(event: RenewedDomainEvent): void {
...
 
 export function handleTransfer(event: TransferEvent): void {
-  let domain = Domain.load(event.params.tokenId.toHex());
+  let domain = Domain.load(hashByName(event.params.domainName).toHex());
   if (!domain) {
-    return;
+    domain = new Domain(hashByName(event.params.domainName).toHex());
+    domain.name = event.params.domainName;
+    domain.createdAt = event.block.timestamp;
+    domain.labelName = event.params.domainName;
```
sevenzing, line 78: `event.params.domainName.split(".")[0]`

sevenzing:
merged with https://github.com/blockscout/blockscout-rs/pull/897

PR #889, opened 2024-05-20
[TEMPLATE] Add a prefix into Cargo.toml template file

rimrakhimov:
Updates the template directory not to have `Cargo.toml` file (and use `_Cargo.toml` instead).

### Motivation
When the crate is specified as dependency via git url (https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories)
> Cargo fetches the git repository at that location and traverses the file tree to find Cargo.toml file for the requested crate anywhere inside the git repository. 

As a result, when blockscout-rs git repository is specified as a dependency, cargo tries to process `service-template/Cargo.toml` as a usual Cargo.toml file and fails due to unexpected symbols (%{}) inside. That results in compilation error messages when compiled via CLI, and build errors in RustRover IDE.

When prefixed, cargo does not find any `Cargo.toml` file inside service-template directory and does not try to parse it as a usual crate


service-template/README.md
```diff
@@ -21,3 +21,5 @@ Template for blockscout Rust service
 🔧   Initializing a fresh Git repository
 ✨   Done! New project created /Users/levlymarenko/poa/blockscout-rs/test-project
 ```
+
+3. Go into created service directories and recursively rename all `_Cargo.toml` to `Cargo.toml`
```
sevenzing, line 25: just use post hook of `cargo-generate` to automate this:

https://cargo-generate.github.io/cargo-generate/templates/scripting.hook-types.html#post

Issue #845, opened 2024-04-24
stats: optimism `new_contracts` bug


bragov4ik:
Issue https://github.com/blockscout/blockscout-rs/issues/841 follows from inaccurate data in chart `new_contracts`.

## Possible reasons and 
Apparently, blockscout reindexes old transactions, which is not considered for chart updating.

## Fix directions
Changing logic of the update should help. In particular, utilizing field `updated_at` in table `transactions` with the help of corresponding index. I.e. query txn's added after last sync and add the result to stats DB.
Note: need to consider updates that do not result in changing `created_contract_address_hash` field. How - not sure rn.

bragov4ik:
related, but seems more general:
https://github.com/blockscout/blockscout-rs/issues/514

PR #891, opened 2024-05-24
Update stats architecture

bragov4ik:
allows more complex relations between charts


stats/stats-proto/proto/stats.proto
```diff
@@ -6,7 +6,7 @@ option go_package = "github.com/blockscout/blockscout-rs/stats";
 
 service StatsService {
   rpc GetCounters(GetCountersRequest) returns (Counters);
-  rpc GetLineCharts(GetLineChartsRequest) returns (LineCharts);
+  rpc GetLineCharts(GetLineChartsRequest) returns (LineCharts); // todo: rename
```
sevenzing, line 9: this line is added to swagger, are you sure?

stats/stats-server/src/config/read/merge.rs
```diff
@@ -0,0 +1,369 @@
...
+            }
+        }
+    };
+}
+
+impl_get_key!(CounterInfo<C>);
+impl_get_key!(LineChartInfo<C>);
+impl_get_key!(LineChartCategory<C>);
+
+/// `overwrite_f` should update 1st parameter with values from the 2nd
```
sevenzing, line 59: `update_t` or `overwrite_f` ?

stats/stats/src/update_group.rs
```diff
@@ -0,0 +1,523 @@
...
+/// construct_update_group!(ExampleUpdateGroup {
+///     name: "exampleGroup",
+///     charts: [DummyChartSource],
+/// });
+/// ```
+///
+#[macro_export]
+macro_rules! construct_update_group {
+    ($group_name:ident {
+        name: $name:literal,
```
sevenzing, line 212: could be obtained from `$group_name` using https://github.com/dtolnay/paste?tab=readme-ov-file#case-conversion

sevenzing:
please add `[STATS]` to name of the PR

stats/stats/src/update_group.rs
bragov4ik: decided to not change the case for easier comprehension of logs

bragov4ik: still removed the name field

stats/stats-server/src/config/types.rs
```diff
@@ -0,0 +1,115 @@
...
+                            id: c,
+                            settings: e.settings.clone(),
+                        }
+                        .into()
+                    })
+                    .clone()
+            })
+            .collect();
+        let Some(charts) = charts else {
+            tracing::error!("as");
```
sevenzing, line 94: as?

stats/stats-server/src/update_groups.rs
```diff
@@ -0,0 +1,87 @@
...
+});
+
+construct_update_group!(NewContractsGroup {
+    charts: [
+        NewContracts,
+        ContractsGrowth,
+        // currently can be in a separate group but after #845
+        // it's expected to join the group. placing it here to avoid
+        // updating config after the fix (todo: remove (#845))
+        TotalContracts,
```
sevenzing, line 59: I didn't get it. 

Total Contracts should be in separated groups, right?
`stats` cannot recalculate charts after indexing of internal transactions, so we wrote separate SQL for total_contract. But now it looks like you're merging it again

stats/stats-server/src/settings.rs
```diff
@@ -51,6 +53,8 @@ impl Default for Settings {
             concurrent_start_updates: 3,
             limits: Default::default(),
             charts_config: PathBuf::from_str("config/charts.json").unwrap(),
+            layout_config: PathBuf::from_str("config/layout.json").unwrap(),
```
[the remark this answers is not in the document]
bragov4ik: yep, pushed the version before config migration

stats/stats-server/src/update_groups.rs
bragov4ik: We still have the separate SQL, `TotalContracts` being in this group only means that its update will be triggered simultaneously to other charts in the group.

Why I added it:
Let's say `TotalContracts` is in it's own group. Now there's an additional group in the config. Once (if) #845 is fixed, `TotalContracts` will have to be moved to `NewContractsGroup` anyway. However, this means that the configs have to be updated as well.
Therefore, by adding it to the group in advance we won't need to make additional config changes.

bragov4ik: Feel free to reopen the thread; I'll close it for now

stats/stats-server/src/config/types.rs
bragov4ik: no double s, indeed 😢 (forgot, thx for catching)

stats/stats/src/charts/lines/new_accounts.rs
```diff
@@ -1,114 +1,157 @@
...
+        // filter
+        if let Some(range) = range {
+            Statement::from_sql_and_values(
+                DbBackend::Postgres,
+                r#"
+                        SELECT
+                            first_tx.date as date,
+                            count(*) as value
+                        FROM (
+                            SELECT DISTINCT ON (t.from_address_hash)
```
sevenzing, line 43: Cannot we use `sql_with_range_filter_opt` here?

stats/stats/src/charts/lines/new_accounts.rs
```diff
@@ -1,114 +1,157 @@
...
+        PassVecStep,
+        // see `NewAccountsRemote` docs
+        BatchMax,
+        Properties,
+    >,
+    DefaultQueryVec<Properties>,
+    Properties,
+>;
+
+pub type NewAccountsInt = MapParseTo<NewAccounts, DateValueInt>;
```
sevenzing, line 143: It looks strange that you query Int, then cast it to String and then cast it to String again, is it intended?

stats/stats-server/src/read_service.rs
```diff
@@ -52,88 +57,104 @@ fn map_read_error(err: ReadError) -> Status {
...
         &self,
-        _request: Request<GetLineChartsRequest>,
-    ) -> Result<Response<LineCharts>, Status> {
-        Ok(Response::new(self.charts.config.lines.clone().into()))
+        _request: Request<proto_v1::GetLineChartsRequest>,
+    ) -> Result<Response<proto_v1::LineCharts>, Status> {
+        let layout = self.charts.lines_layout.clone();
+        let settings = self.charts.charts_info.clone();
+        let sections = add_settings_to_layout(layout, settings).ok_or(
+            Status::internal(""), // todo: request id??
```
sevenzing, line 155: 💯 we need to say something to user and log the error

stats/stats/src/charts/lines/new_accounts.rs
bragov4ik: Because only the upper bound of the range is used in the statement. We don't want to ignore previous transactions to get correct values for "new accounts".

For example, if address A appeared both before `range.start()` and inside the range, we don't want it to be counted within our range (it was `created` only once - on the first appearance).

stats/stats/src/charts/lines/new_accounts.rs
bragov4ik: Apparently, it is queried as int and not string. I copied the logic from original implementation. Changing it to string (in query) results in:
```
thread 'charts::lines::new_accounts::tests::update_new_accounts' panicked at stats/src/tests/simple_test.rs:41:38:
called `Result::unwrap()` on an `Err` value: BlockscoutDB(Query(SqlxError(ColumnDecode { index: "\"value\"", source: "mismatched types; Rust type `core::option::Option<alloc::string::String>` (as SQL type `TEXT`) is not compatible with SQL type `INT8`" })))
```

PR #915, opened 2024-06-11
[SCOUTCLOUD] Stats, check health, more info in API

sevenzing:
+ add `promo` feature
+ add `unhealthy` deployment status
+ add `supervisor` job that handles health of instances
+ add register api-key endpoint
+ update `k8s-dev` default sub domain to `cloud`
+ add new env variables ([env docs](https://www.notion.so/blockscout/c2eb5ebd57e548c1b7f677df76703728?v=78feefe69e6c42b59d21879d96722c41&pvs=4))
+ add custom `stats_enabled` env that checks indexing status of blockscout
+ add `context` for generating config that allows some kind of "dependency" between variables, for example `stats_enabled` depends on `instance_url`
+ update swagger: add field_behavior for required fields


scoutcloud/scoutcloud/src/logic/users/handlers/crud.rs
```diff
@@ -1,11 +1,56 @@
...
+            note: Set(Some(format!("From promo code: {}", promo.code))),
+            register_promo_id: Set(Some(promo.id)),
+            ..Default::default()
+        }
+        .insert(&tx)
+        .await?;
+    }
+    let token = issue_auth_token(&tx, user.id, "initial token").await?;
+    tx.commit().await?;
+    let user_token = UserToken::try_from_token_value(db, &token.token_value).await?;
```
rimrakhimov, line 45: As I understand it, that should always succeed. But in case if it fails, now we return `Authenticated` to the users, while actually that is internal error

scoutcloud/scoutcloud/src/logic/users/handlers/crud.rs
```diff
@@ -21,6 +66,37 @@ pub async fn get_profile(
...
+        created_at: user_token.token.created_at.to_string(),
+    })
+}
+
+pub async fn get_auth_tokens<C: ConnectionTrait>(
+    db: &C,
+    user_token: &UserToken,
+) -> Result<Vec<proto::AuthTokenInternal>, AuthError> {
+    let tokens = scoutcloud_entity::auth_tokens::Entity::find()
+        .filter(scoutcloud_entity::auth_tokens::Column::UserId.eq(user_token.user.id))
```
rimrakhimov, line 87: Missed filter of deleted tokens?

PR #958, opened 2024-06-28
fix(stats): incorrect last point of previous batch step (force update)

bragov4ik:
* fix(stats): Fixed last point of previous batch step being retrieved incorrectly when forcing reupdate of an existing data. Also added regression on this bug to prevent it in the future.
* fix(stats): In the meantime fixed a small overlapping between batch steps (1 ns due to using `RangeInclusive`): end of batch step range is now not included (`Range` instead of `RangeInclusive`). Relevant code is changed accordingly.

### DIFF
--- a/stats/stats/src/data_source/kinds/local_db/parameters/update/batching/mod.rs
+++ b/stats/stats/src/data_source/kinds/local_db/parameters/update/batching/mod.rs
@@ -7,12 +7,14 @@ use std::{marker::PhantomData, ops::RangeInclusive, time::Instant};
 
 use blockscout_metrics_tools::AggregateTimer;
 use chrono::{DateTime, Days, Utc};
-use parameter_traits::{BatchSizeUpperBound, BatchStepBehaviour};
+use parameter_traits::BatchStepBehaviour;
 use sea_orm::TransactionTrait;
 
 use crate::{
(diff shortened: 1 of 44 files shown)


stats/stats/src/data_source/kinds/local_db/parameters/update/batching/mod.rs
```diff
@@ -203,12 +239,196 @@ mod tests {
...
+                day_start(d("2015-07-20"))..day_start(d("2015-07-21")),
+                day_start(d("2015-07-21"))..day_start(d("2015-07-22"))
+            ],
+            generate_date_time_ranges(
+                day_start(d("2015-07-20")),
+                day_start(d("2015-07-22")),
+                chrono::Duration::days(1)
+            )
+        )
+    }
```
sevenzing, line 267: you only test 1day batch.
I suggest also test 30 days batch and check if month change works correctly

bragov4ik: but tests for 30 days is right above these?...

sevenzing: right

sevenzing (approved):

Issue #964, opened 2024-07-01
stats: repeated updates within group


bragov4ik:
seems that the metadata check for update time doesn't work

- [x] create unit test (+for future regression testing) and verify the issue
- [x] fix the bug if test fails

bragov4ik:
After adding logs in #967 the following is observed
`{"timestamp":"2024-07-02T15:39:10.059639Z","level":"DEBUG","message":"Performing an update","last_updated_at":"2024-07-02T15:36:37.476549Z","update_timestamp":"2024-07-02T15:36:37.476549314Z","target":"stats::data_source::kinds::local_db","span":{"name":"update_recursively"}}`

So,
```
"last_updated_at":"2024-07-02T15:36:37.476549Z"
```
and 
```
"update_timestamp":"2024-07-02T15:36:37.476549314Z"
```
🥇

bragov4ik:
now it makes sense it didn't repro with manually-set timestamp

bragov4ik:
after the fix the logs indicating update skip are observed and the updates are performed as expected
<img width="1226" alt="image" src="https://github.com/blockscout/blockscout-rs/assets/8144358/b94df590-669b-4b6e-a215-2c1065741131">

PR #970, opened 2024-07-02
fix(stats): updates skip did not consider postgres timestamp precision

bragov4ik:
(regression) test and fix.
Solves #964, see details there

### DIFF
--- a/stats/stats/src/data_source/kinds/local_db/mod.rs
+++ b/stats/stats/src/data_source/kinds/local_db/mod.rs
@@ -461,6 +461,36 @@ mod tests {
                 .update_charts_with_mutexes(parameters, &enabled)
                 .await
                 .unwrap();
+
+            UpdateSingleTriggerAsserter::reset_triggers().await;
+
+            // also test if there is any rounding when inserting metadata
+            let next_next_time = next_time.checked_add_days(Days::new(1)).unwrap();
+            let next_next_time = next_next_time + TimeDelta::nanoseconds(500);
+            let parameters = UpdateParameters {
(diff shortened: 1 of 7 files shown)


stats/stats/src/data_source/kinds/local_db/mod.rs
```diff
@@ -150,7 +150,9 @@ where
     ) -> Result<(), UpdateError> {
         let metadata = get_chart_metadata(cx.db, ChartProps::NAME).await?;
         if let Some(last_updated_at) = metadata.last_updated_at {
-            if cx.time == last_updated_at {
+            // DB stores timestamps with microsecond precision
+            let update_time = cx.time.trunc_subsecs(6);
+            if update_time == last_updated_at {
```
sevenzing, line 155: why not just `if update_time >= last_update_at` ? you wont need to truncate anything

bragov4ik: `>=` is just incorrect, because we don't want to skip all updates in the future
`<=` won't work because after truncating (=`last_updated_at`), we have `cx.time >= last_updated_at`

sevenzing: okay, when let's truncate until seconds and compare only seconds value. if time difference is less than a second, I don't think we might want to perform update

bragov4ik: decided to extract into a separate function for better readability

sevenzing (approved):

PR #985, opened 2024-07-17
feat(libs): add support for any origin in case of cors wildcard

sevenzing:


libs/blockscout-service-launcher/src/launcher/settings.rs
```diff
@@ -80,15 +81,19 @@ impl Default for CorsSettings {
             enabled: false,
             allowed_origin: "".to_string(),
             allowed_methods: "PUT, GET, POST, OPTIONS, DELETE, PATCH".to_string(),
-            allowed_credentials: true,
+            allowed_credentials: false,
```
rimrakhimov, line 84: Why it should be changed?

rimrakhimov (approved):

rimrakhimov (approved):

Issue #937, opened 2024-06-21
BENS: enchaucments


sevenzing:

sevenzing:
Here is list of possible domain names related to this feature for testing


offchain name | url to external resource
-- | --
`lucemans.uni.eth` | https://enstate.rs/n/lucemans.uni.eth
`levvv.xyz` | https://enstate.rs/n/levvv.xyz
`1.offchaindemo.eth` | https://enstate.rs/n/1.offchaindemo.eth
`1.offchainexample.eth` | https://enstate.rs/n/1.offchainexample.eth
`<any>.offchainexample.eth` |  
`itslev.cb.id` | https://enstate.rs/n/itslev.cb.id


onchain dns name | url to external resource
-- | --
`llev.me` | https://enstate.rs/n/llev.me

PR #989, opened 2024-07-23
feat(bens): offchain and wildcard resolution

sevenzing:
related to https://github.com/blockscout/blockscout-rs/issues/937


blockscout-ens/bens-logic/src/subgraph/ens/wildcard.rs
```diff
@@ -0,0 +1,186 @@
...
+    let name_options = from_user
+        .inner
+        .iter_parents_with_self()
+        .map(|name| DomainNameOnProtocol::new(name, from_user.deployed_protocol))
+        .collect::<Vec<_>>();
+    // try to find resolver in db
+    let maybe_domain_with_resolver = any_resolver_in_db(db, name_options.clone()).await?;
+    if let Some((resolver, found_domain)) = maybe_domain_with_resolver {
+        Ok(Some((resolver.resolver_address, Some(found_domain))))
+    } else if protocol.info.registry_contract.is_some() || protocol.info.network_id == 1 {
```
sevenzing, line 105: remove netwrok_id == 1

blockscout-ens/bens-logic/src/subgraph/ens/wildcard.rs
```diff
@@ -0,0 +1,186 @@
...
+                .and_then(|domain| {
+                    let maybe_resolver = domain
+                        .resolver
+                        .as_ref()
+                        .and_then(|r| subgraph::ResolverInSubgraph::from_str(r).ok());
+                    maybe_resolver.map(|r| (r, domain.clone()))
+                })
+        })
+        .next();
+    Ok(result)
```
rimrakhimov, line 146: Nit: using explicit for cycle may be more readable in that case (imo, it better conveys the intention to return the first non-empty result)

```suggestion
    for name in names {
        let maybe_domain = found_parent_domains
            .iter()
            .find(|d| d.name.as_ref() == Some(&name.inner.name));

        if let Some(domain) = maybe_domain {
            let maybe_resolver = domain
                .resolver
                .as_ref()
                .and_then(|r| ResolverInSubgraph::from_str(r).ok());

            if let Some(resolver) = maybe_resolver {
                return Ok(Some((resolver, domain.clone())));
            }
        }
    }

    Ok(None)
```

blockscout-ens/bens-logic/src/subgraph/ens/wildcard.rs
```diff
@@ -0,0 +1,186 @@
...
+        // try to find resolver on chain.
+        // if custom registry is set, we can try to find resolver in registry
+        if let Some(resolver_address) = any_resolver_rpc(name_options).await {
+            Ok(Some((resolver_address, None)))
+        } else {
+            Ok(None)
+        }
+    } else {
+        Ok(None)
+    }
```
rimrakhimov, line 115: Nit: maybe avoid nesting of if else statements

```suggestion
    if let Some((resolver, found_domain)) = maybe_domain_with_resolver {
        return Ok(Some((resolver.resolver_address, Some(found_domain))))
    } else if protocol.info.registry_contract.is_some() || protocol.info.network_id == 1 {
        // try to find resolver on chain.
        // if custom registry is set, we can try to find resolver in registry
        if let Some(resolver_address) = any_resolver_rpc(name_options).await {
            return Ok(Some((resolver_address, None)))
        }
    }
    Ok(None)
```

rimrakhimov (approved):

PR #999, opened 2024-08-01
feat(da_indexer): add l2BatchMetadata endpoint

AllFi:
This PR introduces a new `l2BatchMetadata` endpoint, which can be used to fetch information about an L2 batch by blob height, namespace, and commitment, provided the blob contains an L2 batch. To enable this functionality, the service requires a configuration file that stores the mapping of namespaces to Blockscout instances. The service is intended to act as a proxy between Celenium and specific Blockscout instances.

The service can be run independently of the indexer. Since caching is not yet implemented, the service does not require a database.


da-indexer/da-indexer-logic/src/celestia/l2_router/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+    ) -> Result<Option<L2BatchMetadata>> {
+        let namespace = Bytes::from(namespace.to_vec()).to_string();
+        let config = {
+            let config = self.routes.get(&namespace);
+            if config.is_none() {
+                tracing::debug!("unknown namespace: {}", &namespace);
+                return Ok(None);
+            }
+            config.unwrap()
+        };
```
rimrakhimov, line 52: Nit: maybe a little clearer and allows to get rid of unwrap():
```suggestion
        let config = match self.routes.get(&namespace) {
            Some(config) => config,
            None => {
                tracing::debug!("unknown namespace: {}", &namespace);
                return Ok(None);
            }
        };
```

da-indexer/da-indexer-logic/src/celestia/l2_router/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+        Ok(router)
+    }
+
+    pub async fn get_l2_batch_metadata(
+        &self,
+        height: u64,
+        namespace: &[u8],
+        commitment: &[u8],
+    ) -> Result<Option<L2BatchMetadata>> {
+        let namespace = Bytes::from(namespace.to_vec()).to_string();
```
rimrakhimov, line 44: Nit: in blockscout_display_bytes v1.1.0 a `ToHex` trait was added which allows not to wrap the data into Bytes explicitly
```suggestion
        let namespace = blockscout_display_bytes::ToHex::to_hex(namespace);
```

da-indexer/da-indexer-logic/src/celestia/l2_router/arbitrum.rs
```diff
@@ -0,0 +1,66 @@
...
+    let query = format!(
+        "{}/api/v2/arbitrum/batches/da/celestia/{}/{}",
+        config.l2_api_url, height, commitment,
+    );
+
+    let response = Client::new()
+        .get(&query)
+        .timeout(config.request_timeout)
+        .send()
+        .await?;
```
rimrakhimov, line 38: Usually we use `reqwest_retry` (https://docs.rs/reqwest-retry/latest/reqwest_retry/) client implementation when sending requests to 3-rd party services to handle some transient errors with retry pattern. Maybe you want to consider that here

E.g., https://github.com/blockscout/blockscout-rs/blob/main/sig-provider/sig-provider/src/sources/mod.rs#L58

rimrakhimov (approved):

da-indexer/da-indexer-logic/src/celestia/l2_router/arbitrum.rs
AllFi: Yes, it makes sense, thank you! Fixed in https://github.com/blockscout/blockscout-rs/pull/999/commits/c88d14a797fd6f05354095167cfe55b6814ca6ca

Issue #849, opened 2024-04-26
stats: aggregate result of line chart by days, weeks, months, years


sevenzing:
Add new input to stats page (like interval) called "resolution" to return results for a single day, a week, a month, or a year:

<img width="638" alt="image" src="https://github.com/blockscout/blockscout-rs/assets/41516657/fe74533a-60f7-44cb-ab54-da21b2efe976">

most likely it will change api, or at least `date` key in response

Refactor the service in preparation for feature:
- [x] Modify the service to allow more generic and complex relationships (e.g. between charts) #891

Implement resolutions:
- [x] `Week` data type
- [x] Generic data types that support abstract resolution
- [x] Interface inner resolutioned points <-> DB
- [x] Adapt data source implementors for various resolutions
- [x] Implement one weekly chart with passing test
- [x] Work through using chart info (in DB) for resolutioned charts
- [x] Add resolutions to configs
- [x] Update read service (+protobuf)
- [x] Implement weekly versions for all chart types
- [x] Cover with tests
- [x] Generalize the solution to years & months
- [x] Add new cases to tests
- [x] Smoke run (test) locally

bragov4ik:
Caveat: chart "ActiveAccounts". Adding intervals requires us to calculate/store unique users for each interval. It's quite non-trivial.
Also not really clear what to do with cumulative/increasing graphs. Show result at the start/middle/end of the interval, or even an average value.

bragov4ik:
Almost forgot: graphs with averages are non-trivial as well. To properly compute an average of some interval from averages of subintervals we need to retrieve # of entries for each subinterval.

I.e. we have 100 entries on monday (avg=0.5) and 1 entry on friday (avg=1).
Avg for the week should be ~0.5 instead of 0.75

bragov4ik:
for now will implement it for easier cases, let's decide on "ActiveAccounts" later..

bragov4ik:
## API design (report-ish)
### Request
Add a parameter to query string (enum => string with some set of available values).
Variants:
- day
- week
- month
- year

Pros:

- readable
- clear list of possible inputs
- flexible, string type doesn’t restrict possible resolutions

Cons:

- might become ugly if some non-regular intervals are added later (i.e. for 3 months `..&resolution=3months` looks weird) - minor, seems ok

### Response
#### Variant A
Response API remains the same:
```json

{
    "chart": [
        {
            "date": "2024-07-01",
            "value": "35607"
        },
        {
[...]

PR #1002, opened 2024-08-05
feat(stats): resolutions for line charts

bragov4ik:
#849 
List of changes (not guaranteed to be full, but main changes should be there)

Features/semantic changes:
- add support for resolutions
  - change API
  - db schema
  - chart operation (updating, querying, id'ing)
  - configuration (files + envs)
- identify charts by name and resolution combination instead of name only

Code (quality) updates:
- simpler runtime init (in terms of code complexity, functionality is the same)
- improved logging (for easier debugging/searching)
- some code/responsibility deduplication in db read logic
- more powerful mock line charts for better testing
- remove some not used methods


stats/stats-proto/build.rs
```diff
@@ -22,7 +22,8 @@ fn compile(
             "#[serde(default)]"
         )
         .field_attribute(".blockscout.stats.v1.Point.is_approximate", "#[serde(skip_serializing_if = \"std::ops::Not::not\")]")
-        .field_attribute(".blockscout.stats.v1.Point.is_approximate", "#[serde(default)]");
+        .field_attribute(".blockscout.stats.v1.Point.is_approximate", "#[serde(default)]")
+        .field_attribute(".blockscout.stats.v1.GetLineChartRequest.resolution", "#[serde(default = \"default_resolution_i32\")]");
```
sevenzing, line 26: can you just leave `#[serde(default)]` without writing default function? i32 always defaults to 0 and unspecified option is always the first option, so I think no need to create `default_resolution_i32`

stats/stats-server/tests/lines.rs
```diff
@@ -69,16 +79,31 @@ async fn test_lines_ok() {
...
+            .unwrap_or_else(|| panic!("must return chart info for {}", &line_name));
+        assert!(
+            line_resolutions.contains(&ResolutionKind::Day.into()),
+            "At least day resolution must be enabled for enabled chart"
+        );
+        for resolution in line_resolutions {
+            let chart: serde_json::Value = send_get_request(
+                &base,
+                &format!("/api/v1/lines/{line_name}?resolution={resolution}"),
+            )
```
sevenzing, line 93: add request with empty query to show that `resolution` is optional

stats/stats/src/update_groups.rs
```diff
@@ -5,13 +5,7 @@ use crate::{
         TotalAddresses, TotalBlocks, TotalContracts, TotalNativeCoinHolders,
```
sevenzing, line 5: let's be concise and import all counters as `counters::*` then

stats/stats/src/charts/types/timespans/week.rs
```diff
@@ -0,0 +1,254 @@
...
+        Some(self.checked_first_day()?..=self.checked_last_day()?)
+    }
+
+    /// `None` - out of range (week of `NaiveDate::MIN`)
+    pub fn checked_first_day(&self) -> Option<NaiveDate> {
+        // Current interface of `NaiveWeek` does not provide any way
+        // of idiomatic overflow handling
+        // todo: change after https://github.com/chronotope/chrono/pull/1600
+        // is released
+        std::panic::catch_unwind(|| self.0.first_day()).ok()
```
sevenzing, line 138: i suggest remove panic logs, at least in production.

If there is no way to remove panic logs, lets write our own NaiveWeek

stats/stats/src/charts/chart.rs
```diff
@@ -104,22 +189,63 @@ pub trait ChartProperties: Sync + Named {
     }
 }
 
+#[macro_export]
+macro_rules! delegated_properties_with_resolutions {
```
sevenzing, line 193: nit: I didn't get why its called `delegated_properties_with_resolutions`. maybe `define_and_impl_resolutions`

bragov4ik: Because the new properties' implementations are delegated to the specified one, but the resolution is overwritten. 

`define_and_impl_resolutions` seems to suggest that we define some resolutions, which is not what happens here (?)

stats/stats/src/charts/types/timespans/week.rs
bragov4ik: fixed it with patch, problem with maintaining it should not appear, as these changes are expected to be added to the next release (if not, it will be discovered soon and I'll address this in some other way)

sevenzing (approved):
let's do it!

PR #1006, opened 2024-08-09
investigating moving out common proto files

bragov4ik:


bragov4ik:
(duplicate of https://github.com/blockscout/blockscout-rs/pull/893)

PR #1007, opened 2024-08-13
feat(swagger): Swagger endpoint definition in a separate `blockscout-swagger-endpoint` crate

bragov4ik:


libs/endpoints/swagger/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+}
+
+pub fn register_route(
+    service_config: &mut actix_web::web::ServiceConfig,
+    swagger_file_path: PathBuf,
+) {
+    let path = Arc::new(swagger_file_path);
+    let serve_swagger = move |req: HttpRequest| serve_swagger_from(path.clone(), req);
+    service_config.configure(|config| {
+        config.route("/api/v1/docs/swagger.yaml", get().to(serve_swagger));
```
sevenzing, line 17: I would add variable "route_path": str instead of hardcoding this path here

sevenzing (approved):

PR #1008, opened 2024-08-13
chore(health): extract health into separate crate

bragov4ik:
Extract common logic for setting up health endpoint compilation (+ proto file). This way the implementation details are hidden under the functions and can be reused for other crates.


stats/stats-proto/build.rs
```diff
@@ -16,33 +16,55 @@ fn compile(
...
+        .map(|path| {
+            path.to_str()
+                .expect("Non UTF-8 paths are not supported")
+                .to_string()
+        })
+        .collect()
+}
+
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let swagger_crate_folder = Path::new("../../libs/endpoints/health");
```
sevenzing, line 44: swagger or health ?

libs/endpoints/health/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+//!
+//! Now the types should be available in `grpc::health::v1` module, and swagger
+//! entry for the endpoint should appear
+
+use std::path::{Path, PathBuf};
+
+use prost_build::Config;
+
+pub fn add_to_compile_config(config: &mut Config) {
+    config.bytes([".grpc.health"]).type_attribute(
```
rimrakhimov, line 37: In my opinion it is the client code responsibility to decide should the generated code use bytes or not, so I suggest to remove it from here and only leave `.bytes(["."])` in the client code

libs/endpoints/health/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        ".grpc.health",
+        "#[actix_prost_macros::serde(rename_all=\"snake_case\")]",
```
rimrakhimov, line 39: Also, wdyt about leaving the case definition as the client responsibility as well?

stats/stats-proto/build.rs
rimrakhimov: In general, will this work in other than blockscou-rs repositories and will this build inside docker containers?

stats/stats-proto/build.rs
```diff
@@ -16,33 +16,55 @@ fn compile(
         .protoc_arg("--openapiv2_opt")
         .protoc_arg("grpc_api_configuration=proto/api_config_http.yaml,output_format=yaml,allow_merge=true,merge_file_name=stats,json_names_for_fields=false")
         .bytes(["."])
-        .type_attribute(".", "#[actix_prost_macros::serde(rename_all=\"snake_case\")]")
+        .type_attribute(".blockscout.stats", "#[actix_prost_macros::serde(rename_all=\"snake_case\")]")
         .field_attribute(
             ".blockscout.stats.v1.HealthCheckRequest.service",
             "#[serde(default)]"
         )
```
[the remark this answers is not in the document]
sevenzing: maybe just add `optional` to `service` field?

libs/endpoints/health/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    );
+}
+
+pub fn proto_files(path_to_swagger_crate: &Path) -> Vec<PathBuf> {
```
bragov4ik, line 43:
```suggestion
pub fn proto_files(path_to_health_crate: &Path) -> Vec<PathBuf> {
```

stats/stats-proto/build.rs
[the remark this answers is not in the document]
bragov4ik: need to add to proto then remove

bragov4ik:
need to fix troubles with versioning/path resolution

stats/stats-proto/build.rs
```diff
@@ -16,33 +16,55 @@ fn compile(
...
+
+    for folder in &proto_files_folders {
+        println!("cargo:rerun-if-changed={folder}/");
+    }
+
+    let mut protos = ["proto/stats.proto"].map(PathBuf::from).to_vec();
+    protos.extend(blockscout_endpoint_health::proto_files(
+        swagger_crate_folder,
+    ));
+    let protos = vec_path_buf_to_string(protos);
```
implausibleDeniability, line 60: I wonder if the main() does too many things. Should we may be isolate path setup in a separate function?

implausibleDeniability (commented):
Overall, good job 👍👍👍

PR #1014, opened 2024-08-15
fix(launcher): `RequestIdStorage` fixes

bragov4ik:
- Fix the layer overwriting fields of handled spans (due to setting `extensions` to some value which are prioritized over any fields)
- Add unit test to check it
- Clippy fixes

### DIFF
--- a/libs/blockscout-service-launcher/src/tracing/request_id_layer.rs
+++ b/libs/blockscout-service-launcher/src/tracing/request_id_layer.rs
@@ -186,13 +186,14 @@ mod tests {
     }
 
     fn parse_json(s: &str) -> serde_json::Value {
-        serde_json::from_str::<serde_json::Value>(&s).expect(&format!("failed to parse '{}'", s))
+        serde_json::from_str::<serde_json::Value>(s)
+            .unwrap_or_else(|_| panic!("failed to parse '{}'", s))
     }
 
     fn parse_captured_logs(logs: String) -> Vec<serde_json::Value> {
         logs.split('\n')
             .filter(|l| !l.is_empty())
-            .map(|l| parse_json(l))
+            .map(parse_json)
             .collect()
     }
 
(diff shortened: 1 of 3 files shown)


libs/blockscout-service-launcher/src/tracing/request_id_layer.rs
```diff
@@ -77,3 +87,190 @@ impl<S: Subscriber + for<'lookup> LookupSpan<'lookup>> Layer<S> for RequestIdSto
...
+            let span0_with_request_id = tracing::info_span!(
+                "span0_with_request_id",
+                request_id = %"02f09a3f-1624-3b1d-8409-44eff7708208"
+            );
+            let _e0 = span0_with_request_id.enter();
+            let span1 = tracing::info_span!("span1", x = 42);
+            let _e = span1.enter();
+        });
+        let actual = sanitize_timings(make_writer.get_string());
+        assert_eq!(
```
rimrakhimov, line 230: Did the problem exist when `with_span_list(false)` was activated? And shouldn't we test it when all spans are merged instead?

rimrakhimov (commented):
Could you please briefly describe how your solution fixes the problem?

libs/blockscout-service-launcher/src/tracing/request_id_layer.rs
bragov4ik: add case with `with_span_list(false)`

sevenzing (approved):

lok52 (approved):

rimrakhimov (approved):

libs/blockscout-service-launcher/src/tracing/request_id_layer.rs
bragov4ik: `with_span_list` doesn't really affect the test and `true` only provides more info to check (just for the record)

PR #1037, opened 2024-08-29
chore: improve `get_line_charts` for `ReadService`

yjhmelody:
Reduce some clone for the function.

### DIFF
--- a/stats/stats-server/src/read_service.rs
+++ b/stats/stats-server/src/read_service.rs
@@ -68,12 +68,12 @@ fn map_read_error(err: ReadError) -> Status {
 ///
 /// Returns `None` if info were not found for some chart.
 fn add_chart_info_to_layout(
-    layout: Vec<types::LineChartCategory>,
-    chart_info: BTreeMap<String, EnabledChartEntry>,
+    layout: &[types::LineChartCategory],
+    chart_info: &BTreeMap<String, EnabledChartEntry>,
 ) -> Vec<proto_v1::LineChartSection> {
     layout
-        .into_iter()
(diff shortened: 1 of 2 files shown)


stats/stats-server/src/read_service.rs
```diff
@@ -68,12 +68,12 @@ fn map_read_error(err: ReadError) -> Status {
 ///
 /// Returns `None` if info were not found for some chart.
 fn add_chart_info_to_layout(
-    layout: Vec<types::LineChartCategory>,
-    chart_info: BTreeMap<String, EnabledChartEntry>,
+    layout: &[types::LineChartCategory],
```
bragov4ik, line 71: I agree with `chart_info`, but I think it should be cheaper to perform 1 big clone instead of multiple small (equivalent) clones

bragov4ik (changes requested):

stats/stats-server/src/read_service.rs
yjhmelody: they are in heap，so i think its no difference，but only parts of fields will be cloned，so i think it will be cheaper

bragov4ik: Ok, probably there's no difference because of strings.

However, `intersect_info` effectively clones whole `LineChartCategory`, so I don't see a reason to use reference where move semantics fits ideally. I think it's a good practice to leave decision to clone or not for the caller.

bragov4ik: I.e.
- `id` - cloned directly
- `title` - cloned directly
- `charts_order` - `.iter()` then `clone` on each element

PR #1039, opened 2024-08-29
fix(stats): 'no data' for grafana chart errors count

bragov4ik:
not very descriptive, is it no data because of some metric collection fails or because everything is fine 🤔


bragov4ik:
just threw in some integration tests optimization, to no make too much PRs

PR #1041, opened 2024-08-29
chore(libs): update blockscout db entities to 6.8.0 state

bragov4ik:


bragov4ik:
this doesn't work because
- `accounts_tags` table doesn't have primary key (easily fixable by hand here)
- basically [this issue](https://github.com/SeaQL/sea-orm/issues/2342)

so, waiting for the latter to somehow resolve

bragov4ik:
>to specify tables to import
you mean to exclude tables that face the referenced issue?

libs/blockscout-db/justfile
```diff
@@ -27,8 +27,13 @@ generate-entities database_url="postgres://postgres:admin@localhost:5432/blocksc
...
     cargo fmt --all -- --config imports_granularity=Crate
 
+generate-entities-from-blockscout database_url="postgres://postgres:admin@localhost:5432/blockscout":
+    docker run --network=host -e 'DATABASE_URL={{database_url}}' -e 'ECTO_USE_SSL=false' --entrypoint "/bin/sh" --rm blockscout/blockscout:6.8.0 -c 'bin/blockscout eval "Elixir.Explorer.ReleaseTasks.create_and_migrate()"'
+    DATABASE_URL={{database_url}} sea-orm-cli generate entity --lib -o entity/src
+    cargo fmt --all -- --config imports_granularity=Crate
+
 generate-migration database_url="postgres://postgres:admin@localhost:5432/blockscout?ssl=false" migration_file="migration/src/migration.sql":
-    docker run --network=host -e 'DATABASE_URL={{database_url}}' --entrypoint "/bin/sh" --rm blockscout/blockscout:4.1.8 -c 'bin/blockscout eval "Elixir.Explorer.ReleaseTasks.create_and_migrate()"'
+    docker run --network=host -e 'DATABASE_URL={{database_url}}' -e 'ECTO_USE_SSL=false' --entrypoint "/bin/sh" --rm blockscout/blockscout:6.8.0 -c 'bin/blockscout eval "Elixir.Explorer.ReleaseTasks.create_and_migrate()"'
```
sevenzing, line 36: move `6.8.0` to `version` variable at the top of justfile please

libs/blockscout-db/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-db"
-version = "4.1.8"
+version = "5.0.0"
```
sevenzing, line 3: version of crate should be the same as version of blockscout (6.8.0)

libs/blockscout-db/entity/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-db-entity"
-version = "4.1.8"
+version = "5.0.0"
```
sevenzing, line 3: .

libs/blockscout-db/migration/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-db-migration"
-version = "4.1.8"
+version = "5.0.0"
```
sevenzing, line 3: .

libs/blockscout-db/Cargo.toml
bragov4ik: hmmm... what if a breaking change is introduced but blockscout's major version is not changed? i.e. github.com/blockscout/blockscout-rs/pull/1051

koushiro: how about using `5.0.0+6.8.0` as the version?

bragov4ik: makes sense, @sevenzing do you agree?

bragov4ik: we could even do smth like "5.0.0+blockscout-6.8.0" or "5.0.0+blockscout.6.8.0" or "5.0.0+blockscout6.8.0"

sevenzing (approved):

PR #1043, opened 2024-08-30
chore(libs): add `migration_status` to db entity

bragov4ik:
Generated from latest version & manually picked the file, because straightforward update [fails](https://github.com/blockscout/blockscout-rs/pull/1041) and I need to work with the table.

Note that even thought the code was generated by 1.0.x version, it still compiles well with 0.12.x sea-orm. Did not expect, but it works, so whatever.


bragov4ik:
won't be merged; I managed to update all entities successfully (with a few tweaks) 

see https://github.com/blockscout/blockscout-rs/pull/1041

PR #1044, opened 2024-08-30
perf(stats): query optimization

bragov4ik:
WIP.

Changes:
- blockscout migration detection
- (WIP) conditional statement choice (optimized or old) based on applied migrations


bragov4ik:
blocked by https://github.com/blockscout/blockscout-rs/pull/1041

bragov4ik:
I guess we'll have to kostyl something here..

stats/stats/src/charts/lines/average_gas_price.rs
```diff
@@ -31,7 +34,7 @@ const GWEI: i64 = 1_000_000_000;
 pub struct AverageGasPriceStatement;
 
 impl StatementFromRange for AverageGasPriceStatement {
-    fn get_statement(range: Option<Range<DateTimeUtc>>) -> Statement {
+    fn get_statement(range: Option<Range<DateTimeUtc>>, _: &BlockscoutMigrations) -> Statement {
         sql_with_range_filter_opt!(
             DbBackend::Postgres,
             r#"
```
rimrakhimov, line 40: It is needed to be checked (e.g., via running `explain analyze`), but I may suggest that adding additional `transactions.block_timestamp != to_timestamp(0) AND transactions.block_consensus = true` into where condition may improve the speed of the query because some of transactions would be eliminated before the join. 

It may also help to apply similar logic to range filter condition, but I haven't completely get how the generation of range filter statement works

stats/stats/src/charts/lines/average_txn_fee.rs
```diff
@@ -33,7 +36,7 @@ const ETHER: i64 = i64::pow(10, 18);
 pub struct AverageTxnFeeStatement;
 
 impl StatementFromRange for AverageTxnFeeStatement {
-    fn get_statement(range: Option<Range<DateTimeUtc>>) -> Statement {
+    fn get_statement(range: Option<Range<DateTimeUtc>>, _: &BlockscoutMigrations) -> Statement {
```
rimrakhimov, line 39: Similar to `AverageGasPriceStatement` considerations apply

stats/stats/src/charts/lines/txns_fee.rs
```diff
@@ -31,7 +34,7 @@ const ETHER: i64 = i64::pow(10, 18);
 pub struct TxnsFeeStatement;
 
 impl StatementFromRange for TxnsFeeStatement {
-    fn get_statement(range: Option<Range<DateTimeUtc>>) -> Statement {
+    fn get_statement(range: Option<Range<DateTimeUtc>>, _: &BlockscoutMigrations) -> Statement {
```
rimrakhimov, line 37: Similar to `AverageGasPriceStatement` considerations apply

stats/stats/src/charts/counters/completed_txns.rs
```diff
@@ -13,31 +16,57 @@ use sea_orm::{DbBackend, Statement};
...
+                            WHERE t.status = 1
+                        ), (
+                            SELECT COUNT(*) as all_success_dropped
+                            FROM transactions t
+                            WHERE t.status = 1 AND t.block_consensus = false
+                        ), (
+                            SELECT MAX(b.timestamp)::DATE AS last_block_date
+                            FROM blocks b
+                            WHERE b.consensus = true
+                        )
```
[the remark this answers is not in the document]
bragov4ik: So, the query in this case would be
```sql
SELECT
    (all_success_not_dropped)::TEXT AS value,
    last_block_date AS date 
FROM (
    SELECT (
        SELECT COUNT(*) as all_success_not_dropped
        FROM transactions t
        WHERE t.status = 1 AND t.block_consensus = true
    ), (
        SELECT MAX(b.timestamp)::DATE AS last_block_date
        FROM blocks b
        WHERE b.consensus = true
    )
) AS sub
```

I ran explain for current and suggested queries. Here are the results:
Current:
`Result  (cost=12420.87..12420.89 rows=1 width=36)`
Suggested:
`Result  (cost=23127.56..23127.58 rows=1 width=36)`

According to the rest of the input, the most expensive part of the current query is
```sql
SELECT COUNT(*) AS all_success
    FROM transactions t
    WHERE t.status = 1
```
which is scanned using index: `Parallel Index Only Scan using transactions_status_index on transactions t  (cost=0.42..11158.67 rows=102828 width=0)`. The other part is run very quickly because the `transactions_block_consensus_index` is utilized, which yields negligible amount of transactions (with consensus `false`).

In the suggested query, the expensive part is done twice as costly: `Parallel Seq Scan on transactions t  (cost=0.00..21869.82 rows=102828 width=0)`, likely because the index can't be utilized (for filter `Filter: (block_consensus AND (status = 1))`).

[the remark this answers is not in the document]
bragov4ik: After repeated `ANALYZE` runs the difference is even more apparent: in current query time lies in range~ 20..30, whereas for the suggested it's 10 times larger (200..300).
So, the current version better utilizes indexes that are present in `transactions` table. Apparently, they were created similarly to `blocks` table when denormalization of the table took place.

stats/stats/src/charts/lines/average_gas_price.rs
bragov4ik: Explain on a real data has shown very minor cost increase with the added filter. It seems that removing ~0.2% of rows yields less benefit than checking the condition on each transaction. (It was already scanned sequentially, so the only difference is added condition)

bragov4ik: also tried adding `{filter}` to the transactions - the results got much worse 🫠

bragov4ik: Wait, it seems it was an anomaly. Now with `filter` for `transactions` it's ~3-4 times faster

bragov4ik: also it makes sense according to explain, it now uses indexes instead of sequential scan

bragov4ik: tested on staging sepolia - `explain` with and without filter in txns shows barely any difference in planning. The only differences are:
- added that filter on `Seq scan` on table `transactions`.
- changed order of sort/aggregate at the end
It seems that it makes sense to filter as much txns as possible here.

bragov4ik: Tried executing with everything included. For current query metrics show >30 mins of execution for 1 day update + executing it by hand already took >1 minute (with only consensus & 0 timestamp filter). So I assume the metric is correct.

Now, prepare, executing it with added general timestamp filter took (!) 15 SECONDS. OK this is huge !!!

stats/stats/src/charts/lines/average_txn_fee.rs
bragov4ik: 1233.900141548s -> ~17s
definitely better

stats/stats/src/charts/lines/txns_fee.rs
bragov4ik: 2020.970182048s -> ~9.247s (sepolia) that's a win

stats/stats/src/charts/counters/total_blocks.rs
```diff
@@ -142,6 +143,7 @@ mod tests {
         let parameters = UpdateParameters {
             db: &db,
             blockscout: &blockscout,
+            blockscout_applied_migrations: BlockscoutMigrations::latest(),
```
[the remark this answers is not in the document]
bragov4ik: In most cases it's hidden inside helper `simple_test` (and similar ones), and in other I think it's better to keep it descriptive for better test readability.

lok52 (approved):

PR #1048, opened 2024-09-02
feat(env-collector): a few improvements for using in stats

bragov4ik:
- configurable anchor postfix
- more flexible filtering
- keep user-defined order in the table
- ignore non-ascii symbols in env names (to allow custom formatting with zero-width spaces, for example)
- more compact "Required" column
- more general allowed format for "default value" column

### DIFF
--- a/libs/env-collector/README.md
+++ b/libs/env-collector/README.md
@@ -24,7 +24,8 @@ This is a simple tool to collect possible environment variables from `Settings` 
             "<SERVICE_NAME_PREFIX>",
             "README.md",
             "<PATH TO .TOML/.JSON EXAMPLE CONFIG>",
-            &["<ENV_PREFIX_TO_IGNORE>"],
+            &[PrefixFilter::blacklist("<ENV_PREFIX_TO_IGNORE>")],
+            Some("some_postfix"),
         );
     }
     ```
@@ -33,8 +34,8 @@ This is a simple tool to collect possible environment variables from `Settings` 
     ```markdown
     ## Envs
 
-    [anchor]: <> (anchors.envs.start)
-    [anchor]: <> (anchors.envs.end)
+    [anchor]: <> (anchors.envs.start.some_postfix)
+    [anchor]: <> (anchors.envs.end.some_postfix)
     ```
 
 4. (Optional) In your `justfile` add new command to run `check-envs.rs`:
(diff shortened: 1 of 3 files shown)


libs/env-collector/README.md
```diff
@@ -24,7 +24,8 @@ This is a simple tool to collect possible environment variables from `Settings`
             "<SERVICE_NAME_PREFIX>",
             "README.md",
             "<PATH TO .TOML/.JSON EXAMPLE CONFIG>",
-            &["<ENV_PREFIX_TO_IGNORE>"],
+            &[PrefixFilter::blacklist("<ENV_PREFIX_TO_IGNORE>")],
```
sevenzing, line 27: you need to import `PrefixFilter` at line 20 to use it :)

libs/env-collector/src/lib.rs
```diff
@@ -332,13 +423,16 @@ fn from_key_to_json_path(key: &str, service_prefix: &str) -> String {
 }
 
 fn serialize_env_vars_to_md_table(vars: Envs) -> String {
+    // zero-width spaces in "Required" so that
+    // the word can be broken down and
+    // its colum doesn't take unnecessary space
     let mut result = r#"
-| Variable | Required | Description | Default value |
+| Variable | Req&#x200B;uir&#x200B;ed | Description | Default value |
```
sevenzing, line 430: 🤯👍

sevenzing (approved):

libs/env-collector/src/lib.rs
```diff
@@ -389,7 +491,7 @@ fn _flat_json(json: &Value, prefix: &str, env_vars: &mut BTreeMap<String, String
         _ => {
             let env_var_name = prefix.to_uppercase();
             let env_var_value = match json {
-                Value::String(s) => s.to_string(),
+                Value::String(s) => format!("\"{}\"", s),
```
sevenzing, line 494: well, I don't agree that we need to serialize string with quotes. because when you specify env variables you always specify strings, like `BOOL_VAR=false` and `STR_VAR=string` -- they are essentially the same when we specify them, however in docs you will show that default value is `false` for BOOL_VAR and `"string"` for STR_VAR. therefore this may add confusion and user will specify unnecessary quotes when setting a variable of type string

libs/env-collector/src/lib.rs
bragov4ik: 🙃  it makes sense though!!

PR #1051, opened 2024-09-09
chore(libs): upgrade blockscout-db to v1 sea-orm

bragov4ik:
To merge some day when we migrate to v1 sea-orm :)

### DIFF
--- a/libs/blockscout-db/Cargo.toml
+++ b/libs/blockscout-db/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-db"
-version = "0.1.0-blockscout.6.8.0"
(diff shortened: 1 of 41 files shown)


libs/blockscout-db/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-db"
-version = "5.0.0"
+version = "6.0.0"
```
sevenzing, line 3: version of blockscout-db should be the same as version of blockscout

bragov4ik: https://github.com/blockscout/blockscout-rs/pull/1041#discussion_r1751451240

PR #1069, opened 2024-09-27
refactor(stats):remove resolution dep

yjhmelody:


bragov4ik:
Overall it makes sense, though I think it makes sense to rename `MainDependencies` after this. I'll need some time to think this through if there are any problems; so I'll return to it later when I'll have some time.

PR #1072, opened 2024-09-27
fix(user-ops-indexer): stop indexing from already closed ws connections

k1rill-fedoseev:
When WS connection is closed, user-ops-indexer stream should be restarted immediately, without waiting for the whole stream to end.

### DIFF
--- a/user-ops-indexer/user-ops-indexer-logic/src/indexer/base_indexer.rs
+++ b/user-ops-indexer/user-ops-indexer-logic/src/indexer/base_indexer.rs
@@ -12,12 +12,12 @@ use ethers::prelude::{
     abi::{AbiEncode, Error},
     parse_log,
     types::{Address, Bytes, Filter, Log, TransactionReceipt},
-    EthEvent, Middleware, NodeClient, Provider, ProviderError, H256,
+    EthEvent, Middleware, NodeClient, Provider, ProviderError, WsClientError, H256,
 };
 use futures::{
     stream,
     stream::{repeat_with, BoxStream},
-    Stream, StreamExt,
+    Stream, StreamExt, TryStreamExt,
 };
 use sea_orm::DatabaseConnection;
 use std::{future, num::NonZeroUsize, sync::Arc, time, time::Duration};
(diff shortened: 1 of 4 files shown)


user-ops-indexer/user-ops-indexer-logic/src/indexer/base_indexer.rs
```diff
@@ -216,9 +216,16 @@ impl<L: IndexerLogic + Sync> Indexer<L> {
...
 
         stream_txs
-            .for_each_concurrent(Some(self.settings.concurrency as usize), |tx| async move {
+            .map(Ok)
+            .try_for_each_concurrent(Some(self.settings.concurrency as usize), |tx| async move {
                 let mut backoff = vec![5, 20, 120].into_iter().map(Duration::from_secs);
-                while let Err(err) = &self.handle_tx(tx, variant).await {
+                while let Err(err) = self.handle_tx(tx, variant).await {
+                    // terminate stream if WS connection is closed, indexer will be restarted
+                    if self.client.as_ref().supports_subscriptions() && err.to_string() == WsClientError::UnexpectedClose.to_string() {
```
rimrakhimov, line 224: Nit: just interested if you've checked this works in practice? Doesn't `anyhow` add anything additional to the message that changes it?

In theory, it seems that you may also try to downcast `anyhow::Error` to the `WsClientError` to compare it in a more direct way. But I'm also not sure that would work, so probably better to leave as it is, if you'd checked current approach
```suggestion
                    if self.client.as_ref().supports_subscriptions()
                        && err.downcast_ref::<WsClientError>().is_some_and(|err| matches!(err, WsClientError::UnexpectedClose)) {
```

rimrakhimov (approved):

user-ops-indexer/user-ops-indexer-logic/src/indexer/base_indexer.rs
k1rill-fedoseev: Yes, I've run a manual test to confirm anyhow wrapper converts to string without anything extra.

I've tried to fight with downcast, but the problem is that it's not WsClientError wrapped in anyhow, it's actually first wrapped in another Enum with `Box<dyn ...>`, and then in anyhow `Box<dyn ...>`. Haven't figured out a good and clean way to double-unwrap it. So gave up on downcasting.

k1rill-fedoseev: The final result was `anyhow::Error(ProviderError::JsonRpcClientError(Box::new(WsClientError::UnexpectedClose)))`

PR #1073, opened 2024-09-27
fix: too long db names in tests

k1rill-fedoseev:
Current behavior breaks tests on `postgres:17`


bragov4ik:
LGTM, sourcify tests are failing and its ok ;)

bragov4ik:
Oh, one thing: can you bump version of the launcher so that we can release it on `crates.io`? @k1rill-fedoseev

bragov4ik (approved):

PR #1075, opened 2024-10-02
feat(stats): wait until blockscout is indexed

bragov4ik:
Configurable thresholds for ratios returned from `indexing-status` endpoint; wait until they're satisfied & launch stats

### DIFF
--- a/stats/stats-server/src/server.rs
+++ b/stats/stats-server/src/server.rs
@@ -60,7 +60,7 @@ fn grpc_router<S: StatsService>(
 }
 
 fn is_threshold_passed(
-    threshold: Option<f32>,
+    threshold: Option<f64>,
     float_value: Option<String>,
     value_name: &str,
 ) -> Result<bool, anyhow::Error> {
@@ -262,8 +262,8 @@ mod tests {
 
     #[fixture]
     fn wait_config(
-        #[default(0.9)] blocks: f32,
-        #[default(0.9)] internal_transactions: f32,
+        #[default(0.9)] blocks: f64,
+        #[default(0.9)] internal_transactions: f64,
     ) -> StartConditionSettings {
         StartConditionSettings {
             enabled: true,
--- a/stats/stats-server/src/server.rs
+++ b/stats/stats-server/src/server.rs
@@ -112,11 +112,11 @@ async fn wait_for_blockscout_indexing(
         match blockscout_client::apis::main_page_api::get_indexing_status(&api_config).await {
             Ok(result)
                 if is_threshold_passed(
-                    &wait_config.blocks_ratio_threshold,
+                    &wait_config.blocks_ratio,
                     result.indexed_blocks_ratio.clone(),
                     "indexed_blocks_ratio",
                 )? && is_threshold_passed(
-                    &wait_config.internal_transactions_ratio_threshold,
+                    &wait_config.internal_transactions_ratio,
                     result.indexed_internal_transactions_ratio.clone(),
                     "indexed_internal_transactions_ratio",
                 )? =>
@@ -289,10 +289,8 @@ mod tests {
         #[default(0.9)] internal_transactions: f64,
     ) -> StartConditionSettings {
         StartConditionSettings {
-            blocks_ratio_threshold: ToggleableThreshold::enabled(blocks),
-            internal_transactions_ratio_threshold: ToggleableThreshold::enabled(
-                internal_transactions,
-            ),
+            blocks_ratio: ToggleableThreshold::enabled(blocks),
+            internal_transactions_ratio: ToggleableThreshold::enabled(internal_transactions),
             check_period_secs: 0,
         }
     }
--- a/stats/stats-server/src/server.rs
+++ b/stats/stats-server/src/server.rs
@@ -11,7 +11,7 @@ use crate::{
 
 use anyhow::Context;
 use blockscout_endpoint_swagger::route_swagger;
-use blockscout_service_launcher::launcher::{self, LaunchSettings};
+use blockscout_service_launcher::launcher::{self, ConfigSettings, LaunchSettings};
 use reqwest::StatusCode;
 use sea_orm::{ConnectOptions, Database};
 use stats::metrics;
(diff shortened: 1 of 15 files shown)


stats/stats-server/src/server.rs
```diff
@@ -134,3 +221,178 @@ pub async fn stats(settings: Settings) -> Result<(), anyhow::Error> {
...
+            check_period_secs: 0,
+        };
+        test_wait_indexing(
+            wait_config.clone(),
+            ResponseTemplate::new(200).set_body_string(
+                r#"{
+                    "finished_indexing": true,
+                    "finished_indexing_blocks": true,
+                    "indexed_blocks_ratio": "1.00",
+                    "indexed_internal_transactions_ratio": "1"
```
rimrakhimov, line 277: Nit: to be consistent with actual response
```suggestion
                    "indexed_internal_transactions_ratio": "1.00"
```

stats/stats-server/src/server.rs
```diff
@@ -134,3 +221,178 @@ pub async fn stats(settings: Settings) -> Result<(), anyhow::Error> {
...
+    }
+
+    #[tokio::test]
+    async fn wait_for_blockscout_indexing_works_with_200_response() {
+        let wait_config = StartConditionSettings {
+            enabled: true,
+            blocks_ratio_threshold: Some(0.9),
+            internal_transactions_ratio_threshold: Some(0.9),
+            check_period_secs: 0,
+        };
```
rimrakhimov, line 269: Nit: maybe extract that in a function, so that you don't need to define the same config in all tests. Otherwise, if something will change in the struct, it may be quite painful to update all definitions

I may also suggest to use `rstest::fixtures` (https://docs.rs/rstest/latest/rstest/attr.fixture.html) in order to avoid defining it in all tests manually

stats/stats-server/src/server.rs
```diff
@@ -56,6 +59,85 @@ fn grpc_router<S: StatsService>(
         .add_service(StatsServiceServer::from_arc(stats))
 }
 
+fn is_threshold_passed(
+    threshold: Option<f32>,
```
rimrakhimov, line 63: I believe it make sense to make `threshold` we set during configuration also `f64`, considering we parse api response to `f64` value

rimrakhimov, line 63: I believe it make sense to make `threshold` we set during configuration also `f64`, considering we parse api response to `f64` value

rimrakhimov (approved):

stats/stats-server/src/server.rs
bragov4ik: Sometimes endpoints return "1" :) decided to add to the test just in case

stats/stats-server/src/server.rs
bragov4ik: yeah makes sense

bragov4ik:
just for the record: the new `IGNORE_​​BLOCKSCOUT_​API_​ABSENCE` env is created with some upcoming changes in mind; we'll very likely have some charts depend on data from blockscout api, so this single env should simplify deployment without setting up the API

stats/stats-server/src/server.rs
```diff
@@ -56,6 +59,106 @@ fn grpc_router<S: StatsService>(
...
+                if is_threshold_passed(
+                    &wait_config.blocks_ratio,
+                    result.indexed_blocks_ratio.clone(),
+                    "indexed_blocks_ratio",
+                )? && is_threshold_passed(
+                    &wait_config.internal_transactions_ratio,
+                    result.indexed_internal_transactions_ratio.clone(),
+                    "indexed_internal_transactions_ratio",
+                )? =>
+            {
```
sevenzing, line 123:
```suggestion
            Ok(result)
                if is_threshold_passed(
                    &wait_config.blocks_ratio,
                    result.indexed_blocks_ratio.clone(),
                    "indexed_blocks_ratio",
                ).context("check index block ratio")? && is_threshold_passed(
                    &wait_config.internal_transactions_ratio,
                    result.indexed_internal_transactions_ratio.clone(),
                    "indexed_internal_transactions_ratio",
                ).context("check internal transaction index")? =>
            {
```

stats/stats-server/src/server.rs
```diff
@@ -56,6 +59,106 @@ fn grpc_router<S: StatsService>(
...
+                info!("Blockscout indexing threshold passed");
+                return Ok(());
+            }
+            Ok(_) => {}
```
sevenzing, line 127: lets print some message that we're actually waiting for something every `10*wait_config.check_period_secs`

stats/stats-server/src/server.rs
```diff
@@ -56,6 +59,106 @@ fn grpc_router<S: StatsService>(
...
+        }
+        info!(
+            "Blockscout is not indexed enough. Checking again in {} secs",
+            wait_config.check_period_secs
+        );
+        sleep(Duration::from_secs(wait_config.check_period_secs.into())).await;
+    }
+}
+
+async fn init_api(settings: &Settings) -> anyhow::Result<Option<blockscout_client::Configuration>> {
```
sevenzing, line 143: init_api sounds like we are initializing stats api for some reason. lets name it init_blockscout_client

stats/stats-server/src/server.rs
```diff
@@ -134,3 +244,172 @@ pub async fn stats(settings: Settings) -> Result<(), anyhow::Error> {
 
     launcher::launch(&launch_settings, http_router, grpc_router).await
 }
+
+#[cfg(test)]
+mod tests {
```
sevenzing, line 249: let's move tests and `async fn wait_for_blockscout_indexing(` and all related functions to sep file please, this is too much for `server.rs`

sevenzing: let's name it `blockscout_waiter.rs`  🤪
image

stats/stats-server/src/server.rs
bragov4ik: https://github.com/blockscout/blockscout-rs/pull/1075/files/8befb1ae0760791a272dfa774939ea332e6328ea#diff-acc71d314599718d9f08f24d7a7caa47a45dc53bb235fdfbd00f0e3d42768790R136

stats/stats-server/src/server.rs
bragov4ik: This name is supplied to errors from the 3rd argument anyway, so I see no reason to repeat it.
One could replace it with this context but then we need to figure out how to pass it to logs (spans? sounds kinda repetitive to add)

PR #1077, opened 2024-10-03
feat(libs): add blockscout-chains crate

lok52:


libs/blockscout-chains/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+    let retry_policy = ExponentialBackoff::builder().build_with_max_retries(max_retries);
+    let client = ClientBuilder::new(reqwest::Client::new())
+        .with(RetryTransientMiddleware::new_with_policy(retry_policy))
+        .build();
+    let res = client.get(CHAINS_URL).send().await?;
+    let chains: BlockscoutChains = res.json().await?;
+    Ok(chains)
+}
+
+pub type BlockscoutChains = HashMap<i64, BlockscoutChainData>;
```
rimrakhimov, line 19: Maybe represent `chain_id` as a `u64` instead?

libs/blockscout-chains/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct BlockscoutChainData {
+    pub name: String,
+    pub description: String,
+    pub ecosystem: Ecosystem,
+    pub is_testnet: Option<bool>,
```
rimrakhimov, line 27: Nit: just interested, what is the case when `is_testnet` is missing?

libs/blockscout-chains/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+        .with(RetryTransientMiddleware::new_with_policy(retry_policy))
+        .build();
+    let res = client.get(CHAINS_URL).send().await?;
+    let chains: BlockscoutChains = res.json().await?;
+    Ok(chains)
+}
+
+pub type BlockscoutChains = HashMap<i64, BlockscoutChainData>;
+
+#[derive(Deserialize, Debug)]
```
rimrakhimov, line 21: May we also add `Clone, PartialEq, Eq`, so it be more generic?

libs/blockscout-chains/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct BlockscoutChainData {
+    pub name: String,
+    pub description: String,
+    pub ecosystem: Ecosystem,
+    pub is_testnet: Option<bool>,
+    pub layer: Option<u8>,
```
rimrakhimov, line 28: Nit: also, when it may be missing?

libs/blockscout-chains/src/lib.rs
lok52: Just copied that from the original interface:
https://github.com/blockscout/chainscout/blob/7657da41a7dda8c180cfb9ee2fc858c3d27c2dfc/types/index.ts#L12

libs/blockscout-chains/src/lib.rs
lok52: Even though the original interface specifies this field to be always present, I found some chains where it is `null`.
You can check the response of https://chains.blockscout.com/api/chains and search for `"layer": null`

libs/blockscout-chains/src/lib.rs
lok52: Thanks, updated

libs/blockscout-chains/src/lib.rs
rimrakhimov: Get it, thanks. Let's maybe ask them to fix that inside their interface as well?

libs/blockscout-chains/src/lib.rs
rimrakhimov: Oh, I see. May we also ask Max why it was the case?

rimrakhimov (approved):

libs/blockscout-chains/src/lib.rs
```diff
@@ -0,0 +1,58 @@
+use reqwest_middleware::ClientBuilder;
+use reqwest_retry::{policies::ExponentialBackoff, RetryTransientMiddleware};
+use serde::Deserialize;
+use std::collections::HashMap;
+
+const CHAINS_URL: &str = "https://chains.blockscout.com/api/chains";
+
+pub async fn get_blockscout_chains() -> anyhow::Result<BlockscoutChains> {
```
sevenzing, line 8: why anyhow?

sevenzing: maybe use tokio error?

libs/blockscout-chains/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+    let max_retries = 3;
```
sevenzing, line 9: nit: write builder to have access to max_retries and chains_url, like

```
let chains = BlockscoutChainsClient::default().fetch_all().await.unwrap()
```
or 
```
let chains = BlockscoutChainsClient::builder().with_max_retries(3).with_host(HOST).build().unwrap().fetch_all().await.unwrpa()
```

libs/blockscout-chains/src/lib.rs
lok52: Decided to leave the `Option` type as was discussed on the call

libs/blockscout-chains/src/lib.rs
lok52: Decided to leave the Option type as was discussed on the call

libs/blockscout-chains/src/lib.rs
lok52: Used `reqwest_middleware::Error` instead

libs/blockscout-chains/src/lib.rs
lok52: Done

PR #1084, opened 2024-10-09
fix: change Decimal to BigDecimal

sevenzing:


sl1depengwyn (commented):
Why you haven't change `token_id` type?

sevenzing:
@sl1depengwyn it didn't compile for some reason, will fix it later

PR #1092, opened 2024-10-24
fix(verifier): return evm compiler version actually used for successful verification

rimrakhimov:
Right now the versions used in the request are returned. For example, if user specified a usual v0.8.28+commit.7893614a solidity compiler in the request it will be returned even though era-solidity v0.8.28+commit.acc7d8f9 resulted in successful verification

### DIFF
--- a/smart-contract-verifier/smart-contract-verifier-server/tests/test_cases_zksync_solidity/era_solidity_0.8.28.json
+++ b/smart-contract-verifier/smart-contract-verifier-server/tests/test_cases_zksync_solidity/era_solidity_0.8.28.json
@@ -27,5 +27,7 @@
     },
     "expected_runtime_match_type": "full",
     "expected_runtime_transformations": [],
-    "expected_runtime_values": {}
+    "expected_runtime_values": {},
+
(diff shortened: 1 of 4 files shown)


rimrakhimov:
There is a problem with zksolc_1_3_5 test (https://github.com/blockscout/blockscout-rs/actions/runs/11501807515/job/32015247302?pr=1092). It seems that zksolc v1.3.5 does not differentiate between solidity compilers that results in any era-solidity compiler (and the usual solidity as well) verify given contract. Because ListFetcher keeps versions inside a HashMap internally, that results in random era-solidity compiler version returned in case of verification success, resulting in floppy test case.

In general, that is just a test case problem, which may be fixed by specifying a list of possible compiler responses. But I believe that reveals more generic issue. There may be several era-solidity compilers with the same evm version (e.g., v0.8.28), but different zksync revisions (e.g, v1.0.0 vs v1.0.1). In most cases we want the latest revision to be tried first, as it has the most chances to be successful and reduce the average verification time. But currently, we just use a random order for them. It may be especially noticeable if more compilers with the same evm versions are added in the future.

PR #1101, opened 2024-11-01
feat(bens): d3 connect

sevenzing:
Refactor code and add d3-connect type of protocol. it will resolve domains on-demand using offchain resolution technique https://docs.d3.app/resolve-d3-names

todo:

- [x] handle reverse records
- [x] add env-collector
- [x] move deps into cargo workspace


bragov4ik (approved):
checked rust part on the surface

the rest I don't understand but it's very interesting (ничего не понятно но очень интересно)

PR #1120, opened 2024-11-13
feat(stats): Back3nd chart5 Migration (Part 2: yesterday's transactions)

bragov4ik:
It is the sequel to [The Migration of The Backend charts](#1117) and the second installment in the Migrate Backend Charts franchise. In the PR counter for yesterday transactions is added.

### DIFF
--- a/stats/stats/src/charts/counters/yesterday_txns.rs
+++ b/stats/stats/src/charts/counters/yesterday_txns.rs
@@ -1,0 +1,94 @@
+use std::ops::Range;
+
+use crate::{
+    data_source::{
+        kinds::{
+            local_db::DirectPointLocalDbChartSource,
+            remote_db::{RemoteDatabaseSource, RemoteQueryBehaviour, StatementFromRange},
+        },
+        UpdateContext,
+    },
+    lines::NewTxnsStatement,
+    types::TimespanValue,
+    utils::day_start,
+    ChartProperties, MissingDatePolicy, Named, UpdateError,
+};
+use chrono::{DateTime, Days, NaiveDate, Utc};
+use entity::sea_orm_active_enums::ChartType;
+use sea_orm::FromQueryResult;
+
+pub struct YesterdayTxnsQuery;
+
+impl RemoteQueryBehaviour for YesterdayTxnsQuery {
+    type Output = TimespanValue<NaiveDate, String>;
+
+    async fn query_data(
+        cx: &UpdateContext<'_>,
+        _range: Option<Range<DateTime<Utc>>>,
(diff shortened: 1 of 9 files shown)


stats/stats/src/charts/lines/mod.rs
```diff
@@ -88,6 +88,7 @@ pub use new_native_coin_transfers::{
     NewNativeCoinTransfers, NewNativeCoinTransfersInt, NewNativeCoinTransfersMonthly,
     NewNativeCoinTransfersWeekly, NewNativeCoinTransfersYearly,
 };
+pub(crate) use new_txns::NewTxnsStatement;
```
sevenzing, line 91: either replace all reimports or change to `pub use` for consistency

stats/stats/src/charts/counters/yesterday_txns.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl ChartProperties for Properties {
+    type Resolution = NaiveDate;
+
+    fn chart_type() -> ChartType {
+        ChartType::Counter
+    }
+    fn missing_date_policy() -> MissingDatePolicy {
+        MissingDatePolicy::FillPrevious
```
sevenzing, line 67: why do u need to specify missing_date_policy for counter? maybe leave default trait value?

stats/stats/src/charts/lines/mod.rs
bragov4ik: why? it's only needed within the crate. it is not meant to be exposed for `server` part, for example

stats/stats/src/charts/counters/yesterday_txns.rs
bragov4ik: so you suggest to display zero value from midnight until the update is finished?

bragov4ik: because [this is what it means](https://github.com/blockscout/blockscout-rs/blob/119090cf683a7e3eb0ab1abdb868fd4add9f2121/stats/stats/src/charts/db_interaction/read.rs#L126)

stats/stats/src/charts/lines/mod.rs
bragov4ik: resolving bc vacation + answer given 🙈

stats/stats/src/charts/counters/yesterday_txns.rs
bragov4ik: resolving bc vacation + answer given 🙈

sevenzing (approved):

PR #1127, opened 2024-11-18
fix(stats): v2.2.3 fixes

bragov4ik:
- fix `null` indexing-status: When blocks haven't finished indexed, `indexed_internal_transactions_ratio` is returned `null`. We want to handle it properly instead of falling with an error.
- fix readme
- fix issue with service being terminated continuously


bragov4ik:
I fit many fixes in one PR because we need to release the fix as v2.2.3 asap and `master` contains some new features (which will go to v2.3.0)

stats/stats-server/src/update_service.rs
```diff
@@ -49,32 +55,45 @@ impl UpdateService {
...
+    ) -> JoinHandle<()> {
         let this = self.clone();
         let chart = group_entry.clone();
         let schedule = group_entry
             .update_schedule
             .as_ref()
             .unwrap_or(default_schedule)
             .clone();
-        tokio::spawn(async move { this.run_cron(chart, schedule).await });
+        tokio::spawn(async move { this.run_cron(chart, schedule).await })
```
rimrakhimov, line 96: Nit: probably should do the same
```suggestion
        tokio::spawn(this.run_cron(chart, schedule))
```

rimrakhimov (approved):

PR #1134, opened 2024-12-02
feat(eth-bytecode-db): update verifier alliance schema to v1

rimrakhimov:
1. Updated verifier alliance migration and entity crates to use the latest v1 verifier-alliance database schema: https://github.com/verifier-alliance/database-specs/blob/master/database.sql
2. Updated `verification-common` library to support the latest schema
3. Initialized `verifier-alliance-database` crate responsible for all verifier-alliance data manipulations. Eth-bytecode-db will make use of it in the next PRs

Note: eth-bytecode-db was not updated and still uses v0 schema

## Release Notes

- **New Features**
	- Introduced a new `verifier-alliance-database` package with enhanced dependency management.
	- Added support for new fields in various data models, including timestamps and user tracking for contracts and deployments.
	- New `insert_then_select` function for simplified database operations.
	- Added `SourceId` struct and enhanced `CompilationArtifacts` to include source information.
	- Enhanced `CreationCodeArtifacts` and `RuntimeCodeArtifacts` to ensure all fields are serialized.
	- Updated `Match` structure to simplify match result representation.

- **Bug Fixes**
	- Improved error handling in database operations and tests.

- **Tests**
	- Added comprehensive integration tests for contract deployments and verifications.
	- Introduced new JSON test cases to validate contract deployment scenarios, including various match types.

- **Documentation**
	- Updated dependencies and module structure for better organization and clarity.


eth-bytecode-db/verifier-alliance-database/src/helpers.rs
```diff
@@ -0,0 +1,63 @@
...
+                            $(
+                                .filter($entity_module::Column::$column.eq($value))
+                            )*
+                            .one($txn)
+                            .await
+                            .context(format!("select from \"{}\" by unique columns", stringify!($entity_module)))?
+                            .ok_or(anyhow::anyhow!("select from \"{}\" by unique columns returned no data", stringify!($entity_module)))?;
+                    // The active model have not been inserted.
+                    // Thus, there were a value already that we need to update.
+                    if $update_on_conflict {
```
bragov4ik, line 40: Can you explain why is this custom logic needed? I seem to miss the point 😅

eth-bytecode-db/verifier-alliance-database/src/lib.rs
```diff
@@ -0,0 +1,132 @@
+/// Provides access to internal functions to access the database.
+/// They mostly do not provide transactions consistency, and require
+/// users to be care of transactions themselves.
+///
+/// Are not recommended to be used directly.
+/// Prefer methods exposed to the public instead.
+pub mod internal;
```
bragov4ik, line 7: it seems to not be used anywhere else, maybe make it private?

eth-bytecode-db/verifier-alliance-database/tests/integration/main.rs
```diff
@@ -0,0 +1,23 @@
+mod contract_deployments;
+mod internal_compiled_contracts;
+mod transformations;
+mod transformations_types;
+mod verified_contracts;
+
+macro_rules! database {
```
bragov4ik, line 7: maybe merge the changes to service-launcher and use the macros from there?

bragov4ik (approved):
lgtm overall, see minor comments

sevenzing:
@coderabbitai full review

eth-bytecode-db/Cargo.toml
```diff
@@ -16,8 +17,13 @@ eth-bytecode-db = { path = "eth-bytecode-db" }
 eth-bytecode-db-proto = { path = "eth-bytecode-db-proto" }
 eth-bytecode-db-server = { path = "eth-bytecode-db-server" }
 migration = { path = "eth-bytecode-db/migration" }
-verifier-alliance-entity = { path = "eth-bytecode-db/verifier-alliance-entity" }
-verifier-alliance-migration = { path = "eth-bytecode-db/verifier-alliance-migration" }
+verifier-alliance-entity = { git = "https://github.com/blockscout/blockscout-rs", rev = "ec1c755" }
+verifier-alliance-database = { path = "verifier-alliance-database" }
```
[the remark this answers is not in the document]
rimrakhimov: Currently we use the previous version migrations and entity inside the `eth-bytecode-db` crate, that is why they have been updated to use git tags, while the migration and entity crates themselves have been changed to support the latest database schema. After new schema is integrated inside `eth-bytecode-db` those git dependencies will be removed

eth-bytecode-db/verifier-alliance-database/src/lib.rs
rimrakhimov: It is used inside the integration tests. At the same time, I would like to leave the flexibility of using internal implementation details if user's code needs that. So I prefer to make it public at least for now

eth-bytecode-db/verifier-alliance-database/src/helpers.rs
rimrakhimov: This macro definition was just copied from `eth-bytecode-db` crate. Inside `eth-bytecode-db` there were both cases - when inserted data should override existing ones and when shouldn't. I don't think we update anything inside `verifier-alliance-database` crate, but I'm planning to remove the copypaste and just reuse one macro definition from both crates later, so decided not to update the interface here

eth-bytecode-db/verifier-alliance-database/src/helpers.rs
```diff
@@ -0,0 +1,63 @@
+macro_rules! insert_then_select {
```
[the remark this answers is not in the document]
rimrakhimov: Good idea :) Haven't thought that it is possible to make it a function

eth-bytecode-db/verifier-alliance-database/src/internal.rs
```diff
@@ -0,0 +1,470 @@
...
+}
+
+pub async fn retrieve_contract_deployment<C: ConnectionTrait>(
+    database_connection: &C,
+    contract_deployment: RetrieveContractDeployment,
+) -> Result<Option<contract_deployments::Model>, Error> {
+    let transaction_hash = contract_deployment.transaction_hash.unwrap_or_else(|| {
+        let runtime_code = contract_deployment
+            .runtime_code
+            .expect("either transaction hash or runtime code must contain value");
```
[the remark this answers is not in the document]
rimrakhimov: `RetrieveContractDeployment` has to support invariant that at least one of `transaction_hash` or `runtime_code` is not None. That is maintained by using one of two structure constructors resulting in at least one of values always being present

eth-bytecode-db/verifier-alliance-database/src/helpers.rs
rimrakhimov: As far as I remember it requires the columns to be specified. But in our case some of uniqueness of some of the columns was specified by their hash instead of themselves. That resulted in `.on_conflict` option being not working. I will check if that is still the case in the next PRs

PR #1135, opened 2024-12-02
feat(launcher): add test database initialization helping macros

rimrakhimov:
Can be used inside test in the following way:

```
use blockscout_service_launcher::test_database::database;
...
#[tokio::test]
async fn test() {
    let db_guard = database!(service_migration_crate);
    ...
}
```


bragov4ik:
Let's add the note and usage example to the macro documentation?

bragov4ik (approved):

PR #1140, opened 2024-12-04
Make bens image name the same as release tag

alik-agaev:


sevenzing (approved):

rimrakhimov:
I suggest to rename `blockscout-ens` directory to be `bens` instead of passing `docker-context-path` and `dockerfile-path` explicitly. In my opinion, this kind of flexibility is unnecessary as almost all services except `blockscout-ens` and `multichain-search` follows the same structure with service names being the same as directories. `multichain-search` is deprecated overall, so renaming `blockscout-ens` to `bens` will result in all services being the same

sevenzing:
will rename blockscout-ens in next pr, for now merge it

PR #1152, opened 2024-12-12
fix(libs): change chain_id type in blockscout-chains response

lok52:
## Summary by CodeRabbit

- **New Features**
	- Updated the package version of `blockscout-chains` to 0.1.1.
  
- **Changes**
	- Altered the key type in the `BlockscoutChains` structure from `u64` to `String`, impacting how data is indexed and retrieved from the API.

### DIFF
--- a/libs/blockscout-chains/Cargo.toml
+++ b/libs/blockscout-chains/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-chains"
-version = "0.1.1"
+version = "0.2.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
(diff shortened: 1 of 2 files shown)


libs/blockscout-chains/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-chains"
-version = "0.1.0"
+version = "0.1.1"
```
rimrakhimov, line 3: Do we make use of that somewhere already? It is probably better to bump it to "0.2.0", as otherwise it may be updated automatically and cause compilation errors

rimrakhimov (approved):

libs/blockscout-chains/Cargo.toml
lok52: Yes, we use it in metadata service already. Updated to `0.2.0`

PR #1156, opened 2024-12-18
release(stats): v2.3.0

bragov4ik:
Not releasing from main because main contains WIP main page chart migrations


bragov4ik:
didn't notice any problems from rebase after looking through the code

PR #1161, opened 2024-12-23
feat(api-client-framework): initial implementation

rimrakhimov:
A framework to be used when writing custom http api clients.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new HTTP API client framework for making asynchronous API requests.
	- Added a structured way to define API endpoints through a new `Endpoint` trait.
	- Included comprehensive error handling mechanisms within the API client.
	- Added a new package for the API client framework to the project workspace.

- **Bug Fixes**
	- Enhanced error handling for various API-related issues, including deserialization and HTTP status errors.

- **Documentation**
	- Updated package configurations and added relevant descriptions for the new framework components.

### DIFF
--- a/libs/api-client-framework/src/async_client.rs
+++ b/libs/api-client-framework/src/async_client.rs
@@ -6,6 +6,7 @@ use reqwest_retry::{policies::ExponentialBackoff, RetryTransientMiddleware};
 use serde::Deserialize;
 use std::time::Duration;
 
+#[derive(Clone)]
 pub struct HttpApiClientConfig {
     /// The maximum time limit for an API request. If a request takes longer than this, it will be
     /// cancelled.
@@ -26,6 +27,7 @@ impl Default for HttpApiClientConfig {
     }
 }
 
+#[derive(Clone)]
 pub struct HttpApiClient {
     base_url: url::Url,
     http_client: reqwest_middleware::ClientWithMiddleware,
--- a/libs/api-client-framework/src/async_client.rs
+++ b/libs/api-client-framework/src/async_client.rs
@@ -75,7 +75,7 @@ impl HttpApiClient {
 async fn process_api_response<T: for<'a> Deserialize<'a>>(response: Response) -> Result<T, Error> {
     let status = response.status();
     match status {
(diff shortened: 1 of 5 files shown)


libs/api-client-framework/src/endpoint.rs
```diff
@@ -0,0 +1,55 @@
+use serde::{Deserialize, Serialize};
+use std::{borrow::Cow, fmt::Debug};
+use url::Url;
+
+/// Represents a specification for an API call that can be built into an HTTP request and sent.
+/// New endpoints should implement this trait.
+///
+/// If the request succeeds, the call will resolve to a `Response`.
+pub trait Endpoint {
```
sevenzing, line 9: any way to define custom headers? to pass api key to endpoint for example

sevenzing: btw found this implementation https://github.com/hephex/api/blob/master/src/lib.rs#L162 , Im not sure we want to use it but take a look at different implementation is good practice

libs/api-client-framework/src/async_client.rs
```diff
@@ -0,0 +1,92 @@
...
+    match status {
+        StatusCode::OK => (),
+        StatusCode::NOT_FOUND => return Err(Error::NotFound),
+        status => {
+            return Err(Error::InvalidStatusCode {
+                status_code: status,
+                message: response.text().await?,
+            })
+        }
+    }
```
sevenzing, line 86: What if status code is 201?

libs/api-client-framework/src/endpoint.rs
rimrakhimov: I suggest to add custom headers later when we are required. Agree that it could be useful, but prefer to consider it as mvp for now

libs/api-client-framework/src/async_client.rs
rimrakhimov: Updated in [d9d7e17](https://github.com/blockscout/blockscout-rs/pull/1161/commits/d9d7e17ec1af2c6f9686cebf0adae754dd409711)

PR #1176, opened 2025-01-06
feat(stats): dynamic indexing status wait

bragov4ik:
Start chart updates dynamically when next indexed status is observed. Basically makes launching the statistics service more convenient and simple


bragov4ik:
@coderabbitai review

stats/stats-server/src/update_service.rs
```diff
@@ -27,74 +31,91 @@ fn time_till_next_call(schedule: &Schedule) -> std::time::Duration {
...
+            status_listener,
         })
     }
 
+    /// The main function of the service.
+    ///
     /// Perform initial update and run the service in infinite loop.
     /// Terminates dependant threads if one fails.
-    pub async fn force_async_update_and_run(
+    pub async fn run(
```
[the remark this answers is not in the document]
bragov4ik: 1. what do you mean?
2. no, we don't want to calculate statistics w/o ensuring proper indexation
3. not sure what you mean, the function doesn't return errors
4. they don't increment

stats/stats-server/src/blockscout_waiter.rs
```diff
@@ -4,11 +4,133 @@ use crate::settings::{Settings, StartConditionSettings, ToggleableThreshold};
...
+                            let modified = self.sender.send_if_modified(|val| {
+                                if val != &status {
+                                    *val = status.clone();
+                                    true
+                                } else {
+                                    false
+                                }
+                            });
+                            if modified {
+                                info!("Observed new indexing status: {:?}", status);
```
sevenzing, line 75: we have an agreement to use tracing::info! explicitly

stats/stats-server/src/update_service.rs
```diff
@@ -27,74 +31,98 @@ fn time_till_next_call(schedule: &Schedule) -> std::time::Duration {
...
-        tracing::info!("initial update is done");
+            .collect();
 
         let mut failed = 0;
-        while !updaters.is_empty() {
-            let (res, _, others) = futures::future::select_all(updaters).await;
-            updaters = others;
-            tracing::error!("updater stopped: {:?}", res);
-
+        while let Some(()) = group_update_jobs.next().await {
```
sevenzing, line 104: maybe it was too late on the clock for me to review this but I spend like 10 minutes to understand why `while let Some(()) =` is considered as failed update, let's add comment like

```
// These futures should never complete because they run in infinite loop.
// If any completes, it means something went wrong.
```

stats/stats-server/src/update_service.rs
```diff
@@ -27,74 +31,98 @@ fn time_till_next_call(schedule: &Schedule) -> std::time::Duration {
...
-        let (tasks, mut updaters) = self
-            .charts
-            .update_groups
-            .values()
+        let initial_update_semaphore: Arc<tokio::sync::Semaphore> =
+            Arc::new(tokio::sync::Semaphore::new(concurrent_initial_tasks));
+        let groups = self.charts.update_groups.values();
+        let init_update_tracker = InitialUpdateTracker::new(groups.len() as u64);
+        let mut group_update_jobs: FuturesUnordered<_> = groups
             .map(|group| {
```
sevenzing, line 61: nit: this function is rather big, it would be great to move complicated logic and show only business logic, like

```rust
.map(|group| {
    wait_until_status_ok(group)
    initial_group_update(group)
    run_infinite_update_loop(group)
})
```

stats/stats-server/src/update_service.rs
```diff
@@ -27,74 +31,98 @@ fn time_till_next_call(schedule: &Schedule) -> std::time::Duration {
...
         let mut failed = 0;
-        while !updaters.is_empty() {
-            let (res, _, others) = futures::future::select_all(updaters).await;
-            updaters = others;
-            tracing::error!("updater stopped: {:?}", res);
-
+        while let Some(()) = group_update_jobs.next().await {
+            tracing::error!("update job stopped unexpectedly");
             failed += 1;
             if failed >= FAILED_UPDATERS_UNTIL_PANIC {
```
sevenzing, line 107: it was added before but why do we consider 2 stopped updates as fine behaviour? 👀

bragov4ik: my reasoning is as follows:
- if 1-2 chart updaters failed, it means only a small portion of functionality has malfunctioned. it seems logical to keep running the service with other charts in operation.
- if >= 3 fail - it's a more major disruption, so we might as well stop the whole service

sevenzing: but we dont restart updaters. I mean if we will panic here, pod with stats deployment will be redeployed and error will be fixed. But you dont panic here although some of the functionality does not work (small portion but still)

sevenzing (approved):
fix tests and lets merge

PR #1218, opened 2025-01-29
fix(cicd): update tag extractor to support prereleases

sevenzing:
currently we only support tags like `stats/v1.1.1`. we also want to support tags like`stats/v2.4.0-alpha.0` to produce prelease version `v2.4.0-alpha.0`

## Summary by CodeRabbit

- **Chores**
	- Updated Docker build workflow configuration to support more flexible version tag matching.
	- Expanded tag capture logic to allow broader version format parsing, accommodating pre-release identifiers and build metadata.

### DIFF
--- a/.github/workflows/_push_swagger.yml
+++ b/.github/workflows/_push_swagger.yml
@@ -60,12 +60,14 @@ jobs:
 
       - name: Get the current tag, branch, or commit hash
         id: git_info
+        # regex from:
+        # https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
         run: |
           if [ "${{ github.event_name }}" = "workflow_dispatch" ] && [ ! -z "${{ inputs.version }}" ]; then
             echo "version=${{ inputs.version }}" >> $GITHUB_ENV
           else
-            if [[ "${GITHUB_REF}" =~ refs/tags/${{ inputs.service_name }}/v[0-9]+\.[0-9]+\.[0-9]+ ]]; then
-              version=$(echo "${GITHUB_REF}" | grep -o 'v[0-9]\+\.[0-9]\+\.[0-9]\+')
+            if [[ "${GITHUB_REF}" =~ refs/tags/${{ inputs.service_name }}/v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(-((0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9][0-9]*|[0-9]*[a-zA-Z-][0-9a-zA-Z-]*))*))?(\+([0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*))?$ ]]; then
+              version=$(echo "${GITHUB_REF}" | sed -E 's|.*/v||')
               echo "version=${version}" >> $GITHUB_ENV
 
             # If it's the main branch


.github/workflows/_docker-build-push.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         id: regex
         with:
           text: ${{ github.ref }}
-          regex: '^(refs\/tags\/${{ inputs.service-name }}\/(v\d+\.\d+\.\d+))|(refs\/heads\/(main))$'
+          regex: '^(refs\/tags\/${{ inputs.service-name }}\/v([^\/]+))|(refs\/heads\/(main))$'
```
rimrakhimov, line 51: I suggest not to remote the previous `(v\d+\.\d+\.\d+)` constraint but instead to add an optional `(-[^\/]+)?` suffix to support pre-release versions

bragov4ik (changes requested):
It seems that it will still set latest tag for pre release version. I think it's not what we want though?

bragov4ik:
Do we also want to push swagger for pre releases?

.github/workflows/_docker-build-push.yml
sevenzing: we can do even more accurate: add full semver regex 🤡

.github/workflows/_docker-build-push.yml
```diff
@@ -48,14 +48,28 @@ jobs:
         id: regex
         with:
           text: ${{ github.ref }}
-          regex: '^(refs\/tags\/${{ inputs.service-name }}\/(v\d+\.\d+\.\d+))|(refs\/heads\/(main))$'
+          # https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string
+          regex: '^(refs\/tags\/${{ inputs.service-name }}\/v((0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?))|(refs\/heads\/(main))$'
```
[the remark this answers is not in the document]
sevenzing: meh

rimrakhimov (approved):

.github/workflows/_docker-build-push.yml
```diff
@@ -48,14 +48,28 @@ jobs:
...
 
       - name: Extract tag name
         id: tags_extractor
         run: |
-          t=${{ steps.regex.outputs.group2 }}
-          m=${{ steps.regex.outputs.group4 }}
-          (if ! [[ "$t" == "" ]]; then echo tags=${{ env.IMAGE_NAME }}:$t, ${{ env.IMAGE_NAME }}:latest; elif ! [[ "$m" == "" ]]; then echo tags=${{ env.IMAGE_NAME }}:$m; else echo tags=; fi) >> $GITHUB_OUTPUT
+          version=${{ steps.regex.outputs.group2 }}
```
bragov4ik, line 57: probably capture groups also need to be adjusted? now major, minor, and patch have their own separate groups 👀

bragov4ik: not sure how it works though

bragov4ik: > (so cg1 = major, cg2 = minor, cg3 = patch, cg4 = prerelease and cg5 = buildmetadata)

sevenzing: u can check how groups are located here:

https://regex101.com/r/DgoEdc/1

`group2` is whole tag with major, minor, patch, prerelease, buildmetadata

bragov4ik: ah, I mixed up `()` and `[]` 😅 
ok, nice!

sevenzing:
added semver regex to push-swagger job

bragov4ik (approved):
Very nice!

PR #1224, opened 2025-01-31
feat(blockscout-client): update to the latest swagger

rimrakhimov:


bragov4ik (approved):

bragov4ik:
can we also add https://github.com/blockscout/blockscout/issues/11728#event-16194818579 ?
or we should wait until release

rimrakhimov:
@bragov4ik I suggest to at least make them two separate PRs, so that both versions could be used

rimrakhimov:
I will merge this one and could you please regenerate https://github.com/blockscout/blockscout-rs/pull/1235 from the updated main

PR #1230, opened 2025-02-04
fix(stats): ci fix + small other

bragov4ik:


bragov4ik:
ran tests locally, now everything is cool

PR #1231, opened 2025-02-05
feat(stats): dependency table

bragov4ik:


stats/README.md
```diff
@@ -11,6 +11,19 @@ The service consists of 2 parts, a stats calculation library and a transport lay
...
+### Dependecy version requirements
+
+Usually, it's best to use the latest versions of all components. Nevertheless, if there is a need in using older versions, you can refer to the following table.
+
+The table lists versions of Blockscout and User Ops Indexer that each Stats version was developed for and tested on.
+
+| Stats version | Recommended Blockscout version | Recommended User Ops Indexer version |
+|---|---|---|
+| v2.4.0 | v6.10.1 | - (not yet supported) |
+| v2.5.0 | v6.10.1 | - |
```
sevenzing, line 26: let's sort in desc order

stats/README.md
```diff
@@ -11,6 +11,19 @@ The service consists of 2 parts, a stats calculation library and a transport lay
 
 - Postgresql database for this service
 - Access to Blockscout database
+- (Preferable) Access to Blockscout API
```
rimrakhimov, line 14: What does preferable mean and what happens if an access would not be available?

bragov4ik: I've added some lead to more info, can you recheck and see if it's clear now?

bragov4ik: Maybe someone on tech call will come up with a better approach 😇

PR #1255, opened 2025-02-24
feat(launcher): close test db connections

bragov4ik:


libs/blockscout-service-launcher/src/test_database.rs
```diff
@@ -142,6 +142,27 @@ impl TestDbGuard {
...
+        // https://github.com/SeaQL/sea-orm/pull/2511 is available
+        // to work with other connections
+        self.conn_with_db
+            .get_postgres_connection_pool()
+            .close()
+            .await;
+        self.conn_without_db
+            .get_postgres_connection_pool()
+            .close()
+            .await;
```
sevenzing, line 157: never use result of close() ? looks like function always returns (Ok, Ok)

bragov4ik: it was a temporary solution; now that my PR got merged it's the way it was intended

PR #1267, opened 2025-03-04
feat(multi): multichain service optimizations

lok52:
* Remove `RETURNING` clause from entity batch update functions.
* Add an optional DB replica support which is used to separate search queries from intense batch updates, which could slow down user-facing endpoints.
* Optimize `addresses` search query (read code comments for details).

### DIFF
--- a/multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
+++ b/multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
@@ -1,4 +1,3 @@
-use super::paginate_cursor;
 use crate::types::{addresses::Address, ChainId};
 use alloy_primitives::Address as AddressAlloy;
 use entity::{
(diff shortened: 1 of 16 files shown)


multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
```diff
@@ -79,20 +72,45 @@ where
...
+        .all(db)
+        .await?;
 
-    if let Some(page_token) = page_token {
-        c.after((page_token.0.as_slice(), page_token.1));
+    match addresses.get(page_size as usize) {
+        Some(a) => Ok((
+            addresses[..page_size as usize].to_vec(),
+            Some((
+                AddressAlloy::try_from(a.hash.as_slice()).unwrap(),
```
bragov4ik, line 108: is the removed comment about unwrap no longer relevant?

bragov4ik (approved):
Very nice! 👍

multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
lok52: Accidentally deleted. Just restored it.

PR #1288, opened 2025-03-17
feat(multi): use materialized cte for address text search

lok52:


multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
```diff
@@ -55,25 +58,39 @@ pub async fn list<C>(
...
+                })
+                .to_owned(),
+        )
+        .materialized(is_cte_materialized)
+        .table_name(Alias::new("addresses"))
+        .to_owned();
+
+    let base_select = Query::select()
+        .column(ColumnRef::Asterisk)
+        .from(Alias::new("addresses")) // NOTE: this is the CTE reference
```
bragov4ik, line 82: maybe it's more clear to make an alias before `let addresses_cte` and use it in line 77 and here?

bragov4ik, line 82: maybe it's more clear to make an alias before let addresses_cte and use it in line 77 and here?
```suggestion
    // maybe a different name, didn't think a lot about it
    let addresses_cte_alias = Alias::new("addresses");
    let addresses_cte = CommonTableExpression::new()
        .query(
            QuerySelect::query(&mut Entity::find())
                .apply_if(query, |q, query| {
                    let ts_query = prepare_ts_query(&query);
                    q.and_where(Expr::cust_with_expr(
                        "to_tsvector('english', contract_name) @@ to_tsquery($1)",
                        ts_query,
                    ));
                })
                .to_owned(),
        )
        .materialized(is_cte_materialized)
        .table_name(addresses_cte_alias) // is clone needed?
        .to_owned();

    let base_select = Query::select()
        .column(ColumnRef::Asterisk)
        .from(addresses_cte_alias))
```

bragov4ik (approved):

multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
lok52: Agree, fixed

PR #1290, opened 2025-03-18
feat(multi): uniformly distribute addresses for quick search

lok52:
* `addresses` search category now includes only non-token rows.
* Add chain whitelist for quick search.
* Quick search for `addresses` category now returns exactly one result per chain from whitelist.
* Add validation for chain id parameters from external service endpoints.


multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
```diff
@@ -75,21 +68,112 @@ where
...
+                .to_owned(),
+        )
+        .table_name(ranked_addresses_iden.clone())
+        .to_owned();
+
+    let limit = chain_ids.len() as u64;
+    let base_select = Query::select()
+        .column(ColumnRef::Asterisk)
+        .from(ranked_addresses_iden)
+        .and_where(Expr::col(Alias::new("rn")).eq(1))
```
bragov4ik, line 120: is there any determenistic ordering within groups from `PARTITION BY chain_id`? it seems that there should be some rank that will make `rn=1` entries to be the best matching, but I don't see `ORDER BY` that does this.

Maybe I missed something?

bragov4ik: from what I understand, `ORDER BY array_position($4, chain_id) DESC` only sorts the 'groups' themselves

multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
```diff
@@ -75,21 +68,112 @@ where
...
+        .all(db)
+        .await?;
+
+    Ok(addresses)
+}
+
+pub async fn list<C>(
+    db: &C,
+    address: Option<AddressAlloy>,
+    query: Option<String>,
```
bragov4ik, line 147: I got a bit confused about this. From what I gathered, `Some(_)` query will trigger contract search. Is it true?

Maybe name the parameter `contract_name_query` to make it more clear?  (does it make sense?)

multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
```diff
@@ -75,21 +68,112 @@ where
                 .to_owned(),
         )
         .materialized(is_cte_materialized)
-        .table_name(addresses_cte_iden.clone())
+        .table_name(name)
+        .to_owned()
+}
+
+pub async fn uniform_chain_search<C>(
```
bragov4ik, line 75: maybe it's me but it's not obvious: what's the difference between this and `list`?

multichain-aggregator/multichain-aggregator-proto/proto/v1/multichain-aggregator.proto
```diff
@@ -139,7 +139,7 @@ message ListAddressesResponse {
 
 message ListTokensRequest {
   string q = 1;
-  optional string chain_id = 2;
+  repeated string chain_id = 2;
```
bragov4ik, line 142:
```suggestion
  repeated string chain_ids = 2;
```
or is it too late to rename?

lok52: It would require additional PR in contracts info service and a corresponding version release. I guess we can do it later for consistency

multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
lok52: Yes, I guess changing the name will make it more clear

multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
lok52: > is there any determenistic ordering within groups from PARTITION BY chain_id?

There is no ordering for now. I guess we can add at least some `ts_rank` + `primary_key` order to make deterministic and more looking like a best match search.

>from what I understand, ORDER BY array_position($4, chain_id) DESC only sorts the 'groups' themselves

That is true. This is needed only to sort the chains according to their priority.

multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
lok52: Fixed

PR #1309, opened 2025-04-07
chore(blockscout-db): use new sea orm

bragov4ik:
New sea-orm supports generated `Vec<Vec<u8>>` (`bytea[]`)


bragov4ik:
also bump version

PR #1312, opened 2025-04-07
feat(multi): add is_verified_contract field to tokens

lok52:


multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
```diff
@@ -159,6 +159,122 @@ where
     Ok(addresses)
 }
 
+pub async fn get_batch_in_order<C>(
+    db: &C,
+    pks: Vec<(&AddressAlloy, ChainId)>,
+) -> Result<Vec<Option<Model>>, DbErr>
+where
```
sevenzing, line 166: can we simplify this function by changing result type to `HashMap<((Address, ChainId), Model)>`.
using this approach you can remove complicated logic with ordering by `Position`. also you wont return a lot of `None` in Vec if a lot of addresses is missing in database

multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
```diff
@@ -159,6 +159,46 @@ where
...
+        )
+        .all(db)
+        .await?;
+
+    let pk_to_model = models
+        .into_iter()
+        .map(|m| {
+            (
+                (
+                    AddressAlloy::try_from(m.hash.as_slice()).unwrap(),
```
sevenzing, line 191: `.expect(<reason why you think it wont panic>)`

sevenzing (approved):

PR #1325, opened 2025-04-11
feat: add new indexer for operations in Ton Application Chain

r0wdy1:
## Intro

[Ton Application Chain](https://docs.tac.build/) (TAC) is an EVM-compatible extension for the TON blockchain.
TAC transactions are ordered and executed under a separate consensus mechanism, with only the execution results being posted back on-chain — a pattern somewhat similar to how rollups operate.

Users interact with TAC either through dedicated dApps or via smart contracts on the TON chain. Regardless of the method, the user’s intention — along with all resulting artifacts — is encapsulated under a single concept called an “Operation.”

These Operations can involve different data flows and a variety of status types. This complexity can be confusing for inexperienced users and may hinder debugging for dApp developers.

This feature addresses that by indexing TAC Operations in a way that provides a clear, high-level overview of what occurred as a result of a user’s action — making it easier for both users and developers to understand at a glance.

## Improvements 

Indexer Logic Description
The TAC Operation Lifecycle Indexer follows a three-stage process:
1. Timeline Dissection:
* The indexer divides the timeline into fixed-size `intervals`
* It maintains a `watermark` that marks the latest processed timestamp
* The `watermark` advances as `intervals` are processed
* For historical data, it processes `intervals` from oldest to newest
For realtime data, it continuously creates new `intervals`
2. Interval Processing:
* For each interval, the indexer fetches a list of operations that occurred within that time window
* Operations are stored in the database with a `pending` status
* The interval is marked as `finalized` once operations are fetched
* If fetching fails, the interval is scheduled for retry with exponential backoff
3. Operation Processing:
* For each operation, the indexer fetches detailed stage information
* Operation stages track the lifecycle of the operation across different blockchains
* Once stages are fetched, the operation is marked as `finalized`
* If fetching fails, the operation is scheduled for retry with exponential backoff

```
+----------------------------------------------------------------------------------------+
|                                    TAC OPERATION LIFECYCLE INDEXER                      |
+----------------------------------------------------------------------------------------+
                                                                                          
                                                                                          
+-------------------+     +-------------------+     +-------------------+                  
|                   |     |                   |     |                   |                  
|  Timeline         |     |  Intervals        |     |  Operations       |                  
|  Dissection       |---->|  Processing       |---->|  Processing       |                  
|                   |     |                   |     |                   |                  
+-------------------+     +-------------------+     +-------------------+                  
        |                         |                         |                              
        |                         |                         |                              
        v                         v                         v                              
+-------------------+     +-------------------+     +-------------------+                  
|                   |     |                   |     |                   |                  
|  Watermark        |     |  Fetch            |     |  Fetch            |                  
|  Advancement      |     |  Operations       |     |  Operation        |                  
|                   |     |  for Interval     |     |  Stages           |                  
+-------------------+     +-------------------+     +-------------------+  
```

We persist and track latest saved interval (`watermark`) in the database and advance it alongside with creating new intervals.
Apart from latest interval we also track latest `operation` so that if we get a falsely empty response we would automatically request it again.
This PR follows similar practices  from `da_indexer` specifically the server launches multiple future streams:
* historic operation fetcher that selects `intervals` in ascending order from a configurable starting timestamp
* realtime operation fetcher that selects `intervals` in ascending order after the service has started
* failed intervals and operations fetcher resends failed requests with exponential backoff 

```                                                                                       
+----------------------------------------------------------------------------------------+
|                                    PRIORITIZED STREAMS                                  |
+----------------------------------------------------------------------------------------+
                                                                                          
                                                                                          
+-------------------+     +-------------------+     +-------------------+                  
|                   |     |                   |     |                   |                  
|  Realtime         |     |  Historical       |     |  Operations       |                  
|  Stream           |     |  Intervals        |     |  Stream           |                  
|                   |     |  Stream           |     |                   |                  
+-------------------+     +-------------------+     +-------------------+                  
        |                         |                         |                              
        |                         |                         |                              
        v                         v                         v                              
+----------------------------------------------------------------------------------------+
|                                    COMBINED STREAM                                      |
+----------------------------------------------------------------------------------------+
```


tac-operation-lifecycle/README.md
```diff
@@ -0,0 +1,167 @@
...
++-------------------+     +-------------------+     +-------------------+                  
+        |                         |                         |                              
+        |                         |                         |                              
+        v                         v                         v                              
++----------------------------------------------------------------------------------------+
+|                                    COMBINED STREAM                                      |
++----------------------------------------------------------------------------------------+
+```
+
+##Configuration Parameters
```
bragov4ik, line 87:
```suggestion
## Configuration Parameters
```

tac-operation-lifecycle/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Indexer Settings
+
+| Parameter | Type | Default | Description |
```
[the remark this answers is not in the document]
bragov4ik: It also allows to check missing settings from the table in CI

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/settings.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+fn default_catchup_interval() -> time::Duration {
+    time::Duration::from_secs(5)
+}
+
+impl Default for IndexerSettings {
+    fn default() -> Self {
+        Self {
+            concurrency: 2,
```
bragov4ik, line 49: What do you think about defaulting to number of logical CPUs (as it's done in cargo, for example)?

bragov4ik (commented):
just some surface-level comments; will probably look deeper later

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -0,0 +1,1167 @@
...
+                status: Set(EntityStatus::Pending.to_id() as i32),
+                next_retry: Set(None),
+                retry_count: Set(0), // Initialize retry count
+            };
+
+            tracing::debug!(
+                "[Thread {:?}] Attempting to insert operation: {:?}",
+                thread_id,
+                operation_model
+            );
```
bragov4ik, line 326:
```suggestion
            tracing::debug!(
                thread = thread_id,
                "Attempting to insert operation: {:?}",
                operation_model
            );
```
FYI you can do smth like this, or since it's repeated in other logs, maybe put it in a span?

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/lib.rs
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    pub fn interval_stream(
+        &self,
+        direction: OrderDirection,
+        from: Option<u64>,
+        to: Option<u64>,
+    ) -> BoxStream<'_, IndexerJob> {
+        const INTERVALS_QUERY_RESULT_SIZE: usize = 1;
+        const INTERVALS_LOOP_DELAY_INTERVAL_MS: u64 = 100;
```
bragov4ik, line 218: nit: maybe put it at the top of this file to that it's easier to find?

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -0,0 +1,1167 @@
...
+                )
+                .exec(&txn)
+                .await
+            {
+                Ok(_) => tracing::debug!(
+                    "[Thread {:?}] Successfully inserted or skipped operation",
+                    thread_id
+                ),
+                Err(e) => {
+                    tracing::debug!(
```
bragov4ik, line 343: maybe warn?

bragov4ik: or is it not important?

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
bragov4ik: same for other logs

tac-operation-lifecycle/tac-operation-lifecycle-server/tests/startup_works.rs
```diff
@@ -0,0 +1,15 @@
+mod helpers;
+
+use blockscout_service_launcher::test_server;
+use pretty_assertions::assert_eq;
+
+#[tokio::test]
+// #[ignore = "Needs database to run"]
```
bragov4ik, line 7: it needs db, right?

tac-operation-lifecycle/tac-operation-lifecycle-server/tests/helpers/mod.rs
```diff
@@ -0,0 +1,454 @@
+use std::sync::Arc;
+
+use blockscout_service_launcher::{test_database::TestDbGuard, test_server};
+use futures::StreamExt;
+use reqwest::Url;
+use tac_operation_lifecycle_logic::{client::settings::RpcSettings, database::TacDatabase};
+use tac_operation_lifecycle_server::Settings;
+
+pub async fn init_db(test_name: &str) -> TestDbGuard {
+    let db_name = format!("testdb_{test_name}");
```
bragov4ik, line 10: i think you can avoid this step and just put `test_name` into `TestDbGuard::new`. The parameter naming is kinda confusing though, I agree

tac-operation-lifecycle/tac-operation-lifecycle-server/tests/helpers/mod.rs
```diff
@@ -0,0 +1,454 @@
...
+    use sea_orm::Database;
+    use tac_operation_lifecycle_entity::interval;
+    use tac_operation_lifecycle_logic::{
+        client::Client, database::OrderDirection, settings::IndexerSettings, Indexer, IndexerJob,
+    };
+    use tokio::sync::Mutex;
+    use tracing::Instrument;
+
+    #[tokio::test]
+    async fn test_save_intervals() {
```
bragov4ik, line 63: maybe put it into integration tests folder? It doesn't fit `helpers/mod.rs`

bragov4ik: same for other tests in the folder

tac-operation-lifecycle/tac-operation-lifecycle-server/tests/helpers/mod.rs
```diff
@@ -0,0 +1,454 @@
...
+        let db = init_db("save_intervals").await;
+        let conn_with_db = Database::connect(&db.db_url()).await.unwrap();
+
+        let catchup_interval = time::Duration::from_secs(rand::thread_rng().gen_range(1..100));
```
bragov4ik, line 67: let's seed the random to have reproducible tests

bragov4ik: or just set some fixed value

tac-operation-lifecycle/tac-operation-lifecycle-server/tests/helpers/mod.rs
```diff
@@ -0,0 +1,454 @@
...
+        let tasks_number = rand::thread_rng().gen_range(1..100);
+        let lag = tasks_number * catchup_interval.as_secs();
+        let current_epoch = time::SystemTime::now()
+            .duration_since(time::UNIX_EPOCH)
+            .unwrap()
+            .as_secs();
+        let start_timestamp = current_epoch - lag;
+        tracing::debug!("start_timestamp: {}", start_timestamp);
```
bragov4ik, line 75: do these logs display properly? I don't see tracing being initialized before that.
maybe it's initialized by other tests when ran together, but in that case the logs won't work if this test is ran alone.
`let _ = tracing_subscriber::fmt::try_init();`

bragov4ik (changes requested):
I like the readability of the code, it's pretty straightforward to understand 👍 

- Let's fix CI
- Can you please revise the dependencies and update where possible? I understand that it's because service-template is not well maintained, but it's also not nice to have outdated dependencies.
- Let's move statistics to stats service, to avoid pain we have now with migrating stats from Elixir indexing
- There are so many commented out lines. Are they going to be removed in the follow-up PRs? It seems like a WIP service because of them. I suggest to remove them where possible, and convert some into the issues, so it's clear that it's actually planned and not just something forgotten in the code.

btw, if it's actually forgotten stuff, let's make self-reviews for PRs in the future to catch this stuff early on 👀

+all other the minor comments

tac-operation-lifecycle/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+    "tac-operation-lifecycle-logic",
+    "tac-operation-lifecycle-entity",
+    "tac-operation-lifecycle-migration",
+    "tac-operation-lifecycle-proto",
+    "tac-operation-lifecycle-server",
+    "env-docs-generation",
+]
+
+[workspace.dependencies]
+blockscout-service-launcher = "0.17.0"
```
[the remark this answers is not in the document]
EvgenKor: Sure. Updated in https://github.com/blockscout/blockscout-rs/pull/1325/commits/b866db1a8322fabc25d575a0e0ba5744b43a9619
Also I moved all dependencies into the workspace's `Cargo.toml` to make it easier to keep all version updates in sync

tac-operation-lifecycle/.github/workflows/tac-operation-lifecycle.yml
```diff
@@ -0,0 +1,82 @@
+on:
```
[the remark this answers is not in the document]
EvgenKor: Moved in https://github.com/blockscout/blockscout-rs/pull/1325/commits/145fc942106875959226793b5ca55abb45387953

tac-operation-lifecycle/tac-operation-lifecycle-proto/proto/v1/health.proto
```diff
@@ -0,0 +1,63 @@
...
+  }
+  ServingStatus status = 1;
+}
+
+service Health {
+  // If the requested service is unknown, the call will fail with status
+  // NOT_FOUND.
+  rpc Check(HealthCheckRequest) returns (HealthCheckResponse);
+
+  // // Performs a watch for the serving status of the requested service.
```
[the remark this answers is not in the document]
EvgenKor: Removed in https://github.com/blockscout/blockscout-rs/pull/1325/commits/f0478ceca843f9f636500a5215433e491ca5958e

tac-operation-lifecycle/tac-operation-lifecycle-proto/proto/v1/statistic.proto
```diff
@@ -0,0 +1,54 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
bragov4ik: ok if you don't mind breaking changes (i.e. removing endpoints)

tac-operation-lifecycle/Cargo.toml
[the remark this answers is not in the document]
bragov4ik: nice, that's also what we move towards :)

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -0,0 +1,1218 @@
...
+                Ok(_) => {
+                    // Update watermark to the end of the last interval in this batch
+                    // [assume bathes are sorted by `start` field ascending at that point]
+                    let mut updated_watermark: Option<u64> = None;
+                    if let Some(last_interval_from_batch) = chunk.last() {
+                        if let ActiveValue::Set(finish) = &last_interval_from_batch.finish {
+                            updated_watermark = Some(finish.and_utc().timestamp() as u64);
+                            let _ = self
+                                .set_watermark_internal(&tx, updated_watermark.unwrap())
+                                .await;
```
sevenzing, line 273: you dont need unwrap here
```suggestion
                            timestamp = finish.and_utc().timestamp() as u64;
                            let _ = self
                                .set_watermark_internal(&tx, updated_watermark)
                                .await;
                            updated_watermark = Some(timestamp);
```

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -0,0 +1,1218 @@
...
+        let sql = format!(
+            r#"
+            SELECT 
+                o.id as op_id, o.operation_type, o.timestamp, o.status,
+                s.id as stage_id, s.stage_type_id, s.success as stage_success, s.timestamp as stage_timestamp, s.note as stage_note,
+                t.id as tx_id, t.stage_id as tx_stage_id, t.hash as tx_hash, t.blockchain_type as tx_blockchain_type
+            FROM operation o
+            LEFT JOIN operation_stage s ON o.id = s.operation_id
+            LEFT JOIN transaction t ON s.id = t.stage_id
+            WHERE o.id = '{}'
```
sevenzing, line 1052: I saw it several times in code and it looks like sql injection for me, you just insert raw user input into sql. please consider using sql values, for example:
https://github.com/blockscout/blockscout-rs/blob/c1b8c16e1ab2c46d34f379321a3fc9706fb898d1/stats/stats/src/charts/lines/transactions/average_txn_fee.rs#L54-L81

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/lib.rs
```diff
@@ -0,0 +1,631 @@
...
+    settings: IndexerSettings,
+    // This boundary timestamp will used for distinguish between historical and realtime intervals
+    // The main difference lays within interval processing
+    //  - the historical intervals are fetched by selected boundaries
+    //  - the realtime intervals are always fetched FROM the realtime boundary
+    //    (to avoid remote service sync issues)
+    // The boundary can be updated on inserting new operations
+    realtime_boundary: AtomicU64,
+    database: Arc<TacDatabase>,
+}
```
sevenzing, line 112: NIT: consider moving it to sep file, for example `indexer.rs`. it's common practice to use `lib.rs` as place where to define all public interface, like

```
mod ...
mod ...
pub use ...
pub use ...
pub use ...
pub use ...
```

it's ok to put logic inside this file, but only for small libraries. I think it would be better to split this into files

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/interval.rs
```diff
@@ -0,0 +1,21 @@
...
+#[sea_orm(table_name = "interval")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub start: i64,
+    pub end: i64,
+    pub timestamp: i64,
+    pub status: i16,
+    pub next_retry: Option<i64>,
+    pub retry_count: i16,
```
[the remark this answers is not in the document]
EvgenKor: Afaik sea_orm_codegen cannot produce unsigned fields in generated entities. So to make it unsigned we should change it by hand. Is it a good approach?

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/operation_stage.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "operation_stage")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub operation_id: String,
+    pub stage_type_id: i32,
```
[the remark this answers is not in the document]
EvgenKor: `stage_type_id` is a foreign key which linked with `stage_type` table

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/status.rs
```diff
@@ -0,0 +1,33 @@
+//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.0
+
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "status")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub operation_id: i32,
```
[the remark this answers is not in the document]
EvgenKor: This table is an unused artefact and should be removed. Will do it

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/watermark.rs
```diff
@@ -0,0 +1,16 @@
+//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.0
+
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "watermark")]
```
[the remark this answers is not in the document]
EvgenKor: Yes, we thought about this as well, but couldn't find a better name. In this context, `watermark` refers to a [common term](https://en.wikipedia.org/wiki/Watermark_(data_synchronization)) used in stream and batch processing systems — it marks the latest timestamp up to which all data is guaranteed to have been processed (in our case, when all time intervals are created). It helps coordinate time-based progress, retries, and recovery. Do you have any suggestions for a better name?

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/interval.rs
```diff
@@ -0,0 +1,21 @@
...
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "interval")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub start: i64,
+    pub end: i64,
+    pub timestamp: i64,
+    pub status: i16,
```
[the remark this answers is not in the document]
EvgenKor: It was converted to the enum in the several previous commits. The downside was the enum approach requires a few specific and non-obvious tricks in the source code :)

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/transaction.rs
```diff
@@ -0,0 +1,33 @@
+//! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.0
+
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "transaction")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub stage_id: i32,
```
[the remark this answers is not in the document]
EvgenKor: No, it has a different meaning. The `transaction.stage_id` field is FK to the `operation_stage.id` field. The `operation_stage.operation_id` is FK to the `operation.id` field

tac-operation-lifecycle/tac-operation-lifecycle-server/src/main.rs
```diff
@@ -0,0 +1,27 @@
...
+    let db = Arc::new(TacDatabase::new(Arc::new(db_connection)));
+
+    let client = Arc::new(Mutex::new(Client::new(settings.rpc.clone())));
+
+    let mut indexer = Indexer::new(settings.clone().indexer.unwrap(), db.clone()).await?;
+    let realtime_boundary = indexer.get_realtime_boundary();
+
+    tokio::spawn(async move {
+        indexer.start(client).await.unwrap();
+    });
```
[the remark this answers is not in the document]
EvgenKor: Moved in [257501f](https://github.com/blockscout/blockscout-rs/pull/1325/commits/257501f7019f93c45fe86b4bec5a44f93ad3f4f4)

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
EvgenKor: I completely removed thread ID logging — it was just for optimization and doesn't seem needed anymore https://github.com/blockscout/blockscout-rs/pull/1325/commits/c024b40c2dc4f2da193b7e375f6ffc1a4c38633c

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
EvgenKor: Good point, thank you. I divided it into the two cases (inserted or DO NOTHING on conflic)  in https://github.com/blockscout/blockscout-rs/pull/1325/commits/c024b40c2dc4f2da193b7e375f6ffc1a4c38633c

tac-operation-lifecycle/tac-operation-lifecycle-server/tests/helpers/mod.rs
EvgenKor: Random values are the core of this test. We feed pseudo-random values into the debug log, which makes the test reproducible.

tac-operation-lifecycle/tac-operation-lifecycle-proto/proto/v1/api_config_http.yaml
```diff
@@ -0,0 +1,32 @@
...
+    #################### Operations ####################
+
+    - selector: blockscout.tacOperationLifecycle.v1.TacService.GetOperations
+      get: /api/v1/tac/operations/list
+
+    - selector: blockscout.tacOperationLifecycle.v1.TacService.GetOperationDetails
+      get: /api/v1/tac/operation/{operation_id}
+
+    - selector: blockscout.tacOperationLifecycle.v1.TacService.GetOperationByTransaction
+      get: /api/v1/tac/operation/tx/{tx_hash}
```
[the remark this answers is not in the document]
EvgenKor: Endpoints were changed in https://github.com/blockscout/blockscout-rs/pull/1325/commits/4a09ca0b27a775eedf83fc6c533209e5dd23e8fe. I decide to remain `/stat/` module due to upgrading it in the nearest future

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/lib.rs
EvgenKor: Moved in https://github.com/blockscout/blockscout-rs/pull/1325/commits/23f9ce43046c2811437809b5997a36d77ce08fb3

tac-operation-lifecycle/tac-operation-lifecycle-server/src/server.rs
```diff
@@ -0,0 +1,70 @@
...
+    // TODO: add services here
+    health: Arc<HealthService>,
+    stat: Arc<StatisticService>,
+    operations: Arc<OperationsService>,
+}
+
+impl Router {
+    pub fn grpc_router(&self) -> tonic::transport::server::Router {
+        tonic::transport::Server::builder().add_service(HealthServer::from_arc(self.health.clone()))
+        //.add_service(StatisticService::from_arc(self.stat.clone()))
```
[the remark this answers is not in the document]
EvgenKor: https://github.com/blockscout/blockscout-rs/pull/1325/commits/20d24b0ae53d6ed1244de7076443a7f95ef95fac

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -0,0 +1,1167 @@
...
+
+impl TacDatabase {
+    pub fn new(db: Arc<DatabaseConnection>) -> Self {
+        Self { db }
+    }
+
+    // Retrieves the saved watermark from the database (returns 0 if not exist)
+    // NOTE: the method doesn't perform any cachingm just read from the DB!
+    pub async fn get_watermark(&self) -> anyhow::Result<u64> {
+        let existing_watermark = watermark::Entity::find().one(self.db.as_ref()).await?;
```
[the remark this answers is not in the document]
EvgenKor: Fixed here and also in `set_watermark_internal` routine

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -0,0 +1,1167 @@
...
+        }
+
+        Ok(())
+    }
+
+    pub async fn insert_pending_operations(
+        &self,
+        operations: &ApiOperations,
+    ) -> anyhow::Result<()> {
+        let thread_id = thread::current().id();
```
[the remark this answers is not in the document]
EvgenKor: It was for a multithread debugging purposes. It's removed for now

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -0,0 +1,1167 @@
...
+
+        // Saving changes
+        interval_active
+            .update(self.db.as_ref())
+            .await
+            .map_err(|e| {
+                tracing::error!(
+                    "Failed to update interval {} status to {}: {:?}",
+                    interval_model.id,
+                    status.to_string(),
```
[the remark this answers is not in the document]
EvgenKor: Fixed in https://github.com/blockscout/blockscout-rs/pull/1325/commits/1f7ad2f00d2ae3625004d74007ba525e84a3adc5 https://github.com/blockscout/blockscout-rs/pull/1325/commits/f7d77af97b2a10659feeb7390b0173a76a04cf22

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -0,0 +1,1167 @@
...
+                    UPDATE interval 
+                    SET status = {}
+                    WHERE id IN (SELECT id FROM selected_intervals)
+                    RETURNING id, start, "end", timestamp, status, next_retry, retry_count
+                    "#,
+            conditions.join(" AND "),
+            order_clause,
+            limit,
+            update_status.to_id(),
+        )
```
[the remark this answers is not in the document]
EvgenKor: Yes, it looks better for now https://github.com/blockscout/blockscout-rs/pull/1325/commits/087371c3501eef2bdd094792cda2c05e8434d086

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
EvgenKor: Fixed in https://github.com/blockscout/blockscout-rs/pull/1325/commits/03b4b3186f5e0327e72f729bef048bc117a69546

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
EvgenKor: Thanks for catching that! Definitely a potential vulnerability — fixed it here and also in `get_operation_by_tx_hash` https://github.com/blockscout/blockscout-rs/pull/1325/commits/11fb33faab275c8b0b20a79b2efc2073bfd38046

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/interval.rs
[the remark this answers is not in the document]
bragov4ik: ok, no problem)

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/watermark.rs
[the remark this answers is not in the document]
bragov4ik: TIL :)
it's acceptable then, let's not waste time on thinking too much haha

tac-operation-lifecycle/tac-operation-lifecycle-server/tests/helpers/mod.rs
bragov4ik: This will make tests unreliable.
Imagine there is some bug that only triggers on `catchup_interval=10s..20s` and `tasks_number=<15`.

also imagine the pipelines and local runs did not choose this values by randomness.

now, a few releases down the road, the test fails.
now you need to find where the bug was introduced and copypaste these parameters in different commits. it will be very tedious.

also, if there are some special parameter sets that should be tested, these special cases should be written explicitly rather than being left up to the random.

btw, is this test feasible to run on multiple different parameters without re-initializing the db?
maybe make it smth like

```
let cases = [
    (1, 10),
    (10, 1),
    (10, 10),
    (20, 10),
    (100, 10),
    (100, 1),
    (1, 100),
    (100, 100),
];
for (catchup, tasks) in cases {
    run_test(&db);
}
```? only if possible ofc.

Comment (bragov4ik) [reply tac-operation-lifecycle/tac-operation-lifecycle-server/tests/helpers/mod.rs]: at the same time i see your point - it will be possible to reproduce 🤔

File: .github/workflows/tac-operation-lifecycle.yml [hunk 32]
```diff
@@ -0,0 +1,82 @@
...
+    needs:
+      - test
+      - lint
+    if: |
+      always() &&
+      (needs.test.result == 'success' || needs.test.result == 'cancelled') &&
+      (needs.lint.result == 'success' || needs.lint.result == 'cancelled')
+    uses: ./.github/workflows/_docker-build-push.yml
+    with:
+      service-name: tac-operation-lifecycle
```
sevenzing, line 82:
```suggestion
  docker:
    name: Docker build and docker push
    needs:
      - test
      - lint
    if: |
      always() &&
      (needs.test.result == 'success' || needs.test.result == 'cancelled') &&
      (needs.lint.result == 'success' || needs.lint.result == 'cancelled')
    uses: ./.github/workflows/_docker-build-push.yml
    with:
      service-name: tac-operation-lifecycle
    secrets:
      ARM_RUNNER_HOSTNAME: ${{ secrets.ARM_RUNNER_HOSTNAME }}
      ARM_RUNNER_KEY: ${{ secrets.ARM_RUNNER_KEY }}
      

```

sevenzing, line 82:
```suggestion
    with:
      service-name: tac-operation-lifecycle
    secrets:
      ARM_RUNNER_HOSTNAME: ${{ secrets.ARM_RUNNER_HOSTNAME }}
      ARM_RUNNER_KEY: ${{ secrets.ARM_RUNNER_KEY }}
```

tac-operation-lifecycle/tac-operation-lifecycle-entity/src/interval.rs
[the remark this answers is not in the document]
sevenzing: i'm sure that sea orm can generate values for u16 https://www.sea-ql.org/SeaORM/docs/generate-entity/entity-structure/#column-type

but i think we can leave i16 here, no problem

.github/workflows/tac-operation-lifecycle.yml
sevenzing, line 82: also you can add step to push swagger to our repo

just add this job
```

  push-swagger:
    if: github.event_name == 'push' && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags'))
    uses: ./.github/workflows/_push_swagger.yml
    with:
      service_name: 'tac-operation-lifecycle'
      swagger_path: 'tac-operation-lifecycle/tac-operation-lifecycle-proto/swagger/v1/tac-operation-lifecycle.swagger.yaml'
    secrets:
      api_token_github: ${{ secrets.BLOCKSCOUT_BOT_TOKEN }}

```

bragov4ik (approved):
Overall LGTM, nice work!
Also agreed on a call to check alternatives for random tests + waiting to fix CI.

EvgenKor:
I have fixed all the issues mentioned above. CI is now passing successfully (thanks for the help with that). The tests are now deterministic. Please make a quick review in case I missed anything.

EvgenKor:
Also, as was suggested, I have created a few issues to be worked on in the near future
- #1348
- #1349 
- #1350 
- #1351

bragov4ik (approved):
LGTM from my side, thx for organizing the issues

sevenzing (approved):
lgtm, let's merge

PR #1341, opened 2025-04-24
feat(stats): network utilization chart

### LINKED ISSUE #1342: stats: display charts queued for initial update
distinguish between charts whose requirements are not met & those who just wait for their turn

bragov4ik:
network utilization is computed as gas used divided by gas limit.
`average/average` also works because `total=average*count` and count just cancels out.

also added another initial update status state: queued for initial update 
it should exempt such charts from not being shown in charts page (like those whose requirements are not met, e.g. user ops charts when user ops indexer is disabled) (closes #1342)
its scope is only one commit: https://github.com/blockscout/blockscout-rs/pull/1341/commits/1074351f493f0b3ea0b04f64fb1446091e8b278c


bragov4ik:
@coderabbitai review pls

stats/stats/src/charts/lines/gas/average_gas_used.rs
```diff
@@ -0,0 +1,104 @@
...
+define_and_impl_resolution_properties!(
+    define_and_impl: {
+        WeeklyProperties: Week,
+        MonthlyProperties: Month,
+        YearlyProperties: Year,
+    },
+    base_impl: Properties
+);
+
+pub type TotalGasUsed = Delta<GasUsedGrowthInt>;
```
sevenzing, line 51: nit: maybe `GasUsedPerDay` or `NewGasUsed` to show that this is not comulative

sevenzing (approved):

PR #1368, opened 2025-05-17
chore(tac): Tune endpoints

EvgenKor:
To comply with other APIs, the `/api/v1/tac/operations/search/quick?q=string` endpoint was moved to `/api/v1/tac/operations?q=string`. An additional `q` parameter was added to the regular operations list endpoint to support inline filtering by `operation_id`, `tx_hash`, and `sender_address`.

**NOTE1:** _Filtering by sender will be implemented in a neighboring PR, as this field does not exist in the current version of the database._

**NOTE2:** _It's assumed that the global frontend search engine will be use this method as well_

### DIFF
--- a/tac-operation-lifecycle/tac-operation-lifecycle-server/src/services/operations.rs
+++ b/tac-operation-lifecycle/tac-operation-lifecycle-server/src/services/operations.rs
@@ -99,25 +99,75 @@ impl TacService for OperationsService {
     ) -> std::result::Result<tonic::Response<OperationsResponse>, tonic::Status> {
         let inner = request.into_inner();
 
-        match self
-            .db
-            .get_operations(PAGE_SIZE, inner.page_token, OrderDirection::LatestFirst)
-            .await
-        {
-            Ok(operations) => {
-                let last_timestamp = operations
-                    .last()
-                    .map(|op| op.timestamp.and_utc().timestamp() as u64);
+        match inner.q {
+            Some(q) => {
+                // multisearch by the following fields: operation_id, tx_hash, sender
+                if is_generic_hash(&q) {
+                    // operation_id or tx_hash
+                    let map_internal = |e: anyhow::Error| tonic::Status::internal(e.to_string());
 
-                Ok(tonic::Response::new(OperationsResponse {
-                    items: OperationsService::convert_short_db_operation_into_response(operations),
-                    next_page_params: last_timestamp.map(|ts| Pagination {
-                        page_token: ts,
-                        page_items: inner.page_items.unwrap_or(0) as u32 + PAGE_SIZE as u32,
-                    }),
-                }))
+                    let operations = match self
+                        .db
+                        .get_brief_operation_by_id(&q)
+                        .await
+                        .map_err(map_internal)?
+                    {
+                        Some(op) => vec![op],
+                        None => self
+                            .db
+                            .get_brief_operations_by_tx_hash(&q)
+                            .await
+                            .map_err(map_internal)?,
+                    };
+
+                    Ok(tonic::Response::new(OperationsResponse {
+                        items: OperationsService::convert_short_db_operation_into_response(
+                            operations,
+                        ),
+                        next_page_params: None,
+                    }))
+                } else if is_tac_address(&q) || is_ton_address(&q) {
+                    // sender (TON-TAC format)
+                    // TODO: unimplemented for this version of the database.
+                    // The corresponding field doesn't exist for the operation entity.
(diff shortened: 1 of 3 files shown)


bragov4ik:
Hi, what "other APIs" do you mean? Just for the review context

bragov4ik:
By the way, how does it compile with these lines in config not being changed 🤔 
https://github.com/blockscout/blockscout-rs/blob/a3fa46865cf06281d36f5972c257b21713c5936f/tac-operation-lifecycle/tac-operation-lifecycle-proto/proto/v1/api_config_http.yaml#L12-L13

EvgenKor:
> Hi, what "other APIs" do you mean? Just for the review context

E.g `/api/v2/blocks` or `/api/v2/transactions` which support inline filtering

EvgenKor:
> By the way, how does it compile with these lines in config not being changed 🤔
> 
> https://github.com/blockscout/blockscout-rs/blob/a3fa46865cf06281d36f5972c257b21713c5936f/tac-operation-lifecycle/tac-operation-lifecycle-proto/proto/v1/api_config_http.yaml#L12-L13

Oh, it’s my fault. Forgot about it. Will remove it a bit later

tac-operation-lifecycle/tac-operation-lifecycle-server/src/services/operations.rs
```diff
@@ -99,69 +99,75 @@ impl TacService for OperationsService {
...
-
-                Ok(tonic::Response::new(OperationsResponse {
-                    items: OperationsService::convert_short_db_operation_into_response(operations),
-                    next_page_params: last_timestamp.map(|ts| Pagination {
-                        page_token: ts,
-                        page_items: inner.page_items.unwrap_or(0) as u32 + PAGE_SIZE as u32,
-                    }),
-                }))
+        match inner.q {
+            Some(q) => {
```
sevenzing, line 103: move this branch to separate function, such that this main function will be like

```
match inner.q {
   Some(query) => get_operations_with_query(),
   None => get_operations(),
}
```

sevenzing: `impl TacService for OperationsService` should be as short as possible

EvgenKor: Done in https://github.com/blockscout/blockscout-rs/pull/1368/commits/cf7ba5bd167060b9509b6d8b616ede1defa5bdd6 by moving multisearch routine into the logic module (`database.rs`)

EvgenKor:
> By the way, how does it compile with these lines in config not being changed 🤔
> 
> https://github.com/blockscout/blockscout-rs/blob/a3fa46865cf06281d36f5972c257b21713c5936f/tac-operation-lifecycle/tac-operation-lifecycle-proto/proto/v1/api_config_http.yaml#L12-L13

Artefacts were removed in https://github.com/blockscout/blockscout-rs/pull/1368/commits/cf7ba5bd167060b9509b6d8b616ede1defa5bdd6

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -1283,4 +1286,25 @@ impl TacDatabase {
...
+
+            Ok(operations)
+        } else if is_tac_address(q) || is_ton_address(q) {
+            // sender (TON-TAC format)
+            // TODO: unimplemented for this version of the database.
+            // The corresponding field doesn't exist for the operation entity.
+            Ok(vec![])
+        } else {
+            // unknown query string -> return void array without DB interacting
+            Ok(vec![])
```
bragov4ik, line 1307: Does it make sense to return "400 bad request" here? Or maybe it done to comply with other search behaviour?

sevenzing (approved):

bragov4ik (approved):
lgtm

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
sevenzing: I think it's bad to return 400 here. it that case frontend needs some validation of that string and we kinda split this logic between backend and frontend. for search endpoints it's ok to return empty list if string is trash

EvgenKor: I based the behavior on the api/v2/tokens?q= endpoint. I believe that returning an empty vector simply means "Nothing found for this request." The request itself isn’t bad — it just didn’t return any results.

PR #1373, opened 2025-05-20
feat: add a new method to blockscout-chains for fetching a single chain by ID

### LINKED ISSUE #1372: lib/blockscout-chains: add `fetch_one_by_id` method
**Is your feature request related to a problem? Please describe.**
Currently, the `blockscout-chains` library only provides a `BlockscoutChainsClient::fetch_all()` method. This is great for getting a list of all supported chains, but sometimes a user might only need information for a single, specific chain and already knows its ID. Fetching all chains to then filter for one is inefficient in such cases.

The Blockscout Chains API itself supports fetching a single chain via an endpoint like `https://chains.blockscout.com/api/chains/:id` (e.g., `https://chains.blockscout.com/api/chains/1` for Ethereum Mainnet). This library should expose this functionality.

**Describe the solution you'd like**
We should add a new public asynchronous method to `BlockscoutChainsClient` to fetch data for a single chain given its ID.

1.  **New Method:**
    *   Name suggestion: `fetch_one_by_id` (or `fetch_by_id`, `get_chain_by_id`).
    *   It should be an `async` method.
    *   It should take `&self` and a `chain_id: &str` (or appropriate type if the ID is always numeric, but string is safer for API path segments) as parameters.
    *   It should return a `Result<BlockscoutChainData, reqwest_middleware::Error>`. Note that it returns a single `BlockscoutChainData` object, not the `BlockscoutChains` (HashMap) type.

2.  **Implementation Details:**
    *   The method will need to construct the correct URL by appending the `chain_id` to the base URL used by the client (e.g., if `self.url` is `https://chains.blockscout.com/api/chains`, the new URL would be `https://chains.blockscout.com/api/chains/{chain_id}`).
    *   It will use the existing `self.client` (which is a `ClientWithMiddleware`) to make a GET request to this new URL.
    *   The JSON response from the API for a single chain is expected to be a `BlockscoutChainData` object. The method should deserialize the response into this struct.
    *   Error handling should be consistent with `fetch_all()`.

3.  **Testing:**
    *   A new asynchronous test case (e.g., `test_get_specific_blockscout_chain`) must be added in `src/lib.rs` within the `#[cfg(test)] mod tests` module.
    *   This test should:
        *   Create a `BlockscoutChainsClient` instance.
        *   Call the new `fetch_one_by_id` method with a known, valid chain ID (e.g., "1" for Ethereum Mainnet, or "100" for Gnosis Chain - please verify a stable ID from the API: `https://chains.blockscout.com/api/chains`).
        *   Assert that the call is successful (e.g., by using `.unwrap()` or checking the `Result`).
        *   Assert that the returned `BlockscoutChainData` contains expected information (e.g., check the `name` field or another non-volatile field).
    *   The test should follow the pattern of the existing `test_get_blockscout_chains` test.

**Describe alternatives you've considered**
The alternative is to always use `fetch_all()` and then filter the resulting `HashMap` locally. This is less efficient if only one chain's data is needed.

**Additional context**
*   The main client logic is in `libs/blockscout-chains/src/lib.rs`.
*   The `BlockscoutChainsClient` struct and its existing `fetch_all` method are good starting points.
*   The `BlockscoutChainData` struct is the target deserialization type for the single chain data.
*   The `CHAINS_URL` constant defines the base URL for fetching all chains. The new method will need to append to this.
*   Remember that the API endpoint for a single chain (`/api/chains/:id`) returns a single JSON object, not a map/dictionary like `/api/chains`.

This feature would make the `blockscout-chains` library more versatile and efficient for users who need to target specific chains.

akolotov:
Closes #1372

Created with Copilot Agent Mode

### DIFF
--- a/libs/blockscout-chains/Cargo.toml
+++ b/libs/blockscout-chains/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "blockscout-chains"
-version = "0.2.0"
+version = "0.3.0"
 edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
--- a/libs/blockscout-chains/src/lib.rs
+++ b/libs/blockscout-chains/src/lib.rs
@@ -20,6 +20,14 @@ impl BlockscoutChainsClient {
         let chains: BlockscoutChains = res.json().await?;
         Ok(chains)
     }
+
+    /// Fetch a single chain by its ID from the Blockscout Chains API.
+    pub async fn fetch_one_by_id(&self, chain_id: &str) -> Result<BlockscoutChainData, reqwest_middleware::Error> {
+        let url = format!("{}/{}", self.url.trim_end_matches('/'), chain_id);
+        let res = self.client.get(&url).send().await?;
+        let chain: BlockscoutChainData = res.json().await?;
+        Ok(chain)
+    }
 }
 
 impl Default for BlockscoutChainsClient {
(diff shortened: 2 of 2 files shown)


sevenzing:
please bump verions of crate in cargo.toml

PR #1382, opened 2025-05-23
feat(tac): Added support for TAC protocol v3

EvgenKor:
### Several changes were implemented to support the latest TAC protocol (v3) endpoint:

- New data fields are now stored in the database (operation sender, fee info, and valid executors). Currently, only the operation sender is used by the backend; the other fields are stored for future use to avoid a full resync later.

- Search by operation sender has been added. All available TON address formats are supported.

- Pagination logic has been moved to the logic module to improve its behavior. For now, pagination data is returned only if additional rows exist.

- The operation data update routine has been improved: the method that stores the operation state is now included within the update transaction.

**_NOTE:_** It is assumed that the database will be fully resynced manually before deploying the backend. Therefore, there is no point in updating old records with missing data, as this would require the same number of requests as a full resync.


tac-operation-lifecycle/tac-operation-lifecycle-server/src/services/operations.rs
```diff
@@ -98,37 +114,40 @@ impl TacService for OperationsService {
...
-        match inner.q {
-            Some(q) => {
-                let operations = self.db.search_operations(&q).await.map_err(map_db_error)?;
+        let mut input_pagination = None;
+        if let Some(pagination_token) = inner.page_token {
+            input_pagination = Some(LogicPagination {
+                count: inner.page_items.unwrap_or(0) as usize,
+                earlier_timestamp: pagination_token,
+            });
+        }
```
bragov4ik, line 123:
```suggestion
        let input_pagination = inner.page_token.map(|pagination_token|
        LogicPagination {
            count: inner.page_items.unwrap_or(0) as usize,
            earlier_timestamp: pagination_token,
        });
```

also maybe extract it somewhere to keep the endpoint definition as thin as possible

tac-operation-lifecycle/tac-operation-lifecycle-server/src/services/operations.rs
```diff
@@ -98,37 +114,40 @@ impl TacService for OperationsService {
...
-                }))
+                    .map_err(map_db_error)?
             }
-        }
+        };
+
+        let next_page_params = pagination.map(|pag| Pagination {
+            page_token: pag.earlier_timestamp,
+            page_items: pag.count as u32,
+        });
```
bragov4ik, line 145: Ideally make `impl From<LogicPagination> for Pagination`, but it will require adding proto as a dependency to logic.

If it's not convenient, maybe make a function in `server` somewhere to do this conversion

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -1234,24 +1336,50 @@ impl TacDatabase {
...
+        pagination_input: Option<LogicPagination>,
+    ) -> anyhow::Result<(Vec<operation::Model>, Option<LogicPagination>)> {
         let mut query = operation::Entity::find();
 
-        if let Some(ts) = earlier_timestamp {
-            query = query.filter(Column::Timestamp.lt(Self::timestamp_to_naive(ts as i64)));
+        if let Some(pagination) = pagination_input {
+            query = query.filter(Column::Timestamp.lt(Self::timestamp_to_naive(
+                pagination.earlier_timestamp as i64,
+            )));
```
bragov4ik, line 1346: this can be extracted as well

https://github.com/blockscout/blockscout-rs/blob/807f25e7a4343fcf85b225868e302b7817459ce2/blockscout-ens/bens-logic/src/subgraph/sql/utils.rs#L23 
seems pretty convenient to implement something similar

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/indexer.rs
```diff
@@ -532,17 +528,22 @@ impl Indexer {
...
                                 StatusEnum::Pending
                             };
 
-                            let _ = (self
+                            let _ = self
                                 .database
-                                .set_operation_status(&job.operation, &new_status))
-                            .await;
+                                .set_operation_data(&job.operation, operation_data, &new_status)
+                                .await;
```
bragov4ik, line 546: Does it make sense to log the error if it is triggered?

bragov4ik (approved):
LGTM,
minor comments, nothing deal breaking;
can merge after looking into them

tac-operation-lifecycle/tac-operation-lifecycle-server/src/services/operations.rs
EvgenKor: Yes, since the Pagination is autogenerated struct, i decide to make additional routine `OperationsService::convert_logic_pagination` (https://github.com/blockscout/blockscout-rs/pull/1382/commits/f7500bd87a7970c5a048324ac35a8d51ce585a47)

tac-operation-lifecycle/tac-operation-lifecycle-server/src/services/operations.rs
EvgenKor: Moved into the separate routine `extract_input_pagination` in https://github.com/blockscout/blockscout-rs/pull/1382/commits/f7500bd87a7970c5a048324ac35a8d51ce585a47

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
EvgenKor: Not as elegant as in `bens` service (since I work with `Select<>`, which is immutable, unlike `SelectStatement`). But here is my attempt https://github.com/blockscout/blockscout-rs/pull/1382/commits/0eb2e40edb32fa1953288b0bd9591f8fcf7e4384

tac-operation-lifecycle/tac-operation-lifecycle-migration/src/m20250512_135947_add_operation_metainfo.rs
```diff
@@ -0,0 +1,146 @@
...
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Add fields to operation table
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Operation::Table)
+                    .add_column(ColumnDef::new(Operation::SenderAddress).string().null())
+                    .add_column(ColumnDef::new(Operation::SenderBlockchain).string().null())
```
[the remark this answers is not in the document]
EvgenKor: At the beginning of the microservice development, we decided not to add enums to the database for the data received from the remote service. This decision is based on the principle of responsibility boundaries. We introduced enums only for fields that semantically belong to our service — for example, `interval` and `operation` statuses.

Responses from the remote API are validated during the deserialization process (including `sender_blockchain` field). This approach makes it easier to adapt the solution to support other blockchains in the future without needing to rewrite the database layer. Introducing such enums into the database would require significant changes, which seem unnecessary in the current context.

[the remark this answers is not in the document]
bragov4ik: This is reasonable, thanks for explanation

tac-operation-lifecycle/tac-operation-lifecycle-logic/src/database.rs
```diff
@@ -803,77 +771,33 @@ impl TacDatabase {
...
-                    operation_model
-                        .update(tx)
-                        .instrument(tracing::debug_span!(
-                            "updating operation",
-                            tx_id = tx_id.to_string(),
-                        ))
-                        .await?;
+                    Self::remove_operation_associated_stages(tx, &operation.id).await?;
+
+                    // Update operation type, status and sender
```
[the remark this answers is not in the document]
EvgenKor: Really :) https://github.com/blockscout/blockscout-rs/pull/1382/commits/0d6e5704201155d2208819e36f2810bfc6634df5

PR #1407, opened 2025-06-11
feat(multi): add token and coin balance entities

lok52:
## Summary by CodeRabbit

- **New Features**
  - Added support for importing and storing address-level coin and token balances across multiple chains.
  - Extended the batch import API and request formats to accept address coin and token balances.
  - Updated the API documentation and Swagger schema to reflect new balance import fields and data structures.

- **Database Migration**
  - Introduced new database tables for address coin balances and address token balances.

- **Bug Fixes**
  - Improved error handling for decimal parsing in balance imports.


multichain-aggregator/multichain-aggregator-logic/src/services/import.rs
```diff
@@ -36,6 +36,16 @@ pub async fn batch_import(
...
+        .await
+        .inspect_err(|e| {
+            tracing::error!(error = ?e, "failed to upsert address coin balances");
+        })?;
+    repository::address_token_balances::upsert_many(&tx, request.address_token_balances)
+        .await
+        .inspect_err(|e| {
+            tracing::error!(error = ?e, "failed to upsert address token balances");
+        })?;
     tx.commit().await?;
```
bragov4ik, line 49: is it possible that the transaction will just hang if any of errors on the previous lines are triggered?

bragov4ik:
nice primary key work btw 👍

multichain-aggregator/multichain-aggregator-logic/src/services/import.rs
bragov4ik: I use a custom function for this in autoscout:

https://github.com/blockscout/autoscout/blob/f590f9d2386afdb28da74b899e7f211ffb85342d/autoscout/src/logic/utils/transactions.rs#L72

lok52: I guess it should rollback by default if we get an error:
https://www.sea-ql.org/SeaORM/docs/advanced-query/transaction/#begin--commit--rollback

bragov4ik: oh, it's cool! didn't know that they made a workaround for async drop

PR #1409, opened 2025-06-13
feat(blockscout-client): v0.2

sevenzing:
Introduces new approach for code generation of `blockscout-client` crate:

* generate from handlebars template
* use Trait approach instead of functions
* use mockall feature
* use `bon` crate

 see `libs/blockscout-client/crate/tests/it_works.rs` for usage example


sevenzing:
@coderabbitai review

PR #1411, opened 2025-06-16
feat(multi): allow non-whitelisted chains in quick search response

lok52:


multichain-aggregator/multichain-aggregator-logic/src/repository/addresses.rs
```diff
@@ -85,15 +85,11 @@ pub async fn uniform_chain_search<C>(
     db: &C,
     contract_name_query: String,
     token_types: Option<Vec<db_enum::TokenType>>,
-    chain_ids: Vec<ChainId>,
+    limit: u64,
```
bragov4ik, line 88: there's an outside-diff ai's comment "Guard against limit == 0"

makes sense to me, what do you think?

lok52: Agree, thank you

PR #1430, opened 2025-06-30
feat(da-indexer): add s3 storage support

rimrakhimov:
Closes blobs indexing and retrieval part of #1426. The migration of existing blob data from database into s3 storage will be added in a separate PR

## Summary by CodeRabbit

* **New Features**
  * Added optional S3-compatible storage support for blob data, enabling storage and retrieval from S3 or compatible services.
  * Introduced configuration options for S3 storage, including endpoint, credentials, bucket management, and concurrency controls.
  * Enhanced APIs, indexer, and server services to support blob retrieval from both database and S3 storage.

* **Improvements**
  * Updated database schema to allow blobs to store either inline data or an S3 object key with enforced mutual exclusivity.
  * Added Prometheus metrics for monitoring S3 bulk upload operations.
  * Improved error logging formatting for better clarity.

* **Chores**
  * Updated dependencies and test utilities to support S3 integration and schema changes.
  * Refactored tests to validate behavior with and without S3 storage.
  * Added workflow and development environment support for running MinIO as a local S3-compatible service.
  * Simplified server startup by centralizing initialization logic.

### DIFF
--- a/da-indexer/da-indexer-logic/src/common/repository.rs
+++ b/da-indexer/da-indexer-logic/src/common/repository.rs
@@ -1,10 +1,40 @@
-use crate::s3_storage::S3Storage;
+use crate::s3_storage::{S3Object, S3Storage};
 use sea_orm::FromQueryResult;
 
 #[derive(FromQueryResult, Default, Debug)]
 pub struct DbData {
     pub data: Option<Vec<u8>>,
     pub data_s3_object_key: Option<String>,
+}
+
+pub fn convert_blob_data_to_db_data_and_s3_object(
+    maybe_s3_storage: Option<&S3Storage>,
+    da_prefix: &str,
+    blob_id: &[u8],
+    data: Vec<u8>,
+) -> (DbData, Option<S3Object>) {
+    if maybe_s3_storage.is_none() {
+        (
+            DbData {
+                data: Some(data),
+                data_s3_object_key: None,
+            },
+            None,
+        )
+    } else {
+        let object_key = format!("{da_prefix}_{}", hex::encode(blob_id));
+        let s3_object = S3Object {
+            key: object_key.clone(),
+            content: data,
+        };
+        (
+            DbData {
+                data: None,
+                data_s3_object_key: Some(object_key),
+            },
+            Some(s3_object),
+        )
+    }
 }
 
 pub async fn extract_blob_data(
--- a/da-indexer/da-indexer-logic/src/common/repository.rs
+++ b/da-indexer/da-indexer-logic/src/common/repository.rs
@@ -1,0 +1,48 @@
+use crate::s3_storage::S3Storage;
+use sea_orm::FromQueryResult;
+
+#[derive(FromQueryResult, Default, Debug)]
+pub struct DbData {
+    pub data: Option<Vec<u8>>,
+    pub data_s3_object_key: Option<String>,
+}
+
+pub async fn extract_blob_data(
+    s3_storage: Option<&S3Storage>,
+    blob_id: &[u8],
+    db_data: DbData,
+) -> Result<Vec<u8>, anyhow::Error> {
+    if let Some(data) = db_data.data {
+        return Ok(data);
+    }
+
+    let data_s3_object_key = db_data.data_s3_object_key.ok_or_else(|| {
+        // Must never be the case as we added a constraint on the postgres table
+        // ensuring exactly one of `data` or `data_s3_object_key` fields is not null.
+        tracing::error!(
+            blob_id = hex::encode(blob_id),
+            "both blob data and object key are missing from the database. \
+            The database is corrupted!"
+        );
+        anyhow::anyhow!("both blob data and object key are missing from the database")
+    })?;
+
+    let s3_storage = s3_storage.ok_or_else(|| {
+        tracing::error!(
+            blob_id = hex::encode(blob_id),
+            "blob data is located in s3 storage, but the storage is not available. \
+                Please check your configuration."
+        );
+        anyhow::anyhow!("blob data was stored in s3 storage but the storage is not available")
+    })?;
+
+    let object = s3_storage.find_object_by_key(&data_s3_object_key).await?;
+    object.map(|object| object.content).ok_or_else(|| {
+        tracing::error!(
+            blob_id = hex::encode(blob_id),
+            data_s3_object_key = data_s3_object_key,
+            "blob data is missing from s3 storage"
+        );
+        anyhow::anyhow!("blob data is missing from s3 storage")
+    })
+}
--- a/da-indexer/da-indexer-logic/src/common/repository.rs
+++ b/da-indexer/da-indexer-logic/src/common/repository.rs
@@ -13,28 +13,25 @@ pub fn convert_blob_data_to_db_data_and_s3_object(
     blob_id: &[u8],
     data: Vec<u8>,
 ) -> (DbData, Option<S3Object>) {
-    if maybe_s3_storage.is_none() {
-        (
-            DbData {
-                data: Some(data),
-                data_s3_object_key: None,
-            },
-            None,
-        )
-    } else {
-        let object_key = format!("{da_prefix}_{}", hex::encode(blob_id));
-        let s3_object = S3Object {
-            key: object_key.clone(),
-            content: data,
-        };
-        (
-            DbData {
-                data: None,
-                data_s3_object_key: Some(object_key),
-            },
-            Some(s3_object),
-        )
-    }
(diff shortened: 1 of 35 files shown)


da-indexer/da-indexer-migration/src/m20250623_122642_alter_blob_tables_add_data_s3_object_key_column.rs
```diff
@@ -0,0 +1,40 @@
...
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let sql = r#"
+            ALTER TABLE "celestia_blobs"
+                ALTER COLUMN "data" DROP NOT NULL,
+                ADD COLUMN "data_s3_object_key" varchar,
+                ADD CONSTRAINT "data_integrity"
+                    CHECK ("data" IS NOT NULL AND "data_s3_object_key" IS NULL 
+                        OR "data" IS NULL AND "data_s3_object_key" IS NOT NULL);
```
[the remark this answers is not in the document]
rimrakhimov: This is not true, and the check constraints should be unique only within the object they are defined on

da-indexer/da-indexer-logic/src/s3_storage.rs
```diff
@@ -0,0 +1,197 @@
...
+                        return Ok(None);
+                    }
+                }
+                Err(error).context("get object by key failed")
+            }
+        }
+    }
+
+    pub async fn insert_many(&self, objects: Vec<S3Object>) -> Result<(), anyhow::Error> {
+        metrics::S3_BULK_UPLOAD_TOTAL.inc();
```
bragov4ik, line 106: nit: does it make sense to have a metric of total uploads (for each separate item, ignoring bulks)?

da-indexer/da-indexer-logic/src/s3_storage.rs
```diff
@@ -0,0 +1,197 @@
...
+                        .put_object()
+                        .bucket(&self.bucket)
+                        .key(&object.key)
+                        .body(body)
+                        .content_md5(BASE64_STANDARD.encode(content_md5))
+                        .send()
+                        .await
+                        .context(format!("put object {} into s3 storage failed", object.key))?;
+
+                    // Have to always be false, as the integrity should already be validated
```
bragov4ik: I'm trying to understand this part; what does "false" mean? Maybe I miss something

bragov4ik: does it mean that the validation should always fail?

da-indexer/da-indexer-logic/src/common/tests.rs
```diff
@@ -0,0 +1,149 @@
...
+        is_bucket_empty(
+            &client_with_details.client,
+            &client_with_details.bucket_name,
+        )
+        .await
+    }
+
+    async fn initialize_client(test_name: &str) -> ClientWithDetails {
+        let endpoint =
+            std::env::var("S3_ENDPOINT").unwrap_or_else(|_| "http://127.0.0.1:9000".to_string());
```
bragov4ik, line 53: Do tests using it work if `minio` is not running locally?
If not, let's maybe add installation instructions to readme and relevant `justfile` commands?

da-indexer/da-indexer-migration/src/m20250623_122642_alter_blob_tables_add_data_s3_object_key_column.rs
```diff
@@ -0,0 +1,40 @@
...
+        "#;
+        crate::from_sql(manager, sql).await
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let sql = r#"
+            ALTER TABLE "celestia_blobs"
+                DROP COLUMN "data_s3_object_key",
```
bragov4ik, line 30: Have you checked if constraint is dropped automatically?

da-indexer/da-indexer-logic/src/common/tests.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    struct ClientWithDetails {
+        endpoint: String,
+        access_key_id: String,
+        secret_access_key: String,
+        bucket_name: String,
+        client: s3::Client,
+    }
+
+    pub async fn initialize_s3_storage(test_name: &str) -> S3Storage {
```
bragov4ik, line 21: nit: maybe return `Client` here as well to avoid initializing it again on `is_s3_storage_empty` calls

da-indexer/da-indexer-logic/src/common/tests.rs
```diff
@@ -0,0 +1,149 @@
...
+        clear_bucket(client, bucket_name).await;
+        client
+            .delete_bucket()
+            .bucket(bucket_name)
+            .send()
+            .await
+            .expect("failed to delete existing bucket");
+    }
+
+    async fn clear_bucket(client: &s3::Client, bucket_name: &str) {
```
bragov4ik: nit: maybe `empty_bucket` for consistent naming
https://docs.aws.amazon.com/AmazonS3/latest/userguide/empty-bucket.html

da-indexer/da-indexer-logic/src/common/repository.rs
```diff
@@ -0,0 +1,78 @@
...
+        };
+        (
+            DbData {
+                data: None,
+                data_s3_object_key: Some(object_key),
+            },
+            Some(s3_object),
+        )
+    }
+}
```
lok52: Nit: maybe pattern match on `maybe_s3_storage` would look more concise?
```rust
let (data, data_s3_object_key, s3_object) = match maybe_s3_storage {
    ...
};
(DbData {data, data_s3_object_key}, s3_object)
```

da-indexer/da-indexer-logic/src/eigenda/repository/blobs.rs
```diff
@@ -70,6 +97,11 @@ pub async fn upsert_many<C: ConnectionTrait>(
         .on_empty_do_nothing()
         .exec(db)
         .await?;
+
+    if let Some(s3_storage) = s3_storage {
+        s3_storage.insert_many(data_s3_objects).await?;
+    }
```
lok52: Does it make sense to run db batch insert and s3 upload in parallel? Something like `join!` can be used

lok52 (approved):

da-indexer/da-indexer-logic/src/s3_storage.rs
rimrakhimov: Actually, I don't think it make sense for us to measure the total number of files uploaded. The idea behind counting bulks is to obtain the ratio between total and failed bulk uploads which may help to adjust `save_concurrency_limit` parameter. But as for the total number of files, I don't see it may help us in any way. Besides, we may always get that info from the bucket

da-indexer/da-indexer-logic/src/s3_storage.rs
rimrakhimov: That should always be valid, thanks. Fixed in 4eac371

da-indexer/da-indexer-migration/src/m20250623_122642_alter_blob_tables_add_data_s3_object_key_column.rs
rimrakhimov: Yeap, I'm pretty sure the constraint name is associated with the column

da-indexer/da-indexer-logic/src/common/tests.rs
rimrakhimov: Added in 1efdc42

da-indexer/da-indexer-logic/src/common/tests.rs
rimrakhimov: I believe, I prefer to use the current initialization logic as for there is only one such duplication. Will update it later if needed

da-indexer/da-indexer-logic/src/common/tests.rs
rimrakhimov: Fixed in e6cdbe3

da-indexer/da-indexer-logic/src/common/repository.rs
rimrakhimov: Updated in 4cde0db

da-indexer/da-indexer-logic/src/eigenda/repository/blobs.rs
rimrakhimov: Updated in 3c310e1a

PR #1433, opened 2025-07-02
feat(libs): add recache

lok52:


libs/recache/src/handler.rs
```diff
@@ -0,0 +1,216 @@
...
+use dashmap::DashMap;
+use futures::{
+    FutureExt,
+    future::{BoxFuture, Shared},
+};
+use std::{cmp::Eq, fmt::Display, hash::Hash, sync::Arc, time::Duration};
+use thiserror::Error;
+
+type SharedFuture<V, E> = Shared<BoxFuture<'static, Result<V, E>>>;
+type EventHandler<V> = Arc<dyn Fn(&V) + Send + Sync>;
```
rimrakhimov: May the key also be useful during event handling?

libs/recache/src/stores/redis.rs
```diff
@@ -0,0 +1,128 @@
...
+                let ttl_duration = if ttl >= 0 {
+                    Some(Duration::from_millis(ttl as u64))
+                } else {
+                    None
+                };
+                Ok(Some((value, ttl_duration)))
+            }
+            None => Ok(None),
+        }
+    }
```
rimrakhimov: Nit: I suggest to extract value option check and deserialization into a separate function to share it between get and get_with_ttl

libs/recache/src/handler.rs
```diff
@@ -0,0 +1,216 @@
...
+    where
+        Self: Sized,
+    {
+        CacheRequest::builder(
+            self.cache.clone(),
+            self.inflight.clone(),
+            self.on_hit.clone(),
+            self.on_computed.clone(),
+            self.on_refresh_computed.clone(),
+        )
```
rimrakhimov: I probably missed that, but where are `default_ttl` and `default_refresh_ahead` values are set for the request in case no special values has been provided?

libs/recache/src/handler.rs
```diff
@@ -0,0 +1,216 @@
...
+    C: AsyncCacheStore<K, V>,
+    K: Sync + Hash + Eq,
+{
+    #[builder(start_fn)]
+    cache: Arc<C>,
+    #[builder(default)]
+    inflight: Arc<InflightMap<K, V, C>>,
+    on_hit: Option<EventHandler<V>>,
+    on_computed: Option<EventHandler<V>>,
+    on_refresh_computed: Option<EventHandler<V>>,
```
rimrakhimov: Nit: what do you think about having some no-op implementation set by default instead of `Option`s? That allows to get rid of `if let Some(handler) = on_handler` lines and let the default no-op implementation handle the None case

libs/recache/src/handler.rs
```diff
@@ -0,0 +1,216 @@
...
+                on_computed(&val);
+            }
+            Ok(val)
+        }
+        .boxed()
+        .shared();
+
+        self.inflight.insert(self.key.clone(), fut_shared.clone());
+        let result = fut_shared.await;
+        self.inflight.remove(&self.key);
```
rimrakhimov: Imo, it should be only the future that created the thread responsible for removing the key from the map. Otherwise, though quite improbable, it may be possible that a thread will remove the entry for a future added already after its original future returned.

rimrakhimov:
I believe there is a race condition in the `execute` method which may result in 2 threads entering the `compute`

Consider the following case: 
1. Two threads. Both executed lines 111 (`self.cache.get_with_ttl(&self.key).await?`) and 122 (`self.inflight.get(&self.key)`) and did not find anything.
2. Thread 1 created a new compute future and inserted it into the inflight map (line 145 - `self.inflight.insert(self.key.clone(), fut_shared.clone());`)
3. Thread 2 also created a new compute function, replaced first thread's value and continued with computation. There are 2 threads executing computation for the same key now

libs/recache/src/handler.rs
lok52: Agree, added the key

libs/recache/src/handler.rs
lok52: That looks nice, I guess the compiler can optimize such noop function calls anyway

libs/recache/src/handler.rs
lok52: These are stored only to be accessed by a cache request instance later. Example: https://github.com/blockscout/blockscout-rs/blob/366e2f36fec6426b65be93dc11097ea411b1c849/multichain-aggregator/multichain-aggregator-logic/src/services/search.rs#L604-L605

lok52: This is a kind of a workaround of how Bon handles fields internally. If we set a default ttl during request initialization, then we won't be able to overwrite it again. There is an experimental feature to allow overwrites but I decided to not use it - https://bon-rs.com/reference/builder/member/overwritable

lok52: I added a `default_request` abstraction in https://github.com/blockscout/blockscout-rs/pull/1433/commits/fa47f8ff7f1d521e39ce994a114da373ba5d68ee. I guess that should simplify handling default settings

libs/recache/src/stores/redis.rs
lok52: Agree, thank you

libs/recache/src/handler.rs
lok52: I refactored the code a bit to prevent potential race conditions. Now, shared futures incapsulating the primary `compute` logic are not responsible for cache/inflight map updates. Instead, these shared futures are wrapped in another future providing callback-like behaviour. For example, refresh-ahead future looks like this: https://github.com/blockscout/blockscout-rs/blob/9a5406e4540d9514c363093d0b7986651b616add/libs/recache/src/handler.rs#L198-L205
And primary compute future created in case cache and inflight map are empty looks like this:
https://github.com/blockscout/blockscout-rs/blob/9a5406e4540d9514c363093d0b7986651b616add/libs/recache/src/handler.rs#L158-L165
And simple inflight map subscription future is the following:
https://github.com/blockscout/blockscout-rs/blob/9a5406e4540d9514c363093d0b7986651b616add/libs/recache/src/handler.rs#L145-L150

As you can see, in first two cases we have `insert` statements before creating such futures, so they are responsible to remove it later. I guess this should solve the problem

libs/recache/src/handler.rs
```diff
@@ -0,0 +1,241 @@
...
+            }
+            dashmap::Entry::Vacant(entry) => {
+                let fut_shared = Self::build_shared_future(compute());
+
+                entry.insert(fut_shared.clone());
+
+                let inflight = Arc::clone(&self.inflight);
+                async move {
+                    let res = fut_shared.await;
+                    if let Ok(val) = res.as_ref() {
```
rimrakhimov, line 160: Nit: in commit message it was said that the `as_ref` was added to prevent potential memory leaks. Just interested, why there could be the leak if taken by value?

libs/recache/src/handler.rs
rimrakhimov: Get it! Yeap, I see the logic behind requiring the value to be set only once, though in our case it is a little inconvenient. The workaround seems good, imo

libs/recache/src/handler.rs
lok52: `as_ref` was used just to emphasize that we don't need to consume the `val`. The fix itself was only about avoiding early return with `?`, so that we can always reach `inflight.remove(&self.key)` even if internal future returns an error. Also, I think there still could be a memory leak if `fut_shared` panics and we will not remove it from the map, but I guess we can just assume that users should provide panic-free functions.

rimrakhimov: Get it! Agree, that sounds fine

sevenzing (approved):

PR #1435, opened 2025-07-02
feat(libs): add auth logic for user

Ilyak777:
Added new way to authorize user using `/api/account/v2/user/info` route

### DIFF
--- a/libs/blockscout-auth/src/auth.rs
+++ b/libs/blockscout-auth/src/auth.rs
@@ -1,4 +1,4 @@
-use crate::jwt_headers::{build_http_headers, extract_csrf_token, extract_jwt, CommonError};
+use crate::jwt_headers::{build_http_headers, extract_csrf_token, extract_jwt};
 use reqwest::{Client, StatusCode};
 use serde::Deserialize;
 use thiserror::Error;
@@ -26,9 +26,6 @@ struct AuthFailed {
 
 #[derive(Error, Debug)]
 pub enum Error {
-    #[error(transparent)]
-    Common(#[from] CommonError),
-
     #[error("blockscout API error: {0}")]
     BlockscoutApi(String),
 
@@ -37,6 +34,15 @@ pub enum Error {
 
     #[error("forbidden: {0}")]
     Forbidden(String),
+
+    #[error("invalid jwt token: {0}")]
+    InvalidJwt(String),
+
+    #[error("invalid csrf token: {0}")]
+    InvalidCsrfToken(String),
+
+    #[error("cannot build headers: {0}")]
+    HeaderError(String),
 }
 
 pub async fn auth_from_metadata(
@@ -106,8 +112,14 @@ pub async fn auth_from_tokens(
             Ok(success)
         }
         StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN => {
-            let failed: AuthFailed = serde_json::from_str(&response_raw)
-                .map_err(|e| Error::BlockscoutApi(e.to_string()))?;
+            let failed: AuthFailed = serde_json::from_str(&response_raw).map_err(|error| {
+                tracing::warn!(
+                    error = ?error,
+                    body = ?response_raw,
+                    "failed to parse blockscout response body"
+                );
+                Error::BlockscoutApi(format!("invalid body: {error}"))
+            })?;
             if status == StatusCode::UNAUTHORIZED {
                 Err(Error::Unauthorized(failed.message))
             } else {
@@ -123,7 +135,9 @@ mod tests {
     use std::str::FromStr;
 
     use super::*;
-    use crate::{init_mocked_blockscout_auth_service, MockUser};
+    use crate::{
+        init_mocked_blockscout_auth_service, jwt_headers::HEADER_JWT_TOKEN_NAME, MockUser,
+    };
     use reqwest::header::{HeaderMap, HeaderName};
     use serde::Serialize;
     use tonic::{codegen::http::header::CONTENT_TYPE, Extensions, Request};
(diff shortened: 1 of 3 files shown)


libs/blockscout-auth/src/user.rs
```diff
@@ -0,0 +1,85 @@
...
+use serde::Deserialize;
+use thiserror::Error;
+use tonic::metadata::MetadataMap;
+use url::Url;
+
+const API_KEY_NAME: &str = "api_key";
+
+#[derive(Debug, Deserialize)]
+pub struct UserInfo {
+    pub address_hash: String,
```
sevenzing: are you sure `address_hash` is always present in response?
image

libs/blockscout-auth/src/user.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub nickname: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct ErrorBody {
+    message: String,
+}
+
+#[derive(Error, Debug)]
+pub enum Error {
```
sevenzing: why do you need second enum Error? this is like the same thing as auth::Error. Also you dont export this Error in lib.rs so it's not possible to import it from the crate.

sevenzing: this is small crate, there is no need to create 2 errors in move "common" into CommonError. also Common is bad naming because it can mean anything

libs/blockscout-auth/src/lib.rs
```diff
@@ -1,5 +1,9 @@
 mod auth;
+mod common;
```
sevenzing: common is bad name for module. if you want to create file with tools to handle headers, cookies or jwt, you can call it `headers.rs`, `parse_headers.rs` or ask gpt about good naming

libs/blockscout-auth/src/auth.rs
```diff
@@ -70,128 +71,48 @@ pub async fn auth_from_tokens(
...
     match status {
         StatusCode::OK => {
-            let success: AuthSuccess = serde_json::from_str(&response_raw).map_err(|error| {
-                tracing::warn!(
-                    error = ?error,
-                    body = ?response_raw,
-                    "failed to parse blockscout response body"
-                );
-                Error::BlockscoutApi(format!("invalid body: {error}"))
-            })?;
```
sevenzing: why..? is it bad warning?

libs/blockscout-auth/src/auth.rs
```diff
@@ -70,128 +71,48 @@ pub async fn auth_from_tokens(
...
-            .map(|api_key| format!("{API_KEY_NAME}={api_key}"))
+            .map(|k| format!("{API_KEY_NAME}={k}"))
             .as_deref(),
     );
+
     let headers = build_http_headers(jwt, csrf_token)?;
-    let client = reqwest::Client::new();
-    let response = if csrf_token.is_some() {
+    let client = Client::new();
+    let resp = if csrf_token.is_some() {
```
sevenzing: why do you want to replace response with resp? any good reason for that?

libs/blockscout-auth/src/auth.rs
```diff
@@ -70,128 +71,48 @@ pub async fn auth_from_tokens(
 ) -> Result<AuthSuccess, Error> {
     let mut url = blockscout_host
         .join("/api/account/v2/authenticate")
-        .expect("should be valid url");
+        .expect("invalid base url");
     url.set_query(
         blockscout_api_key
-            .map(|api_key| format!("{API_KEY_NAME}={api_key}"))
```
sevenzing, line 76: why do you want to change `api_key`? any good reason for that?

libs/blockscout-auth/src/common.rs
```diff
@@ -0,0 +1,71 @@
...
+        .and_then(|v| v.to_str().ok())
+        .map(|s| s.to_string())
+        .ok_or_else(|| AuthError::InvalidCsrfToken("csrf not found".into()))
+}
+
+fn get_cookies(
+    metadata: &MetadataMap,
+) -> Result<std::collections::HashMap<String, Cookie>, AuthError> {
+    let raw = metadata
+        .get("cookie")
```
sevenzing: let's move "cookie" string into const as we did for `HEADER_JWT_TOKEN_NAME` for example. let's call it `HEADER_COOKIE_NAME`

libs/blockscout-auth/src/jwt_headers.rs
```diff
@@ -0,0 +1,72 @@
...
+
+#[derive(Error, Debug)]
+pub enum AuthError {
+    #[error("invalid jwt token: {0}")]
+    InvalidJwt(String),
+    #[error("invalid csrf token: {0}")]
+    InvalidCsrfToken(String),
+    #[error("cannot build headers: {0}")]
+    HeaderError(String),
+}
```
sevenzing: move to `auth::Error`

libs/blockscout-auth/src/auth.rs
```diff
@@ -107,91 +106,16 @@ pub async fn auth_from_tokens(
             Ok(success)
         }
         StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN => {
-            let failed: AuthFalied = serde_json::from_str(&response_raw).map_err(|error| {
-                tracing::warn!(
-                    error = ?error,
-                    body = ?response_raw,
-                    "failed to parse blockscout response body"
-                );
-                Error::BlockscoutApi(format!("invalid body: {error}"))
```
sevenzing: why?

libs/blockscout-auth/src/auth.rs
```diff
@@ -216,6 +141,7 @@ mod tests {
             );
             headers.insert(COOKIE, cookies.parse().unwrap());
         } else {
+            const HEADER_JWT_TOKEN_NAME: &str = "authorization";
```
sevenzing: import it from src/jwt_headers.rs

sevenzing (approved):

PR #1443, opened 2025-07-15
feat(multi): add token entity

lok52:
## Summary by CodeRabbit

* **New Features**
  * Added support for importing and managing tokens, including metadata, price data, and counters, via new API fields and database structures.
  * Introduced a new token type: ERC-7802.
  * Extended batch import functionality to handle token data and interop message-related token updates.
  * Added efficient batch update operations for token data to improve database performance.
  * Updated API documentation and protobuf definitions to include token import schemas and the new token type.

* **Bug Fixes**
  * Improved error handling and parsing for token-related fields.

* **Documentation**
  * Enhanced Swagger and protobuf documentation to reflect new token import capabilities and enum values.

* **Chores**
  * Updated dependencies and internal macros for improved consistency and maintainability.
  * Refactored test server initialization for better usability in tests.


multichain-aggregator/multichain-aggregator-entity/src/tokens.rs
```diff
@@ -0,0 +1,53 @@
...
+        auto_increment = false,
+        column_type = "VarBinary(StringLen::None)"
+    )]
+    pub address_hash: Vec<u8>,
+    #[sea_orm(column_type = "Text", nullable)]
+    pub icon_url: Option<String>,
+    pub fiat_value: Option<Decimal>,
+    pub circulating_market_cap: Option<Decimal>,
+    #[sea_orm(column_type = "Decimal(Some((78, 0)))", nullable)]
+    pub total_supply: Option<BigDecimal>,
```
EvgenKor, line 28: I tried to regenerate the entities and got `Decimal` here instead of `BigDecimal` (and in several other places as well). Did you change it manually? If so, is there a way to automate this change during entity generation? Should we add a script for that to the justfile, for example?

EvgenKor: Here is AI-generated script. I check it briefly. Maybe using something like that after double checking and adjusting.

```(bash)
#!/bin/bash

find . -type f -name "*.rs" | while read -r file; do
  grep -n '#\[sea_orm(column_type = "Decimal(Some' "$file" | cut -d: -f1 | while read -r lineno; do
    nextline=$((lineno + 1))
    # Print the original line for logging
    orig_line=$(sed -n "${nextline}p" "$file")
    if echo "$orig_line" | grep -q '\bDecimal\b'; then
      echo "File: $file, Line: $nextline"
      echo "Before: $orig_line"
      # Use Perl for in-place replacement with word boundary
      perl -i -pe "s/\\bDecimal\\b/BigDecimal/g if \$. == $nextline" "$file"
      # Print the new line for logging
      new_line=$(sed -n "${nextline}p" "$file")
      echo "After:  $new_line"
      echo "-----------------------------"
    fi
  done
done
```

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
```diff
@@ -0,0 +1,203 @@
...
+async fn update_token_price_data<C>(
+    db: &C,
+    mut updates: Vec<UpdateTokenPriceData>,
+) -> Result<(), DbErr>
+where
+    C: ConnectionTrait + TransactionTrait,
+{
+    updates.sort_by(|a, b| (&a.address_hash, a.chain_id).cmp(&(&b.address_hash, b.chain_id)));
+    let active_models = updates.into_iter().map(|m| m.into_active_model());
+    batch_update(
```
EvgenKor: I'm not completely sure, but it seems to me that this code will update the database field even if the incoming data contains NULL. Unlike the token metadata update, you're not using `COALESCE` or `IS DISTINCT FROM` in this function. But in our API, we might receive only `FiatValue` without an update for `CirculatingMarketCap`, for example. Could you double-check that everything is working correctly here?

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
```diff
@@ -0,0 +1,203 @@
...
+async fn update_token_counters<C>(
+    db: &C,
+    mut updates: Vec<UpdateTokenCounters>,
+) -> Result<(), DbErr>
+where
+    C: ConnectionTrait + TransactionTrait,
+{
+    updates.sort_by(|a, b| (&a.address_hash, a.chain_id).cmp(&(&b.address_hash, b.chain_id)));
+    let active_models = updates.into_iter().map(|m| m.into_active_model());
+    batch_update(
```
EvgenKor: and here

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
```diff
@@ -0,0 +1,203 @@
...
+        db,
+        active_models,
+        vec![Column::TransfersCount, Column::HoldersCount],
+    )
+    .await?;
+
+    Ok(())
+}
+
+async fn batch_update<C, A>(
```
sevenzing: good job! 

nit: if it's not very hard, I would move logic of creating `query` into separate function and add unit tests for to should how it actually will construct queries

multichain-aggregator/multichain-aggregator-migration/src/m20250723_084105_add_tokens.rs
```diff
@@ -0,0 +1,45 @@
...
+            ALTER TYPE token_type ADD VALUE IF NOT EXISTS 'ERC-7802';
+
+            CREATE TABLE tokens (
+                chain_id bigint NOT NULL REFERENCES chains (id),
+
+                name text,
+                symbol text,
+                decimals smallint,
+                token_type token_type NOT NULL,
+                address_hash bytea NOT NULL,
```
EvgenKor: minor remark: let's move this string upper at line 14 for clarification (table definition is usually started with PK(s))

multichain-aggregator/multichain-aggregator-entity/src/tokens.rs
lok52: > Did you change it manually?

Yes, sea-orm-cli cannot infer `BigDecimal` directly, so we have to adjust the type manually

multichain-aggregator/multichain-aggregator-migration/src/m20250723_084105_add_tokens.rs
lok52: Updated

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
lok52: Updated

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
lok52: Agree. Updated in https://github.com/blockscout/blockscout-rs/pull/1443/commits/34a659d849186947df08d8426a9cb9cca4c58201

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
lok52: Updated in https://github.com/blockscout/blockscout-rs/pull/1443/commits/34a659d849186947df08d8426a9cb9cca4c58201

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
```diff
@@ -0,0 +1,120 @@
...
+    let mut metadata_updates = Vec::new();
+    let mut price_updates = Vec::new();
+    let mut counter_updates = Vec::new();
+
+    for token in tokens {
+        match token {
+            TokenUpdate::Metadata(metadata) => metadata_updates.push(metadata),
+            TokenUpdate::PriceData(price_data) => price_updates.push(price_data),
+            TokenUpdate::Counters(counters) => counter_updates.push(counters),
+        }
```
EvgenKor: This code is incorrect in the current context. Only a single push will be made per token. I tried importing token metadata using the following request, and it inserted only the metadata (without price data and counters). To import price data, I had to remove the `metadata` field from the request and so on.

```(json)
{
  "chain_id": "10",
  "tokens": [
    {
      "address_hash": "0x94b008aA00579c1307B0EF2c499aD98a8ce58e58",
      "metadata": {
        "name": "Tether USD",
        "symbol": "USDT",
        "decimals": 6,
        "token_type": "ERC-20",
        "icon_url": "est ullamc",
        "total_supply": "1000000"
      },
      "price_data": {
        "fiat_value": "1.00",
        "circulating_market_cap": "162513452583.22"
      },
      "counters": {
        "holders_count": "1148927",
        "transfers_count": "555"
      }
    }
  ],
  "api_key": "b7e2e2e2-2b6e-4e2a-9c7a-1e2b2e2e2e2e"
}
```

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
EvgenKor: Thank you. I have checked it and it works as expected.

EvgenKor (changes requested):
I ran the branch and tested it with the mock data — looks like the import logic needs to be improved.

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
lok52: Yes, this was intended behaviour that only one type of payload should be provided. But, I guess we can allow multiple payload types at once

EvgenKor: I expect this to be a typical scenario during the initial sync.

lok52: Updated in https://github.com/blockscout/blockscout-rs/pull/1443/commits/2960be575ee6742769d7a6c6fc9a2c673bc55813

EvgenKor: Great! It's working as expected now.

multichain-aggregator/multichain-aggregator-logic/src/repository/batch_update.rs
```diff
@@ -0,0 +1,244 @@
...
+                f_1: Set(1),
+                f_2: Set(vec![1, 2, 3]),
+                f_3: NotSet,
+                f_4: NotSet,
+            },
+            ActiveModel {
+                id_1: Set(1),
+                id_2: Set(2),
+                f_1: NotSet,
+                f_2: Set(vec![1, 2, 3]),
```
sevenzing: change to 1,2,3,4

multichain-aggregator/multichain-aggregator-logic/src/repository/batch_update.rs
```diff
@@ -0,0 +1,244 @@
...
+// https://github.com/SeaQL/sea-orm/blob/c87c0145f56e171b89a3967f95d8b6b7b743bd89/src/query/insert.rs#L173-L238
+fn prepare_batch_update_query<A>(
+    models: impl IntoIterator<Item = A>,
+) -> Result<UpdateStatement, PrepareBatchUpdateError<A>>
+where
+    A: ActiveModelTrait,
+{
+    let cte_name = Alias::new("updates").into_iden();
+
+    let mut columns: Vec<_> = <A::Entity as EntityTrait>::Column::iter()
```
sevenzing: nit: columns_to_update

sevenzing (approved):

multichain-aggregator/multichain-aggregator-entity/src/tokens.rs
lok52: We can try to implement it in another PR as there could be some tricky edge cases. By default, not every `Decimal` type should be replaced with `BigDecimal` because it depends on the specified precision.

PR #1454, opened 2025-07-17
feat(stats): toggle multichain

bragov4ik:


stats/docker-compose-multichain.dev.yml
```diff
@@ -0,0 +1,139 @@
...
+      tags:
+        - stats:local-build
+    pull_policy: always
+    platform: linux/amd64
+    restart: always
+    container_name: 'stats'
+    environment:
+      - STATS__MULTICHAIN_MODE=true
+      - STATS__DB_URL=postgres://stats:n0uejXPl61ci6ldCuE2gQU5Y@stats-db:5432/stats
+      - STATS__BLOCKSCOUT_DB_URL=${STATS__BLOCKSCOUT_DB_URL:-postgresql://multichain_aggregator:ceWb1MeLBEeOIfk65gU8EjF8@db:5432/multichain_aggregator}
```
EvgenKor:
```suggestion
      - STATS__INDEXER_DB_URL=${STATS__BLOCKSCOUT_DB_URL:-postgresql://multichain_aggregator:ceWb1MeLBEeOIfk65gU8EjF8@db:5432/multichain_aggregator}
```

EvgenKor: It would also be nice to update the corresponding environment variables in `docker-compose.dev.yml`

stats/docker-compose-multichain.dev.yml
```diff
@@ -0,0 +1,139 @@
...
+      - STATS__CREATE_DATABASE=true
+      - STATS__RUN_MIGRATIONS=true
+      - STATS__SERVER__HTTP__ENABLED=true
+      - STATS__SERVER__HTTP__ADDR=0.0.0.0:8050
+      - STATS__SERVER__HTTP__MAX_BODY_SIZE=2097152
+      - STATS__SERVER__HTTP__BASE_PATH=/kek
```
EvgenKor: If there are no objections, let's remove this non-obvious prefix

stats/stats/src/data_source/types.rs
```diff
@@ -88,17 +96,25 @@ pub struct BlockscoutMigrations {
 }
 
 impl BlockscoutMigrations {
-    pub async fn query_from_db(blockscout: &DatabaseConnection) -> Result<Self, DbErr> {
+    pub async fn query_from_db(
+        is_multichain: bool,
```
EvgenKor, line 100: I think it's better to remove this flag and add the corresponding condition before calling `query_from_db`. Currently, all it does is skip the entire routine body. From my point of view, we shouldn't call this function in multichain mode at all. The `BlockscoutMigration` entity should not be aware of multichain mode or any other concerns unrelated to the Blockscout database.

stats/stats/src/charts/chart.rs
```diff
@@ -20,8 +20,9 @@ use super::{
 
 #[derive(Error, Debug)]
 pub enum ChartError {
-    #[error("blockscout database error: {0}")]
-    BlockscoutDB(DbErr),
+    // todo: rename
```
EvgenKor:
```suggestion
```

stats/stats/src/charts/db_interaction/read/multichain.rs
```diff
@@ -0,0 +1,37 @@
+use multichain_aggregator_entity::block_ranges;
+use sea_orm::{
+    DatabaseConnection, DbErr, EntityTrait, FromQueryResult, QuerySelect, sea_query::Expr,
+};
+
+#[derive(FromQueryResult)]
+struct MinBlock {
+    min_block: Option<i32>,
```
EvgenKor: Are you sure that the `i32` type is sufficient for storing the block number? Modern rollups have block counts only about an order of magnitude below the limit of i32. E.g., Arbitrum currently has above 360M blocks.

stats/stats/src/charts/db_interaction/read/multichain.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    match min_block.and_then(|r| r.min_block) {
+        Some(min_block) => Ok(min_block as i64),
+        None => {
+            tracing::warn!("no block ranges found in multichain database");
+            // set max so that if ranges appear, the reupdate is triggered
+            Ok(not_found_value as i64)
+        }
+    }
+}
```
EvgenKor: I'm afraid the semantics of this routine may be incorrect. The `block_ranges` table stores the minimum and maximum block numbers for each chain (with one row per supported chain). However, this code finds the **minimum** value **among the maximum block numbers** across all chains. Using this result seems irrelevant in the upstream context.

stats/stats/src/charts/db_interaction/read/multichain.rs
bragov4ik: hmm, as far as I remember, there were problems fetching i64 when the stored type is i32 (it's the one chosen in multichain aggregator)

Let me check

stats/stats/src/charts/db_interaction/read/multichain.rs
bragov4ik: I agree, thank you for the catch!

stats/stats/src/data_source/types.rs
bragov4ik: I don't like to repeat the same code over and over again, so I made it into `IndexerMigration` that can be aware of the multichain mode without any contradicitons

stats/stats/src/charts/db_interaction/read/multichain.rs
bragov4ik: fixed it together

stats/stats/src/charts/db_interaction/read/multichain.rs
bragov4ik: agreed to sum all min blocks across chains

stats/stats/src/data_source/types.rs
EvgenKor: I think it's a better approach

EvgenKor (approved):
Everything looks good on my end. Nice job, thanks!

PR #1477, opened 2025-08-05
feat(multi): remove contract name search & add domains cache

lok52:


multichain-aggregator/multichain-aggregator-logic/src/services/search.rs
```diff
@@ -121,10 +122,11 @@ pub async fn search_addresses(
...
                 (vec![address], None)
             } else if domain_name_with_tld_regex().is_match(&query) {
-                let domains = search_domains(
-                    bens_client,
+                let domains = search_domains_cached(
+                    domain_search_cache,
+                    bens_client.clone(),
                     query.clone(),
-                    bens_protocols,
+                    bens_protocols.map(|p| p.to_vec()),
```
bragov4ik: probably nit, but it might be better to use `Arc` to still be able to pass a "reference" even when owned type is needed for caching
see branch `bragov4ik/review-lok52/multi-remove-contract-search`

multichain-aggregator/multichain-aggregator-logic/src/services/search.rs
```diff
@@ -394,6 +396,39 @@ pub async fn search_domains(
...
+    primary_chain_id: ChainId,
+    page_size: u32,
+    page_token: Option<String>,
+) -> Result<(Vec<Domain>, Option<String>), ServiceError> {
+    let key = format!(
+        "{}:{}:{}:{}:{}",
+        query,
+        protocols.as_ref().map(|p| p.join(",")).unwrap_or_default(),
+        primary_chain_id,
+        page_size,
```
bragov4ik, line 413: As I understand, cache for search with `page_size=5` (and empty token) won't help for exactly the same search but `page_size=1` (quick search??). Is it true?

If so, I'm thinking about ways to improve this; restricting page size is not quite good

Maybe we can at least populate `page_size`=1 cache additionally when querying larger sizes? Vice versa won't work but it sounds useful IMO

bragov4ik (approved):
lgtm

multichain-aggregator/multichain-aggregator-logic/src/services/search.rs
lok52: Good idea. I guess, we could also use something like `Box::leak` to acquire a static reference to the list of protocols and use it throughout in functions. Maybe store `&'static [&'static str]` to reduce indirection even further, but that's too much optimizations I guess.

lok52: Anyway, I tried the `&'static [String]` way in [acc8520](https://github.com/blockscout/blockscout-rs/pull/1477/commits/acc85203683cb9e70bce980ab3d949176f7a401b)

multichain-aggregator/multichain-aggregator-logic/src/services/search.rs
lok52: >Is it true?

Yes

Currently, BENS domain lookups do not support full text search. It searches only for full input string matches. So, the result set for a particular chain id is always either 0 or 1 element.

As a simple optimization I decided to set a default page_size for all queries to 50 (`api_settings.default_page_size`). This should improve number of cache hits while providing the correct data as before.

PR #1482, opened 2025-08-08
feat(stats): zetachain cctx charts

bragov4ik:


stats/stats-server/src/settings.rs
```diff
@@ -312,6 +322,29 @@ pub fn handle_enable_all_eip_7702(
     }
 }
 
+pub fn handle_enable_all_zetachain_cctx(
+    multichain_mode: bool,
+    enable_all: bool,
+    charts: &mut config::charts::Config<AllChartSettings>,
+) {
```
EvgenKor: I’m a bit confused about the purpose of the `enable_all` and `multichain_mode` flags here. Since the name suggests an “enable all” operation, having both flags feels redundant and makes it easy to call in an incompatible mode. Could we simply prevent incompatible calls at the call sites?

stats/stats-server/src/settings.rs
```diff
@@ -312,6 +322,29 @@ pub fn handle_enable_all_eip_7702(
...
+        if multichain_mode {
+            panic!(
+                "Multichain mode is not compatible with enabling local cross chain transactions stats"
+            );
+        }
+        enable_charts(
+            &[
+                // TODO: uncomment!
+                // NewCrossChainTxns::key().name(),
+                // CrossChainTxnsGrowth::key().name(),
```
EvgenKor: Does the PR not include any new charts for now?

stats/stats-server/src/settings.rs
bragov4ik: I'm not sure if this comment was written before I added panic, but the idea is to have all zetachain-enabled logic here; and it also checks if `multichain` is turned on or not, since this "crosschain" can be mixed up with "multichain" easily

stats/stats/src/preludes.rs
```diff
@@ -0,0 +1,113 @@
...
+//         MissingDatePolicy, Named, ResolutionKind, counters,
+//         db_interaction::read::{
+//             ApproxUnsignedDiff, QueryAllBlockTimestampRange, ReadError, RequestedPointsLimit,
+//             zetachain_cctx::query_zetachain_cctx_indexed_until,
+//         },
+//         indexing_status,
+//         indexing_status::IndexingStatus,
+//         lines, query_dispatch, types,
+//     };
+// }
```
sevenzing: are you sure you need it?

stats/stats/Cargo.toml
```diff
@@ -32,6 +35,7 @@ prometheus = "0.13"
 anyhow = "1"
 itertools = "0.13.0"
 rust_decimal = "1.27"
+syn = { workspace = true }
```
sevenzing: could you please check once again, are you using syn here? didnt find `syn::` in PR

stats/stats-server/src/settings.rs
```diff
@@ -312,6 +323,35 @@ pub fn handle_enable_all_eip_7702(
     }
 }
 
+pub fn handle_enable_zetachain_cctx(
```
sevenzing, line 326: maybe `if settings.second_indexer_db_url.is_none() { panic!() }`?

bragov4ik: it's checked in `connect_to_second_indexer_db`, do you think the second check is useful?

sevenzing: okay, lets leave it as it is

EvgenKor (approved):

PR #1485, opened 2025-08-11
feat(stats): multichain common counters

EvgenKor:
## There are following counters and charts were added for stat service in multichain mode:
- Number of interop transfers over all time (counter)
- Number of transactions across all chains in the cluster (counter)
- Number of addresses across all chains in the cluster (counter)
- Number of new transactions across all chains (chart)
- Daily transactions across all chains for the past 30 days (chart)

### DIFF
--- a/stats/stats-proto/proto/stats.proto
+++ b/stats/stats-proto/proto/stats.proto
@@ -185,6 +185,8 @@ message UpdateStatus {
   ChartSubsetUpdateStatus internal_transactions_dependent_status = 4;
   // Status of all enabled charts that depend on indexing user operations and internal transactions
   ChartSubsetUpdateStatus user_ops_dependent_status = 5;
+  // Status of all enabled charts that depend on indexing zetachain cctx
+  ChartSubsetUpdateStatus zetachain_cctx_dependent_status = 6;
 }
 
 message BatchUpdateChartsRequest {
--- a/stats/stats-proto/swagger/stats.swagger.yaml
+++ b/stats/stats-proto/swagger/stats.swagger.yaml
@@ -434,6 +434,9 @@ definitions:
       user_ops_dependent_status:
         $ref: '#/definitions/v1ChartSubsetUpdateStatus'
         title: Status of all enabled charts that depend on indexing user operations and internal transactions
+      zetachain_cctx_dependent_status:
+        $ref: '#/definitions/v1ChartSubsetUpdateStatus'
+        title: Status of all enabled charts that depend on indexing zetachain cctx
     title: |-
       Each field represents status of charts with corresponding dependency
       as well as all charts with weaker requirements
--- a/stats/stats-server/src/update_service.rs
+++ b/stats/stats-server/src/update_service.rs
@@ -478,26 +478,18 @@ impl UpdateService {
     }
 
     pub async fn get_initial_update_status(&self) -> proto_v1::UpdateStatus {
+        let tracker = &self.init_update_tracker;
         proto_v1::UpdateStatus {
-            all_status: self.init_update_tracker.get_all_status().await.into(),
-            independent_status: self
-                .init_update_tracker
-                .get_independent_status()
-                .await
-                .into(),
-            blocks_dependent_status: self
-                .init_update_tracker
-                .get_blocks_dependent_status()
-                .await
-                .into(),
-            internal_transactions_dependent_status: self
-                .init_update_tracker
+            all_status: tracker.get_all_status().await.into(),
+            independent_status: tracker.get_independent_status().await.into(),
+            blocks_dependent_status: tracker.get_blocks_dependent_status().await.into(),
+            internal_transactions_dependent_status: tracker
                 .get_internal_transactions_dependent_status()
                 .await
                 .into(),
-            user_ops_dependent_status: self
-                .init_update_tracker
-                .get_user_ops_dependent_status()
+            user_ops_dependent_status: tracker.get_user_ops_dependent_status().await.into(),
+            zetachain_cctx_dependent_status: tracker
+                .get_zetachain_cctx_dependent_status()
                 .await
                 .into(),
         }
--- a/stats/stats-server/src/update_tracker.rs
+++ b/stats/stats-server/src/update_tracker.rs
@@ -84,6 +84,14 @@ impl InitialUpdateTracker {
         self.inner.lock().await.user_ops_dependent.get_status()
     }
 
+    pub async fn get_zetachain_cctx_dependent_status(&self) -> proto_v1::ChartSubsetUpdateStatus {
+        self.inner
+            .lock()
+            .await
+            .zetachain_indexer_dependent
+            .get_status()
+    }
+
     pub async fn get_all_status(&self) -> proto_v1::ChartSubsetUpdateStatus {
         self.inner.lock().await.get_all_status()
     }
--- a/stats/stats-server/tests/it/common.rs
+++ b/stats/stats-server/tests/it/common.rs
@@ -47,6 +47,7 @@ pub enum ChartSubset {
     InternalTransactionsDependent,
     #[allow(unused)]
     UserOpsDependent,
+    ZetachainCctxDependent,
     AllCharts,
 }
 
@@ -61,6 +62,7 @@ pub async fn wait_for_subset_to_update(base: &Url, subset: ChartSubset) {
                 statuses.internal_transactions_dependent_status()
             }
             ChartSubset::UserOpsDependent => statuses.user_ops_dependent_status(),
+            ChartSubset::ZetachainCctxDependent => statuses.zetachain_cctx_dependent_status(),
             ChartSubset::AllCharts => statuses.all_status(),
         };
         if matching_status == proto_v1::ChartSubsetUpdateStatus::CompletedInitialUpdate {
--- a/stats/stats-server/tests/it/mock_blockscout_simple/stats_not_indexed.rs
+++ b/stats/stats-server/tests/it/mock_blockscout_simple/stats_not_indexed.rs
@@ -130,6 +130,7 @@ pub async fn run_tests_with_user_ops_not_indexed() {
     init_server(|| stats(settings, Some(shutdown_cloned)), &base).await;
 
     wait_for_subset_to_update(&base, ChartSubset::InternalTransactionsDependent).await;
+    wait_for_subset_to_update(&base, ChartSubset::ZetachainCctxDependent).await;
 
     let tests: JoinSet<_> = [
         test_lines_ok(


stats/Dockerfile
```diff
@@ -32,9 +32,9 @@ COPY --from=build /app/target/release/stats-server /app/stats-server
 RUN chown -R $APP_USER:$APP_USER /app
 USER app
 
-COPY ./config/charts.json ./config/charts.json
-COPY ./config/layout.json ./config/layout.json
-COPY ./config/update_groups.json ./config/update_groups.json
+COPY ./config/blockscout_instance/charts.json ./config/blockscout_instance/charts.json
+COPY ./config/blockscout_instance/layout.json ./config/blockscout_instance/layout.json
+COPY ./config/blockscout_instance/update_groups.json ./config/blockscout_instance/update_groups.json
```
bragov4ik, line 37: these configs are also considered in multichain mode, let's maybe leave them in `config/` folder as a base?
see `charts_config_paths`, for example

bragov4ik: alternatively, let's only consider one config file instead of joining base with multichain.

i.e. remove the logic I added with `charts_config_paths`, since it's a multichain "mode" which implies that regular mode is not applicable in this case

stats/config/multichain/charts.json
```diff
@@ -4,7 +4,28 @@
...
+            "description": "Number of transactions across all chains in the cluster"
+        },
+        "total_addresses_number": {
+            "title": "Total addresses",
+            "description": "Number of addresses across all chains in the cluster"
         }
     },
-    "line_charts": {}
+    "line_charts": {
+        "new_txns_multichain": {
```
bragov4ik, line 22: do we need a cumulative chart with total transactions as well? it's easy to compute based on `new_txns` and I was often requested to add a total number together with daily increase

stats/config/multichain/update_groups.json
```diff
@@ -1,5 +1,10 @@
 {
     "schedules": {
-        "total_interop_messages_group": "0 0 4 * * * *"
+        "total_interop_messages_group": "0 2 * * * * *",
+        "total_interop_transfers_group":  "0 2 * * * * *",
+        "total_txns_number_group": "0 10 * * * * *",
+        "total_addresses_number_group": "0 10 * * * * *",
+        "new_txns_multichain_group": "0 0 0 1 * * *",
+        "new_txns_multichain_window_group": "0 0 0 1 * * *"
```
bragov4ik: it looks like updates will happen once per month, is it intended?

stats/stats/src/charts/counters/multichain/total_addresses_number.rs
```diff
@@ -0,0 +1,106 @@
...
+
+        let data = DateValue::<String> {
+            timespan: timespan.date_naive(),
+            value: result.to_string(),
+        };
+        Ok(data)
+    }
+}
+
+pub type TotalAddressesNumberRemote = RemoteDatabaseSource<TotalAddressesNumberQueryBehaviour>;
```
bragov4ik: Just to clarify, is `PullOneNowValue` not used here only because `.unwrap_or(BigDecimal::from(0))` is needed?

bragov4ik: If it's the case, I suggest to create another type like `PullOneNowValue` and reuse it in different charts;

Also it can be used instead of `TotalInteropMessagesQueryBehaviour`

stats/stats/src/charts/counters/multichain/total_addresses_number.rs
```diff
@@ -0,0 +1,106 @@
...
+
+pub struct Properties;
+
+impl Named for Properties {
+    fn name() -> String {
+        "totalAddressesNumber".into()
```
bragov4ik: nit: for consistency I suggest
```suggestion
        "totalMultichainAddresses".into()
```

bragov4ik: The names must be unique (I added this to be mentioned more explicitly in my PR); so I think this is the better way to ensure uniqueness (`totalAddresses` vs `totalAddressesNumber` is kinda ambiguous)

stats/stats/src/tests/mock_multichain.rs
```diff
@@ -73,7 +101,9 @@ fn mock_interop_message(
     let from_address_hash = accounts[account_index].hash.as_ref().to_vec();
     let account_index = (account_index + 1) % accounts.len();
     let to_address_hash = accounts[account_index].hash.as_ref().to_vec();
+
     interop_messages::ActiveModel {
+        id: Set(index as i64 + 1), // для моков задаём id вручную
```
bragov4ik: english please :)

stats/stats/src/charts/db_interaction/read/mod.rs
```diff
@@ -28,19 +29,29 @@ pub enum ReadError {
     IntervalTooLarge(u32),
 }
 
-pub struct QueryAllBlockTimestampRange;
+pub struct QueryFullIndexerTimestampRange;
```
bragov4ik, line 33: cool!

stats/stats/src/charts/db_interaction/read/mod.rs
```diff
@@ -28,19 +29,29 @@ pub enum ReadError {
...
         Ok(start_timestamp..cx.time)
     }
 }
+
+pub async fn get_min_date(
+    indexer_db: &DatabaseConnection,
+    is_multichain: bool,
+) -> Result<NaiveDateTime, DbErr> {
+    if is_multichain {
+        get_min_date_multichain(indexer_db).await
```
bragov4ik, line 54: I like the idea, but I worry that it might lead to some errors.
`data_source_query_range_to_db_statement_range` replaces one unbounded side with the range returned by (now) `QueryFullIndexerTimestampRange`. if there are some interop transfers (for example), that are older than earliest imported counter (sounds reasonable), then the earliest date will be incorrect

for main blockscout it's the earliest block because no transfers/txns/blocks can be earlier than the first block

bragov4ik (approved):
will check the rest later

stats/stats/src/charts/lines/multichain/new_txns_multichain_window.rs
```diff
@@ -0,0 +1,222 @@
...
+    types::{Timespan, TimespanDuration, TimespanValue, timespans::DateValue},
+    utils::day_start,
+};
+
+use chrono::{DateTime, NaiveDate, Utc};
+use entity::sea_orm_active_enums::ChartType;
+use sea_orm::Statement;
+
+use super::new_txns_multichain::NewTxnsMultichainStatement;
+//use super::arbitrum_new_operational_txns::ArbitrumCalculateOperationalTxnsVec;
```
bragov4ik:
```suggestion
```

stats/stats/src/tests/mock_multichain.rs
```diff
@@ -114,3 +159,75 @@ fn mock_block_ranges() -> Vec<block_ranges::ActiveModel> {
...
+        date: Set(date),
+        daily_transactions_number: Set(Some(daily_transactions)),
+        total_transactions_number: Set(Some(total_transactions)),
+        total_addresses_number: Set(Some(total_addresses)),
+        created_at: Set(Default::default()),
+        updated_at: Set(Default::default()),
+    }
+}
+
+pub async fn imitate_reindex_multichain(indexer: &DatabaseConnection) {
```
bragov4ik: is it reindex somewhere in the past or just new data?

stats/stats/src/tests/mock_multichain.rs
EvgenKor: 😬

stats/config/multichain/update_groups.json
EvgenKor: Of course no, it's a mistake. It should Set to execute it each day after 10 minutes after midnight. https://github.com/blockscout/blockscout-rs/pull/1485/commits/a2849a7e3df98138f04e0b7f644d9017b133c8d2

stats/stats/src/tests/mock_multichain.rs
EvgenKor: It just add new data

stats/stats/src/charts/counters/multichain/total_addresses_number.rs
EvgenKor: Yes, I also renamed all associated names to avoid ambiguity. `total_txns_number` counter was renamed as well (to `total_multichain_txns`) https://github.com/blockscout/blockscout-rs/pull/1485/commits/2a75352ad88922294d63c842ae6e78c6c6ac4b16

stats/stats/src/charts/db_interaction/read/mod.rs
EvgenKor: Yes, I understood it when implementing this routine. But since the stats service cannot recover data for that period (when no counter data was imported), there is no reason to include this interval in the full time range. In other words, this time range reflects the period that is 'complete' for statistics. 

By the way, which possible errors did you mean?

bragov4ik: In current state yes.

However, if we add a chart like `newInteropTransfers` or similar that will show number of transfers for each day, the behaviour won't be correct.
Let's say we've indexed interop messages from May 1st, but also we've started the multichain indexer on May 15th. 
The counters would've been imported from the 15th as well.
But if a user were to query `newInteropTransfers` until May 31st (with no lower bound), the stats service would determine that only data from May 15th until May 31st is possible, therefore it won't return points for 1-14.

EvgenKor: Agree with that. Unfortunately, the `multichain-aggregator` database doesn’t have direct timestamp data that reflects the lower boundary of the time range (unlike the Blockscout instance database, where we can use the first block’s timestamp). Currently, only the `global_counters_imported` and `interop_messages` tables contain relevant data. Therefore, I’ve updated this routine to return the minimum timestamp from these tables. https://github.com/blockscout/blockscout-rs/pull/1485/commits/cbd6abad6cd41ee18836ee05f0dbeeaed26cac64

stats/stats/src/tests/mock_multichain.rs
bragov4ik: let's maybe name it accordingly (or remove if not needed?)

stats/stats/src/charts/counters/multichain/total_addresses_number.rs
EvgenKor: I’m just using this code as a reference from a randomly picked counter. :)
Now I’m rewriting all counters to work with `PullOneNowValue`.
I’ve customized the SQL query to avoid unnecessary unwrapping by using `COALESCE`.

https://github.com/blockscout/blockscout-rs/pull/1485/commits/724cd3f39ac63d9abca952682a6c501acfb159b0

stats/stats/src/tests/mock_multichain.rs
EvgenKor: Renamed to `add_mock_multichain_data` in https://github.com/blockscout/blockscout-rs/pull/1485/commits/8f0a90d4e3691d32d77ba70048065213f2244032

stats/config/multichain/charts.json
EvgenKor: Good point, added `txns_growth_multichain` and `accounts_growth_multichain` line charts. These charts are based on the associated imported values

stats/Dockerfile
EvgenKor: I preferred the second approach. First of all, it seems strange to load all charts and then disable all but the required ones in multichain mode. Secondly, introducing a new counter or chart would require manually adding it to the `disable_all_non_multichain_charts` routine. Therefore, I decided to remove this logic.

In the current setup, the `STATS__CHARTS_CONFIG`, `STATS__LAYOUT_CONFIG`, and `STATS__UPDATE_GROUPS_CONFIG` environment variables should to be configured manually for multichain mode. 

https://github.com/blockscout/blockscout-rs/pull/1485/commits/b32d422e23e2b9e0fe836d58d18b744646d1af0f

PR #1490, opened 2025-08-22
feat(multi): add token endpoints

lok52:
## Summary by CodeRabbit

- **New Features**
  - Cluster Explorer: list cluster tokens (filters: type, chain_id, cursor pagination) and fetch an aggregated token by address; address-token listing now accepts chain_id filters.

- **Performance**
  - Switched to cursor-based keyset pagination with multi-key ordering and NULLs handling; added DB index for faster, stable token and balance listings.

- **Chores**
  - Normalized chain_id inputs; adjusted aggregated token and address-balance shapes and pagination token formats.

- **Documentation**
  - Proto/HTTP/OpenAPI and serialization updated to expose new endpoints and pagination/query params.

### DIFF
--- a/multichain-aggregator/multichain-aggregator-logic/src/repository/pagination.rs
+++ b/multichain-aggregator/multichain-aggregator-logic/src/repository/pagination.rs
@@ -233,7 +233,7 @@ mod tests {
         ];
 
         let cursor = Cursor::new(
-            Some((123, None::<Vec<u8>>, "test", None::<String>, 42)),
+            Some((123, None::<Vec<u8>>, "test", None::<String>, 42, 1)),
             specs,
         )
         .unwrap();
@@ -259,7 +259,9 @@ mod tests {
                       AND (("test_model"."f_4" IS NULL
                         AND ("test_model"."f_5" > 42
                           OR ("test_model"."f_5" = 42
-                            AND FALSE)))
+                            AND ("test_model"."id" < 1
+                              OR ("test_model"."id" = 1
+                                AND FALSE)))))
                       OR "test_model"."f_4" IS NOT NULL)))))
             ORDER BY "test_model"."f_1" DESC NULLS FIRST,
               "test_model"."f_2" DESC NULLS LAST,
--- a/multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
+++ b/multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
@@ -2,7 +2,8 @@ use crate::{
     repository::{
         batch_update::batch_update,
         macros::{is_distinct_from, update_if_not_null},
-        pagination::{Cursor, KeySpec, PageOptions},
+        paginate_query,
+        pagination::KeySpec,
     },
     types::{
         ChainId,
@@ -16,7 +17,7 @@ use alloy_primitives::Address;
 use entity::tokens::{Column, Entity};
 use rust_decimal::Decimal;
 use sea_orm::{
-    ColumnTrait, ConnectionTrait, DbErr, EntityTrait, FromQueryResult, IdenStatic, IntoActiveModel,
+    ColumnTrait, ConnectionTrait, DbErr, EntityTrait, IdenStatic, IntoActiveModel,
     PartialModelTrait, QueryFilter, QuerySelect, QueryTrait, Select, TransactionError,
     TransactionTrait, prelude::Expr, sea_query::OnConflict,
 };
@@ -233,7 +234,7 @@ pub async fn list_aggregated_tokens<C>(
 where
     C: ConnectionTrait + TransactionTrait,
 {
-    let mut query = AggregatedToken::select_cols(base_normal_tokens_query(chain_ids, token_types))
+    let query = AggregatedToken::select_cols(base_normal_tokens_query(chain_ids, token_types))
         .as_query()
         .to_owned();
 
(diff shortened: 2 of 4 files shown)


multichain-aggregator/multichain-aggregator-logic/src/repository/pagination.rs
```diff
@@ -0,0 +1,263 @@
...
+        // nullable, NULLS LAST, cursor NULL: (col IS NULL AND suffix)
+        (true, NullOrdering::Last, true) => is_null.and(suffix),
+        // nullable, NULLS LAST, cursor NOT NULL: (col IS NULL) OR (col *cmp* v) OR (col = v AND suffix)
+        (true, NullOrdering::Last, false) => is_null.or(cmp).or(eq_suffix),
+        // nullable, NULLS FIRST, cursor NULL: (col IS NULL AND suffix) OR (col IS NOT NULL)
+        (true, NullOrdering::First, true) => is_null.and(suffix).or(is_not_null),
+        // nullable, NULLS FIRST, cursor NOT NULL: (col *cmp* v) OR (col = v AND suffix)
+        (true, NullOrdering::First, false) => cmp.or(eq_suffix),
+    }
+}
```
bragov4ik: I've attempted to make it more readable; do you think it makes it easier to comprehend or the original is better at this?

```suggestion
fn fold_key(k: &KeySpec, v: &Value, suffix: SimpleExpr) -> SimpleExpr {
    let eq_to_value = k.expr.clone().eq(SimpleExpr::Value(v.clone()));
    let is_null = k.expr.clone().is_null();
    let is_not_null = k.expr.clone().is_not_null();
    let is_value_null = &v.as_null() == v;
    let cmp = cmp_expr(&k.expr, &k.dir, v);

    // If value of a row is `v`, filter according to the remaining
    // pagination columns (`suffix`).
    let column_equal_to_value_case = match is_value_null {
        true => is_null.clone().and(suffix),
        false => eq_to_value.and(suffix),
    };

    // Include all strictly inequivalent rows as well as NULLs
    // (if applicable).
    let column_after_value_case = match (k.nullable, k.nulls, is_value_null) {
        // no need to consider `NULL`s for non-nullable columns
        (false, _, _) => Some(cmp),
        // nothing exists after the `NULL`s for `NullOrdering::Last`
        (true, NullOrdering::Last, true) => None,
        (true, NullOrdering::Last, false) => Some(cmp.or(is_null)),
        // after `NULL` everything is `IS NOT NULL`
        (true, NullOrdering::First, true) => Some(is_not_null),
        (true, NullOrdering::First, false) => Some(cmp),
    };

    let mut where_expr = column_equal_to_value_case;
    if let Some(column_after_value_case) = column_after_value_case {
        where_expr = column_after_value_case.or(where_expr);
    }

    where_expr
}
```

multichain-aggregator/multichain-aggregator-logic/src/repository/pagination.rs
```diff
@@ -0,0 +1,263 @@
...
+        let page_token = match self.page_token {
+            Some(ref page_token) => page_token.clone(),
+            None => return None,
+        };
+
+        let vals = page_token.clone().into_iter();
+
+        let mut expr = SimpleExpr::Value(Value::Bool(Some(false)));
+
+        for (k, v) in self.specs.iter().zip(vals).rev() {
```
bragov4ik, line 62: I think it would be nice to somehow report if `self.specs.len() != vals.len();`

Maybe error in `new` in this case and make fields private, or return error from `build_where_expr`

multichain-aggregator/multichain-aggregator-logic/src/repository/pagination.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+pub struct PageOptions {
+    pub page_size: u64,
+}
+
+#[derive(Clone, Debug)]
+pub struct KeySpec {
+    pub expr: SimpleExpr,
+    pub dir: Ordering,
```
bragov4ik: let's name it `direction`, since it's public api

multichain-aggregator/multichain-aggregator-logic/src/repository/pagination.rs
bragov4ik: It seems that the suggested code is more complex but more clearly shows the intent

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
```diff
@@ -175,61 +177,107 @@ where
...
-        .to_owned()
+pub fn base_normal_tokens_query(
+    chain_ids: Vec<ChainId>,
+    token_types: Vec<TokenType>,
+) -> Select<Entity> {
+    Entity::find()
+        .select_only()
+        // Use raw injected enum value to avoid suboptimal query plans
+        .filter(Expr::cust_with_expr(
+            "token_type <> 'ERC-7802'::token_type",
```
bragov4ik: It seems that there's no placeholder, so `Expr::cust` works here, doesn't it?

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
```diff
@@ -175,61 +177,107 @@ where
...
+                (
+                    a.circulating_market_cap,
+                    a.fiat_value,
+                    a.holders_count,
+                    a.name.clone(),
+                    *a.address_hash,
+                    a.chain_id,
+                )
+            }),
+        ))
```
bragov4ik: Looks to be a common pattern that can be extracted with `Cursor` later on. What do you think?

bragov4ik (approved):
lgtm!

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
lok52: Yes, missed it. Thank you!

multichain-aggregator/multichain-aggregator-logic/src/repository/pagination.rs
lok52: Agree. Changed the type of `new` to `Result` with custom error type

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
lok52: I added a `paginate_query` helper in https://github.com/blockscout/blockscout-rs/pull/1490/commits/bb7b183714b8dd473be8e8b439aaf2353757a18f. I think it should be more concise now

multichain-aggregator/multichain-aggregator-logic/src/repository/pagination.rs
bragov4ik: we decided on the existing variant with clearly visible SQLs

multichain-aggregator/multichain-aggregator-logic/src/repository/tokens.rs
bragov4ik: great!

bragov4ik (approved):

PR #1494, opened 2025-09-02
feat(bens): add domain names in domain events list

ilshat-fatkhullin:
Add domain names in domain events list

### DIFF
--- a/blockscout-ens/bens-logic/src/entity/subgraph/domain_event.rs
+++ b/blockscout-ens/bens-logic/src/entity/subgraph/domain_event.rs
@@ -12,6 +12,7 @@ pub struct DomainEvent {
     pub transaction_hash: TxHash,
     pub block_number: i64,
     pub timestamp: String,
+    pub name: String,
     pub from_address: Address,
     pub method: Option<String>,
     pub actions: Vec<String>,
--- a/blockscout-ens/bens-logic/src/subgraph/reader.rs
+++ b/blockscout-ens/bens-logic/src/subgraph/reader.rs
@@ -258,6 +258,7 @@ impl SubgraphReader {
                 .blockscout_client
                 .clone(),
             domain_txns,
+            name.inner.name
         )
         .await?;
         Ok(domain_events)
@@ -465,6 +466,7 @@ where
 async fn events_from_transactions(
     client: Arc<BlockscoutClient>,
     txns: Vec<DomainEventTransaction>,
+    domain_name: String,
 ) -> Result<Vec<DomainEvent>, SubgraphReadError> {
     let txn_ids: Vec<TxHash> = txns
         .iter()
@@ -494,6 +496,7 @@ async fn events_from_transactions(
                     transaction_hash: t.hash,
                     block_number: t.block_number,
                     timestamp: t.timestamp,
+                    name: domain_name.clone(),
                     from_address: t.from.hash,
                     method: t.method,
                     actions: txn.actions,
@@ -872,7 +875,7 @@ mod tests {
         let history = reader
             .get_domain_history(GetDomainHistoryInput {
                 network_id: DEFAULT_CHAIN_ID,
-                name,
+                name: name.clone(),
                 sort: Default::default(),
                 order: Default::default(),
                 protocol_id: None,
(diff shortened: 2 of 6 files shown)


blockscout-ens/bens-logic/src/entity/subgraph/domain_event.rs
```diff
@@ -15,4 +15,5 @@ pub struct DomainEvent {
     pub from_address: Address,
     pub method: Option<String>,
     pub actions: Vec<String>,
+    pub name: Option<String>,
```
sevenzing: lets call it `from_address_primary_name`

blockscout-ens/bens-logic/src/subgraph/reader.rs
```diff
@@ -429,6 +432,70 @@ impl SubgraphReader {
...
+        let addresses: Vec<Address> = events.iter().map(|e| e.from_address).collect();
+        if !addresses.is_empty() {
+            let names_map = self
+                .batch_resolve_address_names(BatchResolveAddressNamesInput {
+                    network_id,
+                    addresses,
+                })
+                .await?;
+            // Fill the `name` field using a resolved map (keys are 0x-lowercase strings)
+            for ev in &mut events {
```
sevenzing:
```suggestion
            for event in &mut events {
```

blockscout-ens/bens-logic/src/entity/subgraph/domain_event.rs
ilshat-fatkhullin: Applied

blockscout-ens/bens-logic/src/subgraph/reader.rs
ilshat-fatkhullin: Applied

blockscout-ens/bens-logic/src/subgraph/reader.rs
```diff
@@ -429,6 +432,70 @@ impl SubgraphReader {
...
+        if !addresses.is_empty() {
+            let names_map = self
+                .batch_resolve_address_names(BatchResolveAddressNamesInput {
+                    network_id,
+                    addresses,
+                })
+                .await?;
+            // Fill the `name` field using a resolved map (keys are 0x-lowercase strings)
+            for event in &mut events {
+                let key = hex(event.from_address);
```
sevenzing: let's make some changed in `batch_resolve_address_names`. I think this function should return `Map<Address, String>` and users of that function should decide themself if they need to serialize Address to string. In that case we dont need to serialize, but other cases may need to. By changing, we can omit this hex

sevenzing: found that in `batch_resolve_address_names` we return String because database returns string, not bytes 😠 

in that case i would change

```rust
let address_to_name: BTreeMap<String, String> = iter_to_map(
            result
                .into_iter()
                .map(|d| (d.resolved_address, d.domain_name)),
        );
```

to something like "skip wrong addresses"

```rust
let address_to_name: BTreeMap<String, String> = iter_to_map(
            result
                .into_iter()
                .filter_map(|d| d.resolved_address.parse::<Address>().map(|addr| (addr, d.domain_name)),
        );
```

ilshat-fatkhullin: Applied

blockscout-ens/bens-server/src/conversion/domain.rs
```diff
@@ -106,13 +106,13 @@ pub fn batch_resolve_from_inner(
...
 pub fn batch_resolve_from_logic(
-    output: BTreeMap<String, String>,
+    output: BTreeMap<Address, String>,
     chain_id: i64,
 ) -> Result<proto::BatchResolveAddressNamesResponse, ConversionError> {
     let names = output
         .into_iter()
         .map(|(address, name)| {
-            let address = address_from_str_logic(&address, chain_id)?.hash;
+            let address = address_from_logic(&address, chain_id, None).hash;
```
sevenzing:
```suggestion
            let address = checksummed(address, chain_id);
```

ilshat-fatkhullin: Applied

PR #1508, opened 2025-09-18
fix(verifier): empty file names when compiled with cli solc

rimrakhimov:
## Motivation
Contracts compiled with solc v0.4.10 and below are compiled not via standard-json but a usual cli interface. It is the case because those compilers do not support standard-json. Some of such compilers (at least v0.4.8 and below) as compilation result return only contract names and omit corresponding file names. 

In our previous implementation, we also ignored file names completely and used the empty string as a file name for all contracts. However, such approach resulted in inconsistency in verifier response, where `file_name` field did not exist in any of `sources`. Blockscout expects that file name has some value inside the source in order to show what the main contract is. As a result, no contracts with solc v0.4.10 could be stored in the blockscout database.

## Solution
Here we decided to completely remove support for verification of multiple files via cli solc compilers. And started to return the correct and only file name for sources where just one file exists. The decision to skip multiple files support is motivated by the fact that at the moment we have seen no verified contracts with compiler version v0.4.10 and below which has more than a single file. Besides, if such contracts existed they could be verified by flattening initial files. We can support them later if required.

### DIFF
--- a/smart-contract-verifier/smart-contract-verifier-server/tests/solidity_multiple.rs
+++ b/smart-contract-verifier/smart-contract-verifier-server/tests/solidity_multiple.rs
@@ -54,9 +54,9 @@ async fn test_setup(
 
     let prefix = format!("{CONTRACTS_DIR}/{dir}");
     let suffix = if input.is_yul { "yul" } else { "sol" };
-    input.file_path = Some(format!("{prefix}/source.{suffix}"));
+    let file_path = format!("{prefix}/source.{suffix}");
 
-    let file_path = input.file_path.clone().unwrap();
+    input.file_path = Some(file_path.clone());
     input.source_code =
         Some(input.source_code.clone().unwrap_or_else(|| {
             fs::read_to_string(&file_path).expect("Error while reading source")
--- a/smart-contract-verifier/smart-contract-verifier-server/tests/solidity_multiple.rs
+++ b/smart-contract-verifier/smart-contract-verifier-server/tests/solidity_multiple.rs
@@ -54,10 +54,13 @@ async fn test_setup(
 
     let prefix = format!("{CONTRACTS_DIR}/{dir}");
     let suffix = if input.is_yul { "yul" } else { "sol" };
-    let contract_path = format!("{prefix}/source.{suffix}");
-    input.source_code = Some(input.source_code.clone().unwrap_or_else(|| {
-        fs::read_to_string(&contract_path).expect("Error while reading source")
-    }));
+    input.file_path = Some(format!("{prefix}/source.{suffix}"));
+
+    let file_path = input.file_path.clone().unwrap();
+    input.source_code =
+        Some(input.source_code.clone().unwrap_or_else(|| {
+            fs::read_to_string(&file_path).expect("Error while reading source")
+        }));
     input.creation_tx_input = if !input.ignore_creation_tx_input {
         Some(input.creation_tx_input.clone().unwrap_or_else(|| {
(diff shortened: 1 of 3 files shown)


smart-contract-verifier/smart-contract-verifier/src/verify/solc_compiler_cli.rs
```diff
@@ -231,31 +233,79 @@ mod types {
...
+                    .map_err(|_| SolcError::Message("compiled file has unexpected prefix".into()))?
+                    .to_string_lossy()
+                    .to_string();
+
+                let (name, output) = convertible_output
+                    .output_json
+                    .contracts
+                    .into_iter()
+                    .next()
+                    .unwrap();
```
sevenzing: expect("checked that len of contracts is 1 in debug_assert above")

smart-contract-verifier/smart-contract-verifier/src/verify/solc_compiler_cli.rs
```diff
@@ -231,31 +233,79 @@ mod types {
...
+                let (_, contract_name) = split_file_and_contract_names(name);
```
sevenzing, line 275: why do u need file name if you dont use it here?

function `fn remove_path_from_contract_name(name: String) -> String` become `fn split_file_and_contract_names(name: String) -> (Option<String>, String)` for no reason 👀

smart-contract-verifier/smart-contract-verifier-server/tests/solidity_multiple.rs
```diff
@@ -54,10 +54,13 @@ async fn test_setup(
 
     let prefix = format!("{CONTRACTS_DIR}/{dir}");
     let suffix = if input.is_yul { "yul" } else { "sol" };
-    let contract_path = format!("{prefix}/source.{suffix}");
-    input.source_code = Some(input.source_code.clone().unwrap_or_else(|| {
-        fs::read_to_string(&contract_path).expect("Error while reading source")
-    }));
+    input.file_path = Some(format!("{prefix}/source.{suffix}"));
+
+    let file_path = input.file_path.clone().unwrap();
```
sevenzing:
```suggestion
    let file_path = format!("{prefix}/source.{suffix}");
    input.file_path = Some(file_path.clone());

    
```

sevenzing (approved):

smart-contract-verifier/smart-contract-verifier/src/verify/solc_compiler_cli.rs
rimrakhimov: More or less that is just a subproduct of my attempt to support multiple files in cases when the compiler specifies those file names in the output. Though eventually we decided to give up multi-file support, I believe returning both file and contract names does no harm here and allows to use this method in the multi-file `else` branch if needed. We anyway do the same in case of `remove_path_from_contract_name` but just discard file names in the function itself instead of the caller

smart-contract-verifier/smart-contract-verifier-server/tests/solidity_multiple.rs
rimrakhimov: Updated at 78b60b9910806661dfd027117cc230c3b2614bc4

smart-contract-verifier/smart-contract-verifier/src/verify/solc_compiler_cli.rs
rimrakhimov: Updated at 78b60b9910806661dfd027117cc230c3b2614bc4

PR #1563, opened 2025-11-19
feat(multi): add check redirect endpoint

lok52:
## Summary by CodeRabbit

* **New Features**
  * New "check redirect" API to detect if a search query redirects to a block, transaction, or address (includes chain info).
  * Search improved to parse hashes, addresses, and block numbers for more accurate redirect resolution.
  * Server and API docs updated to expose the new redirect check endpoint.

* **Chores**
  * Added a new dependency to support parsing/utilities.


multichain-aggregator/multichain-aggregator-logic/src/services/quick_search.rs
```diff
@@ -203,28 +226,102 @@ impl SearchTerm {
...
+                .search_hashes(hash.to_string(), None, active_chain_ids, 1, None)
+                .await?
+                .0
+                .into_iter()
+                .next()
+                .map(|h| {
+                    let redirect_type = match h.hash_type {
+                        HashType::Block => RedirectType::Block,
+                        HashType::Transaction => RedirectType::Transaction,
+                    };
```
bragov4ik, line 251: nit: `impl From<HashType> for RedirectType` makes sense for me

multichain-aggregator/multichain-aggregator-logic/src/services/quick_search.rs
```diff
@@ -203,28 +226,102 @@ impl SearchTerm {
...
+                    Redirect::new(redirect_type, h.hash, Some(h.chain_id))
+                }),
+            SearchTerm::Domain(query) => search_context
+                .cluster
+                .search_domains_cached(
+                    query,
+                    vec![search_context.domain_primary_chain_id],
+                    QUICK_SEARCH_NUM_ITEMS,
```
bragov4ik, line 259: nit: 1 is probably enough since only the first is taken?

bragov4ik (approved):
I'm not really in the context of what's the meaning of redirect and checking, but LGTM :)

multichain-aggregator/multichain-aggregator-logic/src/services/quick_search.rs
lok52: This parameter is used as a part of the cache key, so I used it here to mimic other domains requests to increase cache utilization

PR #1565, opened 2025-11-21
feat(stats): Filter data by the set of chain IDs in multichain mode

EvgenKor:
In this PR, the `stats` microservice is given the ability to work with a limited subset of networks from the `multichain-aggregator`. This is necessary when we need to build statistics for individual clusters within the multichain-aggregator. The `multichain-aggregator` database does not contain any attributes indicating a cluster association, so the `stats` microservice cannot separate data by clusters on its own. In this case, a separate `stats` service is expected to be deployed for each cluster.

To support this functionality, an optional environment variable `STATS__MULTICHAIN_FILTER` has been added to the `stats` service. It can accept a string containing chain IDs separated by commas (other delimiters, including spaces, are not allowed). If this variable is set, an additional network filter will be added to all queries made to the `multichain-aggregator` database.

To propagate the filters into the queries and minimize the amount of changes across the codebase, the following traits were extended: `StatementFromUpdateTime` and `StatementFromRange`. A `get_statement_with_context` function was added to them, along with a default trait implementation.

All multichain counters and charts were updated, and tests were added to verify correct behavior on mock data.

### DIFF
--- a/.github/workflows/stats.yml
+++ b/.github/workflows/stats.yml
@@ -51,14 +51,22 @@ jobs:
         with:
           working-directory: stats
 
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@main
+        with:
+          tool-cache: false
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: true
+          docker-images: true
+          swap-storage: true
+
       - name: Set Swap Space
         id: swapfile
         uses: pierotofy/set-swap-space@master
         with:
           swap-size-gb: 14
-
-      - name: Show disk usage
-        run: df -h
 
       - name: Build
         id: build


stats/stats/src/tests/mock_zetachain_cctx.rs
```diff
@@ -123,7 +123,7 @@ fn mock_cross_chain_tx(index: usize, ts: NaiveDateTime) -> cross_chain_tx::Activ
 }
 
 fn mock_cctx_status(index: usize, ts: NaiveDateTime) -> cctx_status::ActiveModel {
-    let status = if index % 2 == 0 {
+    let status = if index.is_multiple_of(2) {
```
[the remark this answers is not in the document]
EvgenKor: It is on a stable API currently and the compiler emits warning with expression `index % 2 == 0`

stats/stats-server/src/config/types.rs
```diff
@@ -159,6 +159,7 @@ impl EnabledChartSettings {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
+#[allow(dead_code)]
```
bragov4ik: It seems I forgot to delete these types during some refactor.
I think they're safe to remove

stats/stats/src/data_source/types.rs
```diff
@@ -39,6 +41,7 @@ impl<'a> UpdateParameters<'a> {
     pub fn query_parameters(
         db: &'a DatabaseConnection,
         is_multichain_mode: bool,
+        multichain_filter: Option<Vec<u64>>,
```
bragov4ik: kinda nit and not obvious:
`query_parameter` is for parameters passed when querying data from stats service. Since both "update charts" and "query data from stats" use these params, I've created this method to not overcomplicate querying part when only update-specific stuff is changed.

So, since we can't filter by chain id when querying data available at `stats` service, I suggest to remove this parameter and pass `None` to `Self {..}`

I will also appreciate documenting the info above in the code, since it caused a confusion! (if you have time)

bragov4ik: also the same seems to be true for `multichain_mode`. plus it's always `false` when calling `query_parameters` (like `multichain_filter` is `None`)

stats/stats/src/charts/counters/multichain/total_interop_messages.rs
```diff
@@ -1,18 +1,29 @@
...
+        if let Some(filter) = &cx.multichain_filter
+            && !filter.is_empty()
+        {
+            let chain_ids: Vec<i64> = filter.iter().map(|&id| id as i64).collect();
+            query = query.filter(
+                Condition::any()
+                    .add(interop_messages::Column::InitChainId.is_in(chain_ids.clone()))
+                    .add(interop_messages::Column::RelayChainId.is_in(chain_ids)),
+            );
+        }
```
bragov4ik: We can extract it into a function

stats/stats/src/utils.rs
```diff
@@ -47,6 +47,38 @@ pub(crate) fn produce_filter_and_values(
...
+/// ### Results
+/// Vec should be appended to the args.
+/// String should be inserted in places for filter.
+pub(crate) fn produce_multichain_filter_and_values(
+    multichain_filter: Option<&Vec<u64>>,
+    filter_by: &str,
+    filter_arg_number_start: usize,
+) -> (String, Vec<Value>) {
+    if let Some(filter) = multichain_filter {
+        if !filter.is_empty() {
```
bragov4ik: can this be rewritten as `if let Some(filter) = multichain_filter && !filter.is_empty() {`?

stats/stats-server/src/config/types.rs
EvgenKor: Removed in https://github.com/blockscout/blockscout-rs/pull/1565/commits/e83be995fb802c8f858beee972d6ce95b8aed04e

stats/stats/src/data_source/types.rs
EvgenKor: Yes, it's a good note. It really was not obvious to me. Refactored in [e83be99](https://github.com/blockscout/blockscout-rs/pull/1565/commits/e83be995fb802c8f858beee972d6ce95b8aed04e)

stats/stats/src/charts/counters/multichain/total_interop_messages.rs
EvgenKor: Introduced `apply_multichain_filter_to_interop_messages_query` helper routine in [e83be99](https://github.com/blockscout/blockscout-rs/pull/1565/commits/e83be995fb802c8f858beee972d6ce95b8aed04e)

stats/stats/src/utils.rs
EvgenKor: Sure, it's definetely clearer. Rewritten in https://github.com/blockscout/blockscout-rs/pull/1565/commits/e83be995fb802c8f858beee972d6ce95b8aed04e

PR #1566, opened 2025-11-21
feat(da-indexer): add support for eigenda_v2 blobs retrieval via proxy

rimrakhimov:
The PR adds a new endpoint which allows to retrieve blobs from eigenda V2 via 3-rd party [eigenda-proxy](https://github.com/Layr-Labs/eigenda/tree/master/api/proxy). The service, accepts the eigenda commitments and try to retrieve corresponding blob through the proxy. If successful it stores that blob in the database, so that the next time the same commitment is requested, it can be retrieved from the local database (and, thus, definitely has the corresponding blob available)

### DIFF
--- a/da-indexer/da-indexer-logic/src/eigenda/repository/blobs_v2.rs
+++ b/da-indexer/da-indexer-logic/src/eigenda/repository/blobs_v2.rs
@@ -1,7 +1,7 @@
 use crate::{common, common::repository::DbData, s3_storage::S3Storage};
 use anyhow::Context;
 use da_indexer_entity::{eigenda_v2_commitments, prelude::EigendaV2Commitments};
-use sea_orm::{sea_query::OnConflict, DatabaseConnection, EntityTrait};
+use sea_orm::{sea_query::OnConflict, DatabaseConnection, EntityTrait, TransactionTrait};
 use sha3::{Digest, Keccak256};
 
 pub async fn find_by_commitment(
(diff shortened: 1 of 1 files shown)


.github/workflows/da-indexer.yml
```diff
@@ -104,7 +106,7 @@ jobs:
       always() &&
       (needs.test.result == 'success' || needs.test.result == 'cancelled') &&
       (needs.lint.result == 'success' || needs.lint.result == 'cancelled')
-    uses: ./.github/workflows/_docker-build-push.yml
```
bragov4ik, line 107: is it meant to be reverted as name suggests? 👀

da-indexer/da-indexer-logic/src/s3_storage_test_helpers.rs
```diff
@@ -0,0 +1,150 @@
...
+        secret_access_key: String,
+        bucket_name: String,
+        client: s3::Client,
+    }
+
+    pub async fn initialize_s3_storage(test_name: &str) -> S3Storage {
+        let initialized_settings = initialize_s3_storage_and_return_settings(test_name).await;
+        S3Storage::new(initialized_settings)
+            .await
+            .expect("cannot initialize s3 storage")
```
bragov4ik, line 21: nit: it seems that `initialize_s3_storage_and_return_settings` doesn't actually initialize the storage, as initialization happens right after;

maybe rename them? not sure how though :(

bragov4ik (commented):
had some pending comments I forgot about, might check again later 🙈

PR #1568, opened 2025-11-24
fix(proxy-verifier): update responses to reflect already verified contracts

rimrakhimov:


proxy-verifier/proxy-verifier-server/src/services/mod.rs
```diff
@@ -233,6 +233,26 @@ fn process_results_response(
                 message: err.to_string(),
                 status: contract_validation_result::Status::InternalError.into(),
             },
+            Err(proxy_verifier_logic::Error::AlreadyPartiallyVerifiedContract(url)) => {
+                let message = format!(
+                    "Your code is valid and verifies this contract (partial match). However, this contract is already partially verified with identical functionality [{}].",
+                    url.replace('-', "-\u{2060}")
```
sevenzing, line 240: could you please describe in comment why u need word joiner here, not clear

rimrakhimov: Processed in [16f07a6](https://github.com/blockscout/blockscout-rs/pull/1568/commits/16f07a68cfa16d47c0ad6c91846e1dfe97b76cb1)

PR #1577, opened 2025-12-03
feat(stats): add flag to update all charts

bragov4ik:
## Summary by CodeRabbit

* **New Features**
  * Added an "update all charts" option for batch updates. When enabled, all available charts are updated regardless of the specified chart list.

### DIFF
--- a/stats/stats-server/tests/it/mock_blockscout_reindex.rs
+++ b/stats/stats-server/tests/it/mock_blockscout_reindex.rs
@@ -5,7 +5,7 @@ use blockscout_service_launcher::{
     test_server::{init_server, send_get_request},
 };
 use chrono::{Days, NaiveDate, Utc};
-use pretty_assertions::assert_eq;
+use pretty_assertions::{assert_eq, assert_ne};
 use stats::tests::{
     init_db::init_db_all,
     mock_blockscout::{default_mock_blockscout_api, fill_mock_blockscout_data, imitate_reindex},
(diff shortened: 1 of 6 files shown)


stats/stats-server/tests/it/mock_blockscout_reindex.rs
```diff
@@ -158,6 +158,11 @@ async fn test_reupdate_works() {
             ("2023-03-01", "2"),
         ])
     );
+
+    let reupdate_response =
+        request_reupdate_from(&base, &api_key, "2022-11-11", vec![], true).await;
+    assert_eq!(reupdate_response.total_rejected, 0);
+    assert_ne!(reupdate_response.accepted.len(), 0);
```
EvgenKor, line 165: Should we compare the accepted count with the total number of charts? Or do we not know the total number of charts here?

bragov4ik: It's total number of enabled charts, which is hard to get without going inside the service (we have only api url here). I.e. we need # of lines + # of counters + # of other non-listed charts, like main page/txns page/contracts page. It's if we want to get it dynamically as number of charts changes.

Hardcoding is not cool as we'll have to update the number each time

bragov4ik: With this in mind, IMO checking that it's non-zero and nothing was rejected is fine :)

Especially since this endpoint is for dev use only

PR #1598, opened 2026-01-12
feat(multi): add cluster stats

### LINKED ISSUE #1581: Metrics per supported multichain explorer chain
## What to do? 

We need to collect new metrics per supported chain on Multichain Explorer and use them to rank the order of chains that should be returned to frontend 

More details on Notion task: https://www.notion.so/blockscout/Multichain-explorer-V2-Changes-for-enabling-multiple-chains-2bd3d73641f880278b3cd254cc725e23?source=copy_link#2c43d73641f880d2b5d1ceb5471ae29d

lok52:
Closes #1581 

## Summary by CodeRabbit

* **New Features**
  * Added an API endpoint to list per-chain metrics (TPS, new addresses, daily transactions, active accounts) with week-over-week comparisons; server now exposes and serves these metrics.
  * API docs updated to include the new chain-metrics path.

* **Refactor**
  * Consolidated cache management into a single cache aggregator and wired chain-metrics caching.

* **Chores**
  * Removed environment variable prep from the build/run script.

### DIFF
--- a/multichain-aggregator/multichain-aggregator-logic/src/services/chain_metrics.rs
+++ b/multichain-aggregator/multichain-aggregator-logic/src/services/chain_metrics.rs
@@ -15,9 +15,9 @@ use std::{collections::BTreeMap, sync::Arc};
 
 const SECONDS_IN_DAY: f64 = 86_400.0;
 const DAILY_TXS_COUNTER_ID: &str = "newTxns24h";
-const NEW_ACCOUNTS_CHART: &str = "newAccounts";
-const NEW_TXS_CHART: &str = "newTxns";
-const ACTIVE_ACCOUNTS_CHART: &str = "activeAccounts";
+const NEW_ACCOUNTS_CHART_ID: &str = "newAccounts";
+const NEW_TXS_CHART_ID: &str = "newTxns";
+const ACTIVE_ACCOUNTS_CHART_ID: &str = "activeAccounts";
 
 pub async fn fetch_chain_metrics(
     blockscout_clients: &Arc<BTreeMap<ChainId, Arc<HttpApiClient>>>,
@@ -47,9 +47,9 @@ pub async fn fetch_chain_metrics(
 async fn fetch_single_chain_metrics(chain_id: ChainId, client: &HttpApiClient) -> ChainMetrics {
     let (tps, new_addresses, daily_transactions, active_accounts) = futures::join!(
         fetch_tps(client),
-        fetch_weekly_metric(client, NEW_ACCOUNTS_CHART),
-        fetch_weekly_metric(client, NEW_TXS_CHART),
-        fetch_weekly_metric(client, ACTIVE_ACCOUNTS_CHART),
+        fetch_weekly_metric(client, NEW_ACCOUNTS_CHART_ID),
+        fetch_weekly_metric(client, NEW_TXS_CHART_ID),
+        fetch_weekly_metric(client, ACTIVE_ACCOUNTS_CHART_ID),
     );
 
     ChainMetrics {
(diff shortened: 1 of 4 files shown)


multichain-aggregator/multichain-aggregator-logic/src/services/chain_metrics.rs
```diff
@@ -0,0 +1,178 @@
...
+};
+use api_client_framework::HttpApiClient;
+use chrono::{Duration, Utc};
+use std::{collections::BTreeMap, sync::Arc};
+
+const SECONDS_IN_DAY: f64 = 86_400.0;
+const DAILY_TXS_COUNTER_ID: &str = "newTxns24h";
+const NEW_ACCOUNTS_CHART: &str = "newAccounts";
+const NEW_TXS_CHART: &str = "newTxns";
+const ACTIVE_ACCOUNTS_CHART: &str = "activeAccounts";
```
bragov4ik: nit: `_ID`

multichain-aggregator/multichain-aggregator-logic/src/clients/stats.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    impl Endpoint for GetCounters {
+        type Response = Counters;
+
+        fn method(&self) -> Method {
+            Method::GET
+        }
+
+        fn path(&self) -> String {
+            "/stats-service/api/v1/counters".to_string()
```
bragov4ik, line 26: just fyi, I think this should work for all of our instances, but it's possible that the endpoint is configured to smth different than `/stats-service/...`

bragov4ik: (probably hardcoded will do, just letting you know)

multichain-aggregator/multichain-aggregator-logic/src/services/chain_metrics.rs
```diff
@@ -0,0 +1,178 @@
...
+                chart_name,
+                e
+            ))
+        })?;
+
+    calculate_weekly_metric(&line_chart)
+}
+
+fn calculate_weekly_metric(line_chart: &LineChart) -> Result<WeeklyMetric, ServiceError> {
+    let points = &line_chart
```
bragov4ik: nit: can you name it smth like `precise_points` or `full_weeks_stats`? I just got confused because I missed the filter on `is_approximate`

multichain-aggregator/multichain-aggregator-logic/src/services/chain_metrics.rs
```diff
@@ -0,0 +1,178 @@
...
+        .chart
+        .iter()
+        .filter(|p| !p.is_approximate.unwrap_or(false))
+        .collect::<Vec<_>>();
+
+    if points.is_empty() {
+        return Ok(WeeklyMetric::default());
+    }
+
+    let (previous_week, current_week) = match points.as_slice() {
```
bragov4ik: nit: often actual current week is filtered out and the second point is actually a previous week and the first one is the week before previous

bragov4ik: the logic seems correct but the naming is confusing for me 🙈 ; especially in API

multichain-aggregator/multichain-aggregator-logic/src/clients/stats.rs
lok52: Hmm, then we would have to create a config with path overrides for such special instances. Don't think it is necessary now but we should consider this option later

multichain-aggregator/multichain-aggregator-logic/src/services/chain_metrics.rs
lok52: Wdyt about `current_full_week` and `previous_full_week`?

PR #1625, opened 2026-02-08
feat(stats): Introducing Interchain mode

EvgenKor:
## Summary

This PR adds **interchain mode** to the stats service: a dedicated mode for the Universal Bridge / interchain indexer that exposes charts and counters for cross-chain messages and transfers. It also introduces a single **`Mode`** setting to control service behavior (Blockscout, Aggregator, Zetachain, Interchain) and refactors tests and the update service to use it consistently.

### 1. Mode setting
- New **`Mode`** enum: `Blockscout`, `Aggregator`, `Zetachain`, `Interchain`. Modes are mutually exclusive and drive which DBs are used, which charts/update groups are loaded, and query/update branching.
- **Config / env:** `STATS__MODE` (values: `blockscout`, `aggregator`, `zetachain`, `interchain`). Default: `blockscout`.
- **New env (interchain only):** `STATS__INTERCHAIN_PRIMARY_ID` — optional primary chain ID so send/receive counters and charts are built around it.
- The following envs are removed: `STATS__MULTICHAIN_MODE`, `STATS__ENABLE_ZETACHAIN_CCTX`. Use an associated mode instead.

### 2. Interchain charts and counters
- **Counters:** `total_interchain_messages` (and sent/received), `total_interchain_transfers` (and sent/received), `total_interchain_transfer_users`.
- **Line charts:** New messages/transfers (interchain, sent, received), plus messages growth (sent/received). Data source: interchain indexer DB (e.g. crosschain_messages).
- **Config:** New preset under `config/interchain/` (`charts.json`, `layout.json`, `update_groups.json`) and `update_groups_interchain` in the library.

### 3. Refactors
- **Tests:** Replaced the set of boolean flags (`multichain_mode`, `connect_zetachain_cctx`) with a single **`Mode`** parameter in test helpers (`simple_test_chart_inner`, `simple_test_counter_inner`, `prepare_simple_any_test`, `prepare_chart_test_inner`). Public helpers now pass the appropriate `Mode`.

### 5. Documentation
- README updated with `STATS__MODE` and `STATS__INTERCHAIN_PRIMARY_ID` in the env table.

## Testing

- Existing tests updated to use `Mode` where applicable.
- New interchain charts/counters covered by tests and mock interchain data.

### DIFF
--- a/stats/stats/src/charts/lines/interchain/new_messages_received_interchain.rs
+++ b/stats/stats/src/charts/lines/interchain/new_messages_received_interchain.rs
@@ -20,7 +20,6 @@ use crate::{
     },
     define_and_impl_resolution_properties,
     types::timespans::{Month, Week, Year},
-    utils::produce_filter_and_values,
 };
 use chrono::{DateTime, NaiveDate, Utc};
 use sea_orm::{DbBackend, Statement};
@@ -32,29 +31,28 @@ impl StatementFromRange for NewMessagesReceivedInterchainStatement {
         cx: &UpdateContext<'_>,
         range: Option<Range<DateTime<Utc>>>,
     ) -> Statement {
-        let (chain_condition, mut values): (String, Vec<sea_orm::Value>) =
-            match cx.interchain_primary_id {
-                Some(primary_id) => (
-                    " AND dst_chain_id = $1".into(),
-                    vec![sea_orm::Value::BigInt(Some(primary_id as i64))],
-                ),
-                None => (String::new(), vec![]),
-            };
-        let filter_arg_start = values.len() + 1;
-        let (range_filter, range_values) =
-            produce_filter_and_values(range, "init_timestamp::timestamp", filter_arg_start);
-        values.extend(range_values);
-        let sql = format!(
+        let (chain_condition, values) = match cx.interchain_primary_id {
+            Some(primary_id) => (
+                " AND dst_chain_id = $1".to_string(),
+                vec![sea_orm::Value::BigInt(Some(primary_id as i64))],
+            ),
+            None => (String::new(), vec![]),
+        };
+        sql_with_range_filter_opt!(
+            DbBackend::Postgres,
             r#"
                 SELECT
                     init_timestamp::date AS date,
                     COUNT(*)::TEXT AS value
                 FROM crosschain_messages
-                WHERE dst_tx_hash IS NOT NULL{chain_condition}{range_filter}
+                WHERE dst_tx_hash IS NOT NULL {chain_condition} {filter}
                 GROUP BY init_timestamp::date
-            "#
-        );
-        Statement::from_sql_and_values(DbBackend::Postgres, &sql, values)
+            "#,
+            values,
+            "init_timestamp::timestamp",
+            range,
+            chain_condition = chain_condition,
+        )
     }
 }
 
--- a/stats/stats/src/charts/lines/interchain/new_messages_sent_interchain.rs
+++ b/stats/stats/src/charts/lines/interchain/new_messages_sent_interchain.rs
@@ -20,7 +20,6 @@ use crate::{
     },
     define_and_impl_resolution_properties,
     types::timespans::{Month, Week, Year},
-    utils::produce_filter_and_values,
 };
 use chrono::{DateTime, NaiveDate, Utc};
 use sea_orm::{DbBackend, Statement};
@@ -32,29 +31,28 @@ impl StatementFromRange for NewMessagesSentInterchainStatement {
         cx: &UpdateContext<'_>,
         range: Option<Range<DateTime<Utc>>>,
     ) -> Statement {
-        let (chain_condition, mut values): (String, Vec<sea_orm::Value>) =
-            match cx.interchain_primary_id {
-                Some(primary_id) => (
-                    " AND src_chain_id = $1".into(),
-                    vec![sea_orm::Value::BigInt(Some(primary_id as i64))],
-                ),
-                None => (String::new(), vec![]),
-            };
-        let filter_arg_start = values.len() + 1;
-        let (range_filter, range_values) =
-            produce_filter_and_values(range, "init_timestamp::timestamp", filter_arg_start);
-        values.extend(range_values);
-        let sql = format!(
+        let (chain_condition, values) = match cx.interchain_primary_id {
+            Some(primary_id) => (
+                " AND src_chain_id = $1".into(),
+                vec![sea_orm::Value::BigInt(Some(primary_id as i64))],
+            ),
+            None => (String::new(), vec![]),
+        };
+        sql_with_range_filter_opt!(
+            DbBackend::Postgres,
             r#"
                 SELECT
                     init_timestamp::date AS date,
                     COUNT(*)::TEXT AS value
                 FROM crosschain_messages
-                WHERE src_tx_hash IS NOT NULL{chain_condition}{range_filter}
+                WHERE src_tx_hash IS NOT NULL {chain_condition} {filter}
                 GROUP BY init_timestamp::date
-            "#
-        );
-        Statement::from_sql_and_values(DbBackend::Postgres, &sql, values)
+            "#,
+            values,
+            "init_timestamp::timestamp",
+            range,
+            chain_condition = chain_condition,
+        )
     }
 }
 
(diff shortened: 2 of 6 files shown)


stats/stats/src/mode.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum Mode {
+    /// Single blockscout instance
+    Blockscout,
+    /// Multichain aggregator
+    Aggregator,
```
bragov4ik: nit: let's maybe name it `MultichainAggregator` or smth like that, so that it's clear in other places?

stats/stats-server/src/settings.rs
```diff
@@ -65,20 +80,12 @@ pub struct Settings {
...
-    /// This will run stats service for a multichain explorer.
-    /// It takes precedence over other chart-related flags like `enable_all_arbitrum`,
-    /// `enable_all_op_stack`, `enable_all_eip_7702`.
-    pub multichain_mode: bool,
     /// Filter by chain ids for multichain mode.
     #[serde_as(as = "Option<StringWithSeparator<CommaSeparator, u64>>")]
     pub multichain_filter: Option<Vec<u64>>,
+    /// Set the primary chain_id for Interchain mode
+    /// If the primary chain set, send/receive counters and charts will be built around it
+    pub interchain_primary_id: Option<u64>,
```
bragov4ik, line 92: As I understand, changing this parameter for a running server would not recalculate past data automatically. Which means that that data would have incorrect meaning. It requires calling full reupdate for the relevant charts. I can't think of an easy way to achieve this right now, so let's maybe just mention it in comment?

bragov4ik: I think after making 3 dimensional charts it would become easier, so we can leave it as TODO for that time

stats/stats-server/src/server.rs
```diff
@@ -55,16 +55,18 @@ pub async fn stats(
         &mut charts_config,
     );
     handle_enable_zetachain_cctx(&mut settings, &mut charts_config);
-    if settings.multichain_mode {
-        apply_multichain_mode_settings(&mut settings);
+    match settings.mode {
+        Mode::Aggregator => apply_multichain_mode_settings(&mut settings),
+        Mode::Interchain => apply_interchain_mode_settings(&mut settings),
+        _ => {}
```
bragov4ik: let's make all cases explicit and handle zetachain similarly
```suggestion
    match settings.mode {
        Mode::Aggregator => apply_multichain_mode_settings(&mut settings),
        Mode::Interchain => apply_interchain_mode_settings(&mut settings),
        Mode::Zetachain => apply_zetachain_mode_settings(&mut settings),
        Mode::Blockscout => {}
```

stats/stats-server/src/settings.rs
bragov4ik: What I'm thinking is
- make `ChartKey` dynamic depending on the set `interchain_primary_id`
- this way if the setting is changed then it will be treated as a new chart, making it update from the start
- but it does need careful review of usage of `DataSource` trait, so it's more suitable to do together with 3d charts, as I think the solution to them is similar (only that there are multiple dynamic chart keys for one data source)

stats/stats/src/data_source/kinds/local_db/mod.rs
```diff
@@ -183,14 +183,16 @@ where
...
-            get_min_block_blockscout(cx.indexer_db)
+                .map_err(ChartError::IndexerDB)?,
+            Mode::Aggregator => get_min_block_multichain(cx.indexer_db)
                 .await
-                .map_err(ChartError::IndexerDB)?
+                .map_err(ChartError::IndexerDB)?,
+            _ => get_min_block_blockscout(cx.indexer_db)
+                .await
+                .map_err(ChartError::IndexerDB)?,
         };
```
bragov4ik, line 196: I guess I missed that initially: we can use `get_min_date` here

stats/stats/src/charts/counters/blockscout_instance/zetachain_cctx/pending_cc_txns.rs
```diff
@@ -106,8 +106,9 @@ mod tests {
 
         let mut parameters = UpdateParameters {
             stats_db: &db,
-            is_multichain_mode: false,
+            mode: crate::Mode::Blockscout,
```
bragov4ik: nit: Zetachain

stats/stats/src/charts/db_interaction/read/mod.rs
```diff
@@ -48,11 +51,11 @@ impl RemoteQueryBehaviour for QueryFullIndexerTimestampRange {
...
+    mode: crate::Mode,
 ) -> Result<NaiveDateTime, DbErr> {
-    if is_multichain {
-        get_min_date_multichain(indexer_db).await
-    } else {
-        get_min_date_blockscout(indexer_db).await
+    match mode {
+        Mode::Interchain => get_min_date_interchain(indexer_db).await,
+        Mode::Aggregator => get_min_date_multichain(indexer_db).await,
+        _ => get_min_date_blockscout(indexer_db).await,
```
bragov4ik: let's also put all cases here explicitly
```suggestion
        Mode::Blockscout | Mode::Zetachain => get_min_date_blockscout(indexer_db).await,
```

so that if more modes are added this function is not missed

stats/stats/src/charts/lines/interchain/new_messages_received_interchain.rs
```diff
@@ -0,0 +1,203 @@
...
+            r#"
+                SELECT
+                    init_timestamp::date AS date,
+                    COUNT(*)::TEXT AS value
+                FROM crosschain_messages
+                WHERE dst_tx_hash IS NOT NULL{chain_condition}{range_filter}
+                GROUP BY init_timestamp::date
+            "#
+        );
+        Statement::from_sql_and_values(DbBackend::Postgres, &sql, values)
```
bragov4ik: can you check https://github.com/blockscout/blockscout-rs/pull/1628? I think this way would be cleaner and less error-prone

stats/stats/src/mode.rs
EvgenKor: I thought this naming would be better to keep shorter sentences in the settings and to avoid `snake_sace` notation for them. But yes, this naming can be slightly confused. So I rename it to `MultichainAggregator` in 01f42b55

stats/stats-server/src/settings.rs
EvgenKor: Yes, leaved comments for this parameter (and for `multichain_filter` as well). 7f4f352a

stats/stats-server/src/server.rs
EvgenKor: Done in 7160910

stats/stats/src/data_source/kinds/local_db/mod.rs
EvgenKor: Could you please clarify how you want to use `get_min_date`? Instead of `get_min_block_....` in all clauses? In this case how the result would be used in the following lines?

stats/stats/src/charts/counters/blockscout_instance/zetachain_cctx/pending_cc_txns.rs
EvgenKor: Fixed in fceb84b3

stats/stats/src/charts/db_interaction/read/mod.rs
EvgenKor: Done in 602caff8

stats/stats/src/charts/lines/interchain/new_messages_received_interchain.rs
EvgenKor: Thanks for the suggestion! It really does look better. I just fixed a few issues with docs test and clippy. [7489382](https://github.com/blockscout/blockscout-rs/pull/1625/commits/748938280a620016a16f59733d2d1e704c589bc3)

stats/stats/src/data_source/kinds/local_db/mod.rs
bragov4ik: oh, i mixed up block and date, sorry

bragov4ik (approved):

PR #1695, opened 2026-07-03
feat(stats): add filecoin chain-wide fees charts

akolotov:
> ⚠️ **Merge order:** this PR has a stacked follow-up, #1697, based on `feat/filecoin-chain-fees-chart`. Merge #1697 into this branch (`feat/filecoin-chain-fees-chart`) **first**, then merge this PR into `main`. Do not merge this PR into `main` before #1697 lands on this branch.

## What

Adds a Filecoin-specific view of **total chain-wide fees per day** (REV-style: `burn + miner tips`), served through the standard `/api/v1/lines` endpoints as two public charts with daily/weekly/monthly/yearly resolutions:

- `filecoinNewChainFees` — per-day delta
- `filecoinChainFeesGrowth` — cumulative total

They are composed from two internal (never API-exposed) charts via the established `Map<(A, B), F>` pattern:

- `burnActorBalance` — daily f099 burn-actor balance from `address_coin_balances_daily` (`FillPrevious`); `Delta` over it yields the per-day base-fee + over-estimation burn, floored at zero via a new reusable `ClampNonNegative` map kind (the f099 balance is monotonic on-chain, so a negative delta can only be a data artifact — e.g. an indexer reorg/backfill — and must not reach a public revenue chart or break the growth chart's monotonicity)
- `fevmFeeTips` — per-day FEVM miner-tip sum `(gas_price - base_fee_per_gas) * gas_used`, floored at zero per transaction via a NULL-preserving `CASE` (`FillZero`), with an outer `value IS NOT NULL` guard so a day whose blocks all lack `base_fee_per_gas` is dropped instead of failing the update; a day where only some blocks lack it passes with a deliberately-accepted understated sum (covered by a mixed-day fixture and a characterization test)

## Why

The existing `txnsFee` / `txnsFee24h` charts count **FEVM transactions only**, which on Filecoin invites DefiLlama-style misreadings of total chain fees. The f099 burn-actor balance change plus FEVM tips captures the chain-wide total.

## How it's wired

- All four charts are **disabled by default everywhere**; a new `STATS__ENABLE_ALL_FILECOIN` flag enables only the two public charts (arbitrum/op-stack/eip-7702 pattern). The intermediates stay hidden (404) and update transitively via the public charts' recursive walk.
- New `FilecoinChainFeesGroup` registered in `all_update_groups()`, with a daily cron slot in `update_groups.json`.
- Test data comes from an **isolated Filecoin fixture layer** (`mock_blockscout_filecoin.rs`) applied on top of the shared fixture only by new `simple_test_chart_filecoin*` harness wrappers — the data seen by every existing test is byte-for-byte unchanged (full suite passes unmodified).
- Server-level test `stats_filecoin_enabled.rs` boots with the flag on against a second, filecoin-seeded mock DB and asserts chart listing/resolutions, exact per-day and cumulative values, filled no-data-day behavior, growth monotonicity, and the 404/unlisted intermediates; the shared `test_lines_ok` gained a `filecoin_enabled: bool` parameter (mirroring the existing `zetachain_indexed` flag) and pins that all four names 404 when the flag is off.

## Notes for reviewers

- `burnActorBalance` SQL uses a **closed** `NaiveDate` interval via a new shared `produce_day_filter_and_values` helper (`utils.rs`, next to `produce_filter_and_values`): `sql_with_range_filter_opt!` compares timestamps, and its half-open end would drop the end day whenever a batch seam lands on midnight (same reason as `nativeCoinSupply`'s in-code comment).
- These are the first `f64`-valued `Delta` / `DailyCumulativeLocalDbChartSource` instances in the crate (existing ones wrap `i64`), so the daily chart tests assert exact f64 arithmetic end-to-end. Lower resolutions are covered by the generic resolution machinery and deliberately not re-tested per chart, matching the op-stack/arbitrum precedent.
- Negative tips are impossible by construction: the query floors each transaction's tip at zero, matching Filecoin's own semantics (a message with `gas_fee_cap < base_fee` has its tip clamped to 0 by the protocol, and the miner's shortfall is burned into f099 — i.e. already counted on the burn side). The fixture still keeps `base_fee_per_gas` NULL on blocks carrying a zero-gas-price transaction to exercise the NULL paths.
- **First-point behavior, verified on the live Filecoin mainnet:** `address_coin_balances_daily` history for f099 does **not** reach genesis — indexing starts at the FEVM launch (2023-03-14) and f099 balance rows begin 2024-02-17, already at ~39M FIL. Consequently the first daily point of `filecoinNewChainFees` is a bounded, one-off, *visible* spike (~39M FIL against ~4k FIL/day of real values), not missing data: `Delta` has no prior row and counts the whole starting balance. This is intentional and not suppressed — that same value is what makes `filecoinChainFeesGrowth` report the correct cumulative total of chain fees to date (zeroing it would undercount the total by the whole pre-index burn). Documented in the module docs of `filecoin_new_chain_fees.rs`.
- No proto changes, no migrations, no new dependencies.

## Deployment notes (for the rollout ticket)

1. **Enable the new charts.** Set `STATS__ENABLE_ALL_FILECOIN=true` on the Filecoin stats-service deployment. The full-genesis backfill runs automatically on the next service start (default `force_update_on_start` behavior).

2. **Relabel the existing FEVM-only fee charts** to prevent misreading (display metadata only; numeric values unchanged):

   ```
   STATS_CHARTS__COUNTERS__TXNS_FEE_24H__TITLE="FEVM transaction fees (24h)"
   STATS_CHARTS__COUNTERS__TXNS_FEE_24H__DESCRIPTION="Sum of FIL spent on gas fees by FEVM (Ethereum-compatible) transactions within the last 24 hours. Native Filecoin message fees are not included — see the filecoinNewChainFees chart for chain-wide fees."
   STATS_CHARTS__LINE_CHARTS__TXNS_FEE__TITLE="FEVM transaction fees"
   STATS_CHARTS__LINE_CHARTS__TXNS_FEE__DESCRIPTION="Sum of FIL spent on gas fees by FEVM (Ethereum-compatible) transactions. Native Filecoin message fees are not included — see the filecoinNewChainFees chart for chain-wide fees."
   ```

3. **Out of scope for now:** `averageTxnFee24h` / `averageTxnFee` carry the same misinterpretation risk; relabeling them was deliberately deferred (decision of 2026-07-02).

## Summary by CodeRabbit

* **New Features**
  * Added Filecoin line charts for new chain fees, chain-fee growth, burn actor balance, and FEVM miner tips.
  * Added **STATS__ENABLE_ALL_FILECOIN** to enable the primary Filecoin charts together; other related charts remain disabled by default.

* **Bug Fixes**
  * Improved date-range handling to include end days reliably.
  * Improved series correctness by clamping negative fee components to zero and filling missing days more consistently.

* **Documentation**
  * Documented **STATS__ENABLE_ALL_FILECOIN** behavior.


stats/config/blockscout_instance/update_groups.json
```diff
@@ -48,6 +48,7 @@
         "txns_stats_24h_group": "0 30 * * * * *",
         "new_builder_accounts_group": "0 45 7 * * * *",
         "verified_contracts_page_group": "0 15,45 * * * * *",
-        "zetachain_cross_chain_txns_group": "0 36 2 * * * *"
+        "zetachain_cross_chain_txns_group": "0 36 2 * * * *",
+        "filecoin_chain_fees_group": "0 25 4 * * * *"
```
EvgenKor, line 52: Why does such schedule was selected? Is it empiric one?

stats/stats/src/tests/simple_test.rs
```diff
@@ -62,6 +64,59 @@ where
...
+        test_name,
+        expected,
+        IndexerMigrations::latest(),
+        Mode::Blockscout,
+        None,
+        None,
+        true,
+    )
+    .await;
+    assert!(
```
EvgenKor: Is that assertion relevant here?

EvgenKor (approved):
Looks good to me. The introduced charts aligns well with the existing architecture. I see no obstacles to applying these changes, provided that the current codebase has been tested with a Filecoin database.

stats/stats/src/charts/lines/blockscout_instance/mod.rs
```diff
@@ -168,6 +178,8 @@ pub use verified_contracts_growth::{
     VerifiedContractsGrowthYearly,
 };
 
+pub(crate) use burn_actor_balance::{BURN_ACTOR_HASH_HEX, BurnActorBalance};
```
EvgenKor:
```suggestion
#[cfg(any(feature = "test-utils", test))]
pub(crate) use burn_actor_balance::BURN_ACTOR_HASH_HEX;
pub(crate) use burn_actor_balance::BurnActorBalance;
```

One more small suggestion. Let’s get rid of that annoying warning when compiling the service. The BURN_ACTOR_HASH_HEX constant is declared public solely for use in test modules. However, a warning about the unused constant appears during a normal service build:
```
warning: unused import: `BURN_ACTOR_HASH_HEX`
   --> stats/src/charts/lines/blockscout_instance/mod.rs:181:37
    |
181 | pub(crate) use burn_actor_balance::{BURN_ACTOR_HASH_HEX, BurnActorBalance};
    |                                     ^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` (part of `#[warn(unused)]`) on by default

```
If we move it behind the test-utils feature, it will be ignored during regular compilation.

akolotov:
Confirmed — tested this branch against a real Filecoin testnet Blockscout DB (native build, `STATS__ENABLE_ALL_FILECOIN=true`, indexer DB pointed at the live instance).

Both charts populate correctly from real data:
- `filecoinNewChainFees` — daily values dominated by the f099 burn-actor delta (tens of thousands FIL on active days), matching hand-computed `Δ(burnActorBalance) + fevmFeeTips`.
- `filecoinChainFeesGrowth` — monotonic non-decreasing, never returns to zero after the first non-zero day.

Also cross-checked the daily deltas against `address_coin_balances_daily` for f099 — the chart output matches the indexer data.

stats/config/blockscout_instance/update_groups.json
akolotov: Yes, it's mostly empiric. The rule I followed: run once per day in the early morning (UTC), with a few-hour buffer after midnight so the previous day is fully indexed — otherwise the last daily point would be incomplete.

The exact time (04:25) was chosen so the group does not start at the same moment as another group. The closest one is `active_accounts_weekly_group` at 04:10 (15 minutes earlier), and hour 4 is a relatively quiet slot compared to busier hours like 7. So there is no strict formula behind the value, but it satisfies these constraints.

stats/stats/src/tests/simple_test.rs
akolotov: Good catch — since the mode is hard-coded to `Mode::Blockscout`, `zetachain_cctx` is always `None` here, so the assertion never actually checks anything. Removed it in a0de6b1d.

Note that the same assertion exists in the sibling `simple_test_chart` helper (it was copied from there), where it's equally redundant. I left that one untouched for now to keep this PR focused, but we may want to clean it up separately.

stats/stats/src/charts/lines/blockscout_instance/mod.rs
akolotov: Good catch — the warning is real and fixed in 076ffcfc. I went with a slightly different approach than the suggestion, though.

Instead of gating the re-export behind `#[cfg(any(feature = "test-utils", test))]`, I made the constant and its crate-root re-export `pub` (instead of `pub(crate)`):

```rust
pub use burn_actor_balance::BURN_ACTOR_HASH_HEX;
pub(crate) use burn_actor_balance::BurnActorBalance;
```

Why this over the `#[cfg]` version:

- **It matches an existing pattern in the very same file.** The sibling test-only hash constants `ATTRIBUTES_DEPOSITED_FROM_HASH` / `ATTRIBUTES_DEPOSITED_TO_HASH` are handled exactly this way — defined as `pub const`, used in non-test code, and re-exported at the crate root with `pub use` for the test mock to consume. Reusing that idiom keeps the re-export block consistent rather than introducing a `#[cfg]` attribute that this block otherwise never uses on `use` lines.
- **Same effect, simpler mechanism.** A `pub use` re-export is part of the crate's public API, so the compiler can't prove it's unused and never emits the warning — no feature-conditional compilation needed. The `#[cfg]` route works too, but it couples the re-export's visibility to the test flags, whereas the root cause was simply that a `pub(crate)` re-export was only consumed from test-gated code.

Verified both build configs are clean:
- `cargo check -p stats` (default) — no warning
- `cargo check -p stats --features test-utils` — compiles

Let me know if you'd still prefer the `#[cfg]` gating for tighter visibility.

EvgenKor (approved):
Lgtm, considering the latest changes

PR #1697, opened 2026-07-09
feat(stats): serve txnsFee with the filecoinNewChainFees implementation

akolotov:
> ⚠️ **Merge order:** this PR is stacked on #1695 — base branch is `feat/filecoin-chain-fees-chart`, not `main`. Merge **this PR first**, into `feat/filecoin-chain-fees-chart`. Only after that, merge #1695 into `main` (which will carry this PR's commits along with it). Do not retarget this PR to `main` or merge #1695 before this one.

## Summary

Stacked on #1695 (base branch is `feat/filecoin-chain-fees-chart`; retarget to `main` once that merges).

A customer deployment requires the existing public chart id `txnsFee` to serve the data of the new `filecoinNewChainFees` chart (chain-wide REV-style fees), without exposing `filecoinNewChainFees` as a public id. Today the config-name → chart-implementation link is strictly one-to-one, so this PR introduces a generic, config-driven remapping and redefines `STATS__ENABLE_ALL_FILECOIN` to use it.

## What changed

One commit per phase, plus review-round commits:

- **Config plumbing** (`8dcf73ee`): new optional per-chart setting `implementation` in `charts.json` and env (`STATS_CHARTS__{LINE_CHARTS,COUNTERS}__<NAME>__IMPLEMENTATION`), snake_case in input, camelCase in memory (normalized once on config load). README env table regenerated.
- **Scheduler fix** (`67c4bc81`): `init_update_groups` derives each group's `enabled_members` from the set of enabled `ChartKey`s (via `get_keys()`) instead of a reverse public-name lookup. Behavior-preserving today; under a remap it is what makes the implementation's group update the chart and stops the replaced chart's group from updating for nothing.
- **Startup resolution + validation** (`a222c29d`): an enabled entry with `implementation = Y` claims registered chart Y's per-resolution handles and serves them under the entry's own public name; metadata (title/description/units) stays with the entry. Hard startup errors naming both sides for: unknown implementation, chart-type mismatch, double claims (two mapped entries, or mapped entry + implementation enabled under its own name), self-reference, and explicitly requested resolutions the implementation does not have.
- **Waiting-status filter** (`08dca976`): the `/api/v1/lines` waiting-status filter built `ChartKey`s from public names while the update tracker is keyed by internal names; moved next to `get_keys()` as `EnabledChartEntry::filter_out_waiting_resolutions` so a still-waiting remapped chart is hidden from the listing like any other.
- **Flag redefinition** (`87def6f4`): `STATS__ENABLE_ALL_FILECOIN=true` now enables `filecoinChainFeesGrowth` under its own id and force-enables `txnsFee` served with the `filecoinNewChainFees` implementation (an explicit operator-provided `implementation` wins over the flag). Docs updated everywhere the contract lives (helper doc, `Settings` field doc, README row).
- **Integration test rework** (`77ea0a17`): the filecoin scenario proves the remap end-to-end — filecoin-computed per-day values under `/api/v1/lines/txnsFee` with `info.id == "txnsFee"` (the fixture DB also contains regular transactions, so the old fee computation would yield different numbers), `filecoinNewChainFees` 404s and is never listed, intermediates stay hidden, and the counter set is unchanged under the flag.
- **Review round** (`bf90bedd`): `handle_disable_internal_transactions` now disables config entries by the chart they actually serve (the `implementation` when remapped, the entry's own name otherwise) — previously a remapped entry serving an internal-transactions-dependent chart stayed enabled under `STATS__DISABLE_INTERNAL_TRANSACTIONS`, and an entry merely sharing a dependent chart's name was disabled for the wrong reason (latent: not reachable through the filecoin remap). Also clarifies two validation errors: the unknown-implementation message explains the snake_case → camelCase normalization, and the also-enabled message states a reason that holds for remap chains too.
- **Second review round** (`d770464e`): restores (at `debug!` level) the drift signal dropped with the inverted disable loop — a single summary line names the internal-transactions-dependent registered charts that no config entry serves; rewords the resolution-mismatch startup error to "explicitly configures" since the branch also fires for an explicit *disable* of a missing resolution; documents in the README that an env override can set or replace an `implementation` mapping but cannot clear one defined in `charts.json`; and pins the counter→counter remap success path with a unit test (single Day resolution, singleton-group scheduling).

DB storage stays keyed by the implementation's internal name, so historical `txnsFee` rows are never mixed with the new data; unsetting the flag needs no manual catch-up (the frozen `txnsFee` storage self-heals via the ordinary incremental update on the next start).

## Notes for reviewers

- No proto changes, no migrations, no new dependencies, no changes in the `stats` crate — `stats-server` only.
- Counters (`txnsFee24h` etc.) are deliberately out of scope and unchanged; the remap mechanism supports counters generically but none are configured.
- Setting the flag together with `STATS_CHARTS__LINE_CHARTS__FILECOIN_NEW_CHAIN_FEES__ENABLED=true` fails startup via the double-claim validation — intended, it is exactly the "accidentally public under its own name" case the customer agreement rules out.
- Verified: full unit suites of both crates, all DB-backed server scenarios, fmt/clippy `-D warnings`, and the README env-docs gate.

https://claude.ai/code/session_01KJ35EWLUaZiJfzzj3DvT4L

## Summary by CodeRabbit

* **New Features**
  * Added support for mapping chart and counter configurations to alternative implementations through environment variables.
  * Filecoin mode now serves Filecoin fee data through the existing transaction-fee chart while keeping implementation-only charts hidden.
  * Added validation to prevent invalid, conflicting, or incompatible chart mappings.

* **Bug Fixes**
  * Improved filtering of charts waiting for startup conditions.
  * Ensured internal-transaction settings correctly follow remapped charts.
  * Updated Filecoin chart and counter behavior across API responses and listings.


EvgenKor (commented):
As I understand it, to override the `txnsFee` implementation, it is sufficient to set the `STATS__ENABLE_ALL_FILECOIN=true` variable, and there is no need to use the `STATS_CHARTS__{LINE_CHARTS,COUNTERS}__<NAME>__IMPLEMENTATION` variables.

But I also tried manually overriding an arbitrary counter using `STATS_CHARTS__COUNTERS__X__IMPLEMENTATION=Y`, and I was confused by the fact that, in addition to overriding the implementation for `X`, I also had to disable the `Y` counter (`STATS_CHARTS__COUNTERS__Y__ENABLED=false`).

After experimenting, I realized that I could also set the implementation to `X=Y` and disable `X`. In that case, `X` disappears, while `Y` appears with its original implementation, `Y`. Obviously, this scenario is redundant, since I could achieve the same result simply by disabling `X` without overriding its implementation.

Therefore, when overriding `X=Y`, there seems to be only one valid scenario: disabling `Y` at the same time. Otherwise, the service exits with an error.

I think it would make sense to automatically disable the chart whose implementation is being used for the override. Or, conversely, allow the service to display both charts at the same time, so the operator can disable the unnecessary one. In that case, the second ENV would at least be justified: two actions, two variables. As it stands, we have to set two variables for a single action. Perhaps there are other scenarios I have not considered. Wdyt?

akolotov:
@EvgenKor thanks — the behavior you hit is intentional, but you're right it wasn't documented. Quick clarification on why the two cases differ:

`STATS__ENABLE_ALL_FILECOIN=true` needs no second variable because its remap target, `filecoinNewChainFees`, is an *implementation-only* chart — it's never enabled under its own id. A remap onto such a target is unambiguous, so one action is enough.

Your manual experiment remapped onto `Y`, an ordinary *enabled* counter. Then `Y`'s implementation would be served under both `X` and `Y` at once, which is ambiguous — so we fail fast and ask you to disable one of them, rather than silently dropping or duplicating a chart. That's consistent with how the rest of the crate treats contradictory config (it rejects rather than auto-resolves).

So the intended pattern is: remap onto a dedicated implementation-only chart (one variable), or, if you deliberately reuse an already-active id, disable its own entry too — the second variable is a second, real decision ("remove `Y` from the API surface").

I've documented this in `stats/README.md` under "Serving a chart with another chart's implementation" (commit 844ba1ad), which also covers the `charts.json` syntax CodeRabbit asked about. Keeping the current fail-fast behavior; let me know if the docs read clearly.

EvgenKor (approved):
@akolotov thanks for the clarification. I agree that remapping an already-active chart onto another one is an odd scenario in practice. And once I look at it as: `STATS__ENABLE_ALL_FILECOIN` being the curated customization, and the generic `STATS_CHARTS__{LINE_CHARTS,COUNTERS}__<NAME>__IMPLEMENTATION` envs being the way to override/fine-tune that customization without a new release — everything falls into place.

In hindsight, what actually sent me down the wrong path was the `"disable one of them"` wording in the startup error: it made me explore disabling X instead of Y, although disabling X is never useful. Maybe worth making the error message recommend disabling the target specifically, ideally naming the exact knob, e.g. `set STATS_CHARTS__COUNTERS__<Y>__ENABLED=false or disable the '<Y>' entry in charts.json`. With the latest docs commit the rest is clear to me.

akolotov:
@EvgenKor good catch — you're right that "disable one of them" was misleading, since disabling the remapping entry just drops it (same as not configuring the remap at all), so only disabling the target is ever useful.

Fixed in debb6255: the error tail now reads `…; disable the '<target>' entry instead`. I kept it path-agnostic (it applies whether the target is defined in `charts.json` or via env) and terse to match the crate's error style, rather than embedding a specific `STATS_CHARTS__…__ENABLED=false` key. The existing test still passes (the `also enabled` substring is preserved).

Thanks for the review!

PR #1722, opened 2026-08-12
feat(stats): Filecoin chain fees 24h counter

### LINKED ISSUE #1721: stats: 24-hour chain fees counter for Filecoin
## Context

#1695 added the Filecoin chain-wide fees line charts: `filecoinNewChainFees` is served under the public `txnsFee` id (via the `implementation` remap) when `STATS__ENABLE_ALL_FILECOIN=true`. Chain fees there are REV-style: base-fee burn + over-estimation burn (captured as the balance growth of the f099 burn actor) + FEVM miner tips.

The 24-hour counter `txnsFee24h` was not covered: on Filecoin it still reports FEVM transaction fees only, so the counter and the line chart on the same page disagree about what "fees" means.

The counter cannot be derived from the existing daily chart: both of its parts have calendar-day resolution (`address_coin_balances_daily.day`, `DATE(blocks.timestamp)`), which cannot produce a rolling 24-hour window.

## Proposal

Add an internal counter `filecoin_chain_fees_24h` that queries the indexer database directly, once per hour, with a single statement:

- **Burn part** — from the per-block `address_coin_balances` table: take two *anchor* rows for the f099 burn actor (the last row at or before each edge of the `[T-24h, T)` window, bounded through `blocks` with `consensus = true`), and report the balance difference. The stored value is the exact balance at that block, so gaps between rows lose no burn; on mainnet f099 rows appear ~9.5/hour (median gap 4 min).
- **Tips part** — FEVM miner tips summed over the exact `[T-24h, T)` interval from `transactions` joined to `blocks`, with the same tip formula and NULL `base_fee_per_gas` handling as `fevmFeeTips`. Two statement forms are needed (denormalized/non-denormalized schema).
- Counter value = `max(burn, 0) + tips`, `NULL` parts treated as 0 (same as `txnsFee24h`'s `UnwrapOr<_, Zero>`).

Framework reuse: `PullOne24hCached` for the window, `DirectPointLocalDbChartSource` for storage, `txnsFee24h` as the model.

### Serving (public API surface)

- The counter gets its own `charts.json` entry with `"enabled": false` and is never exposed under its own id.
- The public `txns_fee_24h` id serves it through `implementation` remap; `handle_enable_all_filecoin` force-enables the entry and sets the mapping (unless the operator configured one explicitly) — exactly the pattern already used for `txns_fee` → `filecoin_new_chain_fees`.
- No `layout.json` changes: `txns_fee_24h` is already in `counters_order`.

### Update group

A new singleton hourly group (e.g. `"0 45 * * * * *"`), independent of the daily `filecoin_chain_fees_group` — the counter must not trigger the heavy daily chain.

### Decisions worth recording

- **Window boundaries.** The burn part is measured between two anchor blocks; the tips part over the exact timestamp interval. The skew is a few minutes, two-sided, and self-correcting on the next hourly run (each part is measured exactly within its own boundaries — nothing is lost, the edges just sit slightly differently). Accepted; stated in the module doc. A `tracing::warn!` fires when the anchor span deviates from 24h beyond a threshold (~60 min, calibrated against mainnet anchor lags of ≤47 min), and a separate warning covers degenerate/missing anchors (empty window, single anchor row).
- **Indexing status requirement**: `IndexingStatus::LEAST_RESTRICTIVE`, same as every existing 24h counter — a remap serves the public id with the implementation's properties, and `txnsFee24h` must not start waiting for the indexer on Filecoin alone.

### Testing

- No fixture currently writes `address_coin_balances` rows (the Filecoin mock layer fills `address_coin_balances_daily` only), and there is no counter-side `simple_test_counter_filecoin` helper (chart-side wrappers exist). Both need to be added: extend `mock_blockscout_filecoin.rs` with f099 per-block rows plus blocks inside the default 24h test window (the layer's `MIXED_DAY` technique), keeping the per-block and daily series consistent; thread the `fill_filecoin` flag into the counter test engine.
- DB-backed tests over several `update_time` windows (anchors strictly inside the edges, empty window, degenerate anchor), asserting the chosen anchors and not only the sum; SQL-text pins for both statement forms; `handle_enable_all_filecoin` unit tests and the `stats_filecoin_enabled` integration scenario extended to the counter.

akolotov:
## What

Adds a new Filecoin-specific counter, `filecoin_chain_fees_24h`, served publicly as `txnsFee24h`. It reports total chain fees over a rolling 24-hour window as the sum of two terms:

- **burn** — the f099 burn-actor balance delta between the window's anchor blocks;
- **tips** — priority fees paid by transactions inside the window.

The counter has its own hourly update group (`0 45 * * * * *`) rather than joining the daily `FilecoinChainFeesGroup`, since its cadence differs.

## Why

Closes #1721. The existing Filecoin fee charts are daily line charts; there was no 24h rolling counter for the chain-fees figure.

## Changes

- **New counter** `stats/stats/src/charts/counters/blockscout_instance/filecoin_chain_fees_24h.rs` — SQL statement in both migration forms, `FilecoinChainFees24hValue`, the `CombineBurnAndTips` map function with degenerate-anchor and window-skew warnings, and full unit + DB-backed tests.
- **Wiring** — `CacheValue::ValueFilecoinChainFees24h`, `singleton_groups!` entry, `runtime_setup.rs` registration, cron entry in `update_groups.json`.
- **Config** — `charts.json` entry (disabled by default) and the `handle_enable_all_filecoin` enable-switch remap of `txns_fee_24h` → `filecoin_chain_fees_24h`.
- **Test fixture** — `mock_blockscout_filecoin.rs` gains a per-block `address_coin_balances` fixture and an 8-block counter-window series (dense/sparse/NULL rows, a window-edge boundary block), plus `BOUND_EDGES` and a `block_numbers_follow_time_at_counter_edges` DB test. Existing chart expectations were recomputed from real DB runs, not hand-adjusted.
- **Integration tests** — the Filecoin-enabled scenario asserts the counter's metadata, value and remap; the default-disabled scenario asserts its absence.

No proto, migration, entity, or `Cargo.toml` changes.

## Notes for reviewers

- The window is half-open `[T-24h, T)`. The block at exactly the upper edge is deliberately excluded — this is what the fixture's boundary block and the two differing pins (`3000.0001029` under the frozen chart clock vs `3000.0002058` under the integration server's wall clock) exist to prove.
- Per-block f099 rows are stored in **wei** (`fil * ETHER`), matching the daily builder; `counter_window_fixture()` keeps FIL-unit literals so the expected values stay hand-computable.
- `tips_txns` for the default window is **4**, not 2: the shared blockscout fixture's block at `2023-03-01T10:00:00` also falls inside the window and carries two transactions with NULL tip terms — skipped by `SUM`, still counted by `COUNT(*)`. The sum literal is unaffected.

## Testing

- `cargo test -p stats -p stats-server -j 1` — pass.
- `cargo-db test -p stats -p stats-server -j 1 -- --include-ignored` — 354 + 65 + 13 tests pass, 0 failures.
- `cargo fmt --all --check` and `cargo clippy --all --all-targets --all-features -- -D warnings` — clean on a cold build cache.

One caveat: in the full DB-backed run, two **doctests** in `stats/src/utils.rs` fail with `linking with cc failed`. That file is untouched by this branch; the same doctests pass in the non-DB run and pass when re-run in isolation. It is linker memory pressure from building the full suite alongside the Postgres container, not a code defect.

## Summary by CodeRabbit

- **New Features**
  - Added a Filecoin 24-hour chain fees metric combining burn fees and FEVM miner tips.
  - Exposed the metric through the public `txnsFee24h` chart.
  - Added hourly updates and migration-aware transaction support.
  - Added configurable interchain filters for home chains, counterparties, sources, destinations, bridges, and unindexed chains.

- **Bug Fixes**
  - Improved fee calculations for negative burn values, missing data, and boundary conditions.
  - Updated Filecoin fee totals and chart results to include miner tips consistently.

- **Documentation**
  - Clarified chart ID uniqueness requirements and duplicate handling.


stats/stats-server/src/settings.rs
```diff
@@ -269,31 +271,63 @@ pub fn handle_disable_internal_transactions(
...
+/// `RuntimeSetup::build_charts_info` drops disabled entries before its
+/// collision check — so for a duplicate id whose other side is disabled the
+/// `lines` probe below wins.
+fn find_chart_settings_mut<'a>(
+    charts: &'a mut config::charts::Config<AllChartSettings>,
+    id: &str,
+) -> Option<&'a mut AllChartSettings> {
+    match (charts.lines.get_mut(id), charts.counters.get_mut(id)) {
+        (Some(settings), _) => Some(settings),
+        (_, Some(settings)) => Some(settings),
```
EvgenKor, line 370: Could we add an explicit warning for the (Some(line), Some(counter)) case?
This configuration is admittedly unlikely, but I don't think we should silently ignore it. The helper currently prefers the line-chart entry solely because of the match order. This can make enable_all_filecoin enable or remap a different entry than the operator intended, with the actual failure only appearing later during startup as a type mismatch or duplicate chart error.
The README documents the restrictions around implementation mappings, but it does not state that chart IDs must be unique across the lines and counters sections, nor that a line-chart entry takes precedence when both exist. Therefore, an operator currently has no way to predict this behavior.
At minimum, could we emit a warning containing the conflicting chart ID, both entries’ enabled states, and the fact that the line-chart entry was selected? This case may be rare, but making the ambiguity visible would make it much easier to diagnose.

EvgenKor (approved):
Overall, the implementation looks solid and thoroughly tested. I left one small request about emitting a warning when the same chart ID exists in both the lines and counters sections.
My main remaining concern is the failing CI/CD Docker job. I understand that it fails before the actual build because the PR originates from a fork and the ARM_RUNNER_HOSTNAME / ARM_RUNNER_KEY secrets are unavailable. Therefore, this does not appear to be caused by the changes in this PR, and the workflow should receive those secrets after a push to main.
Still, I am uncomfortable completely ignoring a red CI/CD check. A failing pipeline reduces confidence that the merge result has been fully validated. Ideally, the Docker build should either be verified from a trusted same-repository context before merging, or the workflow should explicitly handle fork PRs by skipping the secret-dependent ARM setup or running a secretless build variant.

stats/stats-server/src/settings.rs
akolotov: Good point — done in de9b0a06.

`find_chart_settings_mut` now matches the `(Some(line), Some(counter))` case explicitly and warns before returning the line-chart entry. The message names the conflicting id, the `enabled` state of both entries, and states that the line-chart entry was selected and may not be the intended one:

```
Chart id '<id>' is served by an entry in both config sections: line charts
(enabled: <bool>) and counters (enabled: <bool>). Chart ids must be unique
across both sections; the line chart entry is selected here, which may not
be the intended one. Rename or remove one of the two entries.
```

I kept the selection itself unchanged rather than, say, preferring the enabled side: that would introduce a resolution rule that exists nowhere else in the config handling, while the two sibling branches of this function already follow the `warn!`-and-continue pattern used throughout `settings.rs`. Making the ambiguity visible is what was missing, not the choice.

On the docs gap: `stats/README.md` now states under "Charts parameters" that ids must be unique across the `counters` and `line_charts` sections, that two enabled duplicates fail at startup, and that with one side disabled the service starts but the `STATS__ENABLE_ALL_*` flags act on the `line_charts` entry, with the warning logged.

Also added `duplicate_id_across_sections_resolves_to_the_line_chart_entry` in `settings.rs` to pin the preference, so a future reordering of the match arms cannot change it silently.

EvgenKor: Thanks for resolving this

PR #1728, opened 2026-08-19
feat(stats): Interchain read filter

EvgenKor:
# Interchain read filter: configurable chain/bridge scope for stats

## Summary

- `Mode::Interchain` gains a configurable read filter that is the deployment-time
  equivalent of the interchain-indexer's read-time API filtering
  (blockscout-rs#1708): a `stats` instance can now scope every counter and line
  chart to a chain/bridge slice, so one deployment can serve one frontend against
  a *universal* interchain-indexer — one instance indexing every chain and bridge
  it can reach, discarding nothing.
- Parity with the indexer's filter is structural, not re-derived: stats takes the
  indexer's `interchain-indexer-filters` crate (merged upstream in
  blockscout-rs#1727) as a dependency and delegates to its `ChainBridgeFilter`
  wholesale, rather than reimplementing the focal/directional/bridge truth table
  locally. Verified end to end against a running dev indexer on one database —
  see Verification. Two named divergences remain, both reachable only by editing
  the indexer's configuration — see Follow-ups.
- All 13 interchain SQL statements plus the timespan floor
  (`get_min_date_interchain`) now apply the shared predicate, including four
  charts (`totalInterchainMessages`, `totalInterchainTransfers`,
  `newMessagesInterchain`, `newTransfersInterchain`) that previously ignored
  filtering entirely; their ids are unchanged, their meaning is now "within the
  configured scope."
- Along the way, fixes a latent, pre-existing over-counting bug: transfer→message
  joins now use the composite `(message_id, bridge_id)` key instead of
  `message_id` alone, closing a fan-out where two bridges reusing the same
  numeric message id inflated transfer counts.
- **Reviewing note on the diff size:** the first commit (`b2c812db`) is unrelated
  tooling — it bootstraps `.memory-bank/` for `stats/` (docs, coding rules, the
  six reusable workflows) mirroring what `interchain-indexer` already has, plus
  the research note that scoped this task. It touches no Rust. Skipping it leaves
  the four `feat`/`test` commits that carry the change, two docs commits, a
  one-line dev-recipe fix, and `c07f3d68`, which acts on an independent review of
  the branch (see Review pass). The graphify knowledge graph is deliberately
  gitignored here, unlike in `interchain-indexer`, so it adds nothing to the
  diff.
- No deployment exists yet in the target (universal-indexer) configuration — the
  universal indexer and the stats services reading it both launch from clean
  databases, later, once frontends are repointed. Numeric impact is therefore
  characterised by test fixtures, not measured against any running deployment.

## Implementation Plan

- **Settings & validation** (`stats-server/src/settings.rs`,
  `stats-server/src/interchain_filter.rs`): six new vars under
  `STATS__INTERCHAIN_FILTER__*`, validated fail-fast at startup — any field set
  outside `Mode::Interchain`, an id overflowing the signed column type, or
  `STATS__INTERCHAIN_PRIMARY_ID` disagreeing with `HOME_CHAIN_ID` all abort
  startup with a message naming the offending variable.
- **Shared predicate** (`stats/src/charts/db_interaction/filters/interchain.rs`):
  `InterchainFilter` wraps the indexer's `ChainBridgeFilter` verbatim instead of
  re-deriving the focal `(home, counterparty)` truth table, the outer-`AND`
  placement of directional/bridge terms, or the transfers-use-their-own-columns
  rule.
- **Observability horizon**: the "indexed chain set" restriction
  (`INCLUDE_UNINDEXED_CHAINS`) is resolved once per update-group cycle directly
  from the indexer's own `bridges` / `bridge_contracts` tables, not from stats
  configuration; a resolution failure skips the group for that cycle rather than
  silently running unfiltered. Deliberately does not filter on `bridges.enabled`,
  since the indexer clears that column for every row inside its own upsert
  transaction. The resolver returns a pair list unconditionally — including an
  empty one for an empty `bridges` table — because that is what the indexer's
  `configured_pairs` does; the "no restriction" variant is unreachable in any
  indexer deployment stats could be reading.
- **Statement conversion** (`stats/src/charts/{counters,lines}/interchain/*.rs`):
  all 13 hand-written SQL statements converted to typed builders over the shared
  filter. Transfer charts now filter on the transfer's own
  `token_src_chain_id`/`token_dst_chain_id` and `bridge_id`, matching the API,
  instead of the joined message's route; the message join survives only where a
  chart needs `init_timestamp` (transfers carry no timestamp of their own), and
  is now composite on `(message_id, bridge_id)`.
- **Timespan floor** (`stats/src/charts/db_interaction/read/interchain.rs`):
  `get_min_date_interchain` is filtered too — unfiltered, it would start a first
  full computation over a universal indexer's whole history at a foreign date and
  batch through a long run of provably empty days.
- **Staleness handling** (`stats/src/data_source/kinds/local_db/mod.rs`,
  `stats/src/charts/db_interaction/read/local_db.rs`): a filter fingerprint rides
  in the existing nullable `chart_data.min_blockscout_block` column (repurposed;
  that column already meant nothing in interchain mode, where
  `get_min_block_interchain` is a constant `i64::MAX`). A mismatch clears the
  chart's stored data and forces a recompute on the next update cycle, gated on
  `Mode::Interchain` and independent of `FORCE_UPDATE_ON_START`. Counters
  self-heal on their own (single point, recomputed from scratch each cycle); line
  charts and the cumulative growth charts do not, hence the explicit clear.
- **Deprecation path**: `STATS__INTERCHAIN_PRIMARY_ID` is retained and mapped to
  `HOME_CHAIN_ID`; the transitional `interchain_primary_id.or(home_chain_id)`
  bridge used partway through the series was removed once every statement had
  been converted, so the field is now gone from `UpdateServiceConfig`/
  `UpdateParameters`/`UpdateContext` — only the settings-layer deprecation
  mapping remains.
- **Test infrastructure** (`stats/src/tests/init_db.rs`,
  `stats/src/tests/mock_interchain.rs`): replaced the hand-rolled fake
  `crosschain_messages`/`crosschain_transfers` schema with the indexer's own
  `Migrator` (test-only git dependency), since the stub could not express
  `bridge_id`, the composite PK/FK, a nullable `dst_chain_id`, or
  `token_*_chain_id` — all required to test parity. Fixtures gained a second
  bridge reusing a numeric message id, a `dst_chain_id IS NULL` message, and a
  transfer whose token chains differ from its message route — the three cases a
  from-scratch universal-indexer deployment can supply but no existing
  production dataset can.
- **Docs**: interchain chart descriptions in `config/interchain/charts.json`
  rewritten around "within the configured chain and bridge scope"; a new
  README operator section maps the six env vars onto the equivalent indexer API
  query parameters, and is explicit that the startup log line renders the
  *configured* filter rather than the effective one — the horizon is read per
  cycle and is logged separately at `DEBUG`.

## API Changes

None. No `stats-proto` or public read-API surface change; filtering is applied
when charts are computed (update time), not when they are served.

## ENV / Config Changes

Six new variables under `STATS__INTERCHAIN_FILTER__*`, each mirroring an
indexer read-API query parameter of the same name and default:

| Variable | Type | Default |
|---|---|---|
| `HOME_CHAIN_ID` | `Option<u64>` | unset |
| `COUNTERPARTY_CHAIN_IDS` | CSV `u64` | unset |
| `SRC_CHAIN_IDS` | CSV `u64` | unset |
| `DST_CHAIN_IDS` | CSV `u64` | unset |
| `BRIDGE_IDS` | CSV `u32` | unset |
| `INCLUDE_UNINDEXED_CHAINS` | `bool` | **`false`** |

`INCLUDE_UNINDEXED_CHAINS` defaults to `false` to match the indexer API's own
default — an unconfigured stats instance and an unparameterised API request
then answer the same question. With nothing configured at all, the rendered
predicate is unchanged (no filter applied), so an existing unfiltered
deployment's behaviour is preserved except for this one default.

`STATS__INTERCHAIN_PRIMARY_ID` is retained, deprecated, and now means exactly
`STATS__INTERCHAIN_FILTER__HOME_CHAIN_ID`. Setting both to the same value logs a
deprecation warning; setting them to **different** values is a hard startup
error, not a silent precedence rule.

All six variables appear in the generated env-var table (`just check-envs`
clean).

## Database / Migration Impact

None for schema — no new migration. The filter fingerprint reuses the existing
nullable `chart_data.min_blockscout_block` column, which never carried meaning
in interchain mode.

It is not a no-op operationally, though: changing the filter configuration
(including the `INCLUDE_UNINDEXED_CHAINS` default landing as `false`) changes
the fingerprint, which makes the next update cycle clear and fully recompute
every affected interchain chart. Budget for that recompute on any deployment
whose filter configuration changes, including the first deployment onto this
change if it inherits a running database.

## Testing

- `just test-with-db` at branch head: **485 passed, 0 failed, 0 ignored** (full
  suite, disposable Postgres), verified directly by the reviewer with the exit
  status checked (not inferred from truncated/piped output).
- `just check` (`cargo check` + `clippy --all --all-targets --all-features -D
  warnings`): clean.
- `just check-envs`: clean — all six new variables present in the generated
  table.
- Per-commit progression, each verified against its own baseline before the
  next commit built on it: 428 → 453 → 483 → 485 passed, with every delta
  enumerated and no pre-existing test changing status. The review-pass commit
  holds at 485 (it renames one test and re-points its assertion; it adds none).
- A coverage test asserts each of the 13 registered statements applies the
  shared predicate exactly once, with the expected SQL fragment derived from the
  `interchain-indexer-filters` dependency at test time (not hand-copied), plus a
  `charts.json` cross-check so a newly enabled chart cannot go uncovered.
- The clear-on-mismatch regression test was shown load-bearing by mutation: with
  the clear disabled, it fails with exactly the predicted corruption (a line
  chart reading 15 days instead of 5, a cumulative-growth prefix reading 13
  instead of 5), while the counter test still passes — confirming counters
  self-heal independently of the clear.
- Expected fixture values for the statement-conversion commit were re-derived
  from an independent model of the fixtures and predicates, not patched to match
  the code's output.

## Verification

Beyond the test suite (485 passed, 0 failed), the branch was run end to end
against a dev `interchain-indexer` on one frozen database — 92,976 messages /
72,733 transfers, Omnibridge on chains 1+100 and Avalanche ICTT on
43114 + 8021 (Numine) + 68414 (Henesys), the last two added through env vars
exactly as production will add them, plus seven chains that appear in messages
but that no bridge indexes.

- **The horizon sets agree.** The indexer's config-derived `indexed_chain_ids`
  (`GET /api/v1/interchain/bridges`) is identical to what stats resolves from
  `bridges LEFT JOIN bridge_contracts`, env-added chains included. This is the
  one place stats reads a different source than the indexer, so it is the check
  that mattered most.
- **The two services return the same numbers.** For six filter shapes —
  unfiltered, focal chain, focal+counterparty, the within-set conjunction,
  directional+bridge, and horizon off — the indexer's own list endpoints and
  stats' counters agreed exactly (92,152 / 5,153 / 21,666 / 70,486 / 3,405 /
  5,154 messages; 5,151 and 105 transfers), as did ~22,500 point-level
  comparisons of every counter, daily series, cumulative series and
  week/month/year rollup against hand-written SQL.
- **The horizon is load-bearing, not a no-op:** it excludes 824 of 92,976
  messages here. `INCLUDE_UNINDEXED_CHAINS=true` admits them and changes the
  fingerprint, so the two configurations cannot share stored points.
- **Fingerprint behaviour is as designed.** Changing the filter cleared exactly
  the 39 stored charts and recomputed them, leaving a single fingerprint in
  `chart_data`; restarting with an unchanged config cleared nothing; a blank
  `…_CHAIN_IDS=` and an absent one produced the *same* fingerprint, so a
  cosmetic config edit does not trigger a recompute; and the deprecated
  `STATS__INTERCHAIN_PRIMARY_ID` produced a byte-identical fingerprint to
  `HOME_CHAIN_ID`.
- **The first full computation takes 1.9 s** for all 131 charts at this volume,
  which retires the unattended-first-run concern here. Plans: the focal
  `(src = N OR dst = N)` term is index-driven (BitmapOr over the indexer's
  `…_src_chain_ts_idx` / `…_dst_chain_ts_idx`); the horizon disjunction is a
  heap filter with no index support, so a horizon-only configuration is a
  sequential scan (35 ms here) growing linearly with the table.
- **Two cases live data cannot reach** were checked by inserting synthetic rows:
  a message with a NULL destination and one addressed to an unindexed chain were
  both correctly excluded, by the incremental update path as well as the full
  one. Also worth knowing: no message id is reused across bridges in this
  dataset, so the composite-join fix is indistinguishable from the
  `message_id`-only join there — the fixture's deliberate collision is the only
  thing covering it.

No defects were found. The one code change this produced is unrelated to the
filter: `just run-interchain` pointed at port 5433, where nothing serves.

## Review pass

The branch was reviewed independently against the task artifacts, and `c07f3d68`
acts on the result. Four findings, all confirmed against the pinned upstream
revision, none rejected:

- **The empty-`bridges` horizon was not the exact mirror the code claimed.** The
  resolver returned "no restriction" for an empty `bridges` table, on the belief
  that the indexer models a no-bridges config as `IndexedChains::AllIndexed`. It
  does not — `server.rs` always constructs via `from_bridges`, which yields
  `PerBridge` even from an empty iterator, and its startup guard explicitly
  permits `bridges.is_empty()`. The indexer would therefore render
  `(TRUE AND dst IS NOT NULL)` where stats rendered no predicate. Fixed in code
  rather than documented as an exception: the resolver now always returns a list
  and `Option` is gone from its signature, the unreachable arm with it. No
  behavioural change — an empty `bridges` table implies no messages by foreign
  key — but the equivalence now holds by construction instead of by that
  invariant.
- **The startup log called itself "effective" and was not.** With the default
  `INCLUDE_UNINDEXED_CHAINS=false` it could print `<unfiltered>` while the applied
  predicate was restrictive. Relabelled "configured", stated as such in
  `render_for_log`'s docs and the README, and the resolved horizon is now logged
  per cycle at `DEBUG` so the effective scope is observable at all.
- **Contract removal is a second parity divergence, opposite in sign to the
  documented one.** Both are now in the README and the research note.
- **Rollback is not self-healing, and the docs implied it was.** The old build
  does recompute on the fingerprint mismatch, but the write path only upserts, so
  a date whose sole row the new predicate admits and the old rejects keeps its new
  value. Documented with an explicit clear-before-rollback instruction.

One recommendation was deliberately not carried out: a rollback regression test.
A test cannot execute a previous binary, so the counterexample is only
expressible as a fixture property — and the property that guards the predicate
choice is already asserted, since `update_new_transfers_sent_interchain`'s
`home_chain_id = 1` expectation omits `2023-02-09` precisely because message 24's
transfer is scoped on `token_src_chain_id`, and would fail if the scoping reverted
to the message route. The remedy for rollback is the runbook.

## Follow-ups

In priority order, none blocking this PR:

1. **Horizon *change* detection is still absent.** The review pass added a
   per-cycle `DEBUG` line carrying the resolved set, so the effective scope is at
   least observable now. What was proposed during planning and still has not
   landed is an `INFO` line for "the resolved indexed-chain set *changed* since
   last cycle" — that needs cross-cycle state. Without it, an upstream horizon
   widening (see below) is undetectable by an operator who is not diffing DEBUG
   output, not just by the fingerprint.
2. **Re-measure plans at production volume.** The numbers above are from a
   ~93k-row table, where every plan is cheap and the planner's choices say
   little about a universal indexer's full history. The shape to watch is the
   horizon-only configuration: it has no index support at all.
3. **Two named parity divergences, documented, not fixed.** Stats derives its
   "indexed chain set" restriction from the indexer's `bridges` /
   `bridge_contracts` tables, which never delete rows; the indexer derives the
   same set from its in-memory configuration. Any *removal* from that
   configuration therefore shrinks the indexer's set and leaves stats' set as it
   was — and the two directions have opposite signs:
   - a **bridge contract** removed from a bridge that still exists ⇒ stats keeps
     admitting rows the API now excludes (stats counts more);
   - a **whole bridge** removed ⇒ the API admits its rows unconditionally (it is
     simply absent from the restriction, though messages with no known
     destination stay excluded) while stats keeps testing them against the stale
     set (stats counts less).

   Neither is repaired by a forced recompute. Closing them needs the indexer to
   publish its effective configured set; worth requesting, not worth blocking on.
4. **Rolling back to a pre-filter build needs a manual clear**, now documented in
   the README. The old build does see a fingerprint mismatch and recompute, but
   the write path only upserts, so a date whose sole row the new predicate admits
   and the old one rejects keeps its new value. Reachable in the transfer
   sent/received series, whose scoping columns changed. No regression test is
   offered: a test cannot execute a previous binary, and the predicate choice
   that makes the case reachable is already pinned by
   `update_new_transfers_sent_interchain`.
5. **Stats-server has no interchain-mode integration test** (pre-existing gap,
   unchanged by this PR — all interchain coverage remains chart-level).


EvgenKor:
@coderabbitai review

EvgenKor:
@coderabbitai review with ignoring `.memory-bank` folder

stats/stats/src/data_source/kinds/local_db/mod.rs
```diff
@@ -196,6 +197,38 @@ where
...
+            && recorded != Some(min_indexer_block)
+        {
+            tracing::warn!(
+                chart =% ChartProps::key(),
+                recorded =? recorded,
+                observed = min_indexer_block,
+                "interchain filter fingerprint changed; clearing stored chart data \
+                 before recomputing"
+            );
+            clear_all_chart_data(cx.stats_db, chart_id)
```
EvgenKor: **correctness** — the clear is not atomic with the recompute, so an interrupted rebuild leaves the chart empty *and* marked fresh.

`clear_all_chart_data` DELETEs the whole series here, and `Update::update_metadata` (which advances `charts.last_updated_at`) only runs after `update_values` succeeds. If `update_values` fails partway — indexer hiccup, `IntervalTooLarge`, pod eviction during what the PR describes as a full backfill — the chart is left with zero rows while `last_updated_at` still holds its pre-clear value. The read API then serves an empty series (and `NoCounterData` for the counters) with nothing reporting the chart as stale, until the group's *next* scheduled cycle completes a full rebuild.

The exposure is highest at the moment this ships: the `include_unindexed_chains=false` default changes every fingerprint, so all 39 interchain charts clear in one cycle and each recompute window is as long as it will ever be.

Wrapping the delete and the writes in one transaction (or deleting only after the recompute has produced rows) would remove the window.

stats/stats/src/charts/db_interaction/read/interchain.rs
```diff
@@ -1,37 +1,325 @@
...
+    let mut horizon: BTreeMap<i32, BTreeSet<i64>> = BTreeMap::new();
+    for BridgeChainPair {
+        bridge_id,
+        chain_id,
+    } in rows
+    {
+        // the bridge key is inserted whether or not the LEFT JOIN produced a
+        // contract row — that is the whole point of the outer join
+        let chains = horizon.entry(bridge_id).or_default();
+        if let Some(chain_id) = chain_id {
```
EvgenKor, line 184: **correctness** — `(bridge, vec![])` is the *most restrictive* entry the horizon can produce, and this resolver cannot tell it apart from a half-written upstream config.

A bridge present in `bridges` with no `bridge_contracts` rows folds to `(b, vec![])`, which the shared predicate renders as `1 = 2` — every one of that bridge's rows is excluded. The doc comment treats that as the deliberate "present but observes nothing" case, and it is. But the indexer writes those two tables in **two non-atomic calls**: `upsert_bridges` (its own transaction) and `upsert_bridge_contracts` (a separate `insert_many`), both in `interchain-indexer-logic/src/database.rs`. A stats cycle that reads between them sees `(b, vec![])` for a bridge whose contracts simply are not in yet.

On a database that already holds that bridge's messages that is not transient: days with rows from several bridges get their lowered values *written* (the upsert overwrites), days that become empty keep their old values, and since the fingerprint deliberately excludes the horizon the next cycle sees no mismatch and only updates incrementally from `last_accurate_point`. The earlier days keep the under-counted values permanently, with no log line to say so.

The `bridges`-is-empty case is safe by the FK argument the doc already makes; this one is not covered by it. Worth either treating a contract-less bridge as "do not narrow, warn instead", or logging at `INFO` when the resolved set shrinks relative to the previous cycle (follow-up 1 in the description would catch it if it were `INFO` and compared).

stats/stats/src/tests/mock_interchain.rs
```diff
@@ -2,158 +2,492 @@
...
+/// The chains [`MOCK_BRIDGE_ID`] has contracts on — i.e. the chains it could
+/// have observed. Deliberately a strict subset of [`MOCK_CHAIN_IDS`].
+pub const MOCK_BRIDGE_CONTRACT_CHAIN_IDS: [i64; 3] = [1, 2, 3];
+
+/// The observability horizon the fixture's `bridges` / `bridge_contracts` rows
+/// resolve to via `resolve_only_indexed_by_bridge`.
+///
+/// [`MOCK_SECOND_BRIDGE_ID`] is listed with an **empty** chain set, which is the
+/// case that excludes all of its own rows.
+pub fn mock_interchain_horizon() -> Vec<(i32, Vec<i64>)> {
```
EvgenKor, line 95: **test-coverage** — this constant is documented as what the fixture's tables resolve to, and nothing checks that.

Around ten `*_horizon` chart tests pass `mock_interchain_horizon()` in explicitly, so the entire horizon dimension is asserted against a hand-written literal rather than against `resolve_only_indexed_by_bridge`, which is what the service actually runs. The three `resolve_horizon_*` tests in `read/interchain.rs` seed their own `bridges`/`bridge_contracts` rows and never touch this fixture, so they do not close the gap.

Edit `fill_mock_interchain_data`'s contract rows — give `MOCK_SECOND_BRIDGE_ID` a contract, or drop chain 3 from `MOCK_BRIDGE_CONTRACT_CHAIN_IDS` — and every horizon test keeps passing while asserting values for a horizon the fixture no longer produces.

One DB-backed test would pin it:

```rust
#[tokio::test]
#[ignore = "needs database to run"]
async fn fixture_tables_resolve_to_mock_interchain_horizon() {
    let indexer = init_db_interchain("fixture_resolves_to_mock_horizon").await;
    fill_mock_interchain_data(&indexer, max_normalization_date()).await;
    assert_eq!(
        resolve_only_indexed_by_bridge(&indexer, None).await.unwrap(),
        mock_interchain_horizon()
    );
}
```

That also becomes the only place the production default path (`include_unindexed_chains = false`, horizon resolved from the DB) is exercised at all — every chart test today either disables the horizon or injects it.

stats/stats/src/charts/interchain_filter_coverage.rs
```diff
@@ -0,0 +1,421 @@
...
+            .unwrap_or_else(|| panic!("charts.json has no `{section}` object"));
+        configured.extend(charts.keys().cloned());
+    }
+    assert!(
+        !configured.is_empty(),
+        "charts.json parsed to no charts at all"
+    );
+
+    for config_key in configured {
+        let chart_name = snake_to_camel(&config_key);
```
EvgenKor: **correctness (coverage guard)** — layer 2 keys on the config *key*, but what a config entry serves is decided by its `implementation` remap.

That is gotcha 3 in this repo's own `CLAUDE.md`: "Chart enable/disable is keyed by what a config entry *serves* (its `implementation` remap), not its config key name." The real mapping lives in `stats-server/src/config/read/charts.rs:20`, which camel-cases *both* the key and the `implementation` value.

So an entry like

```json
"new_messages_interchain": { "implementation": "someUnregisteredInterchainChart", ... }
```

passes this guard — `snake_to_camel("new_messages_interchain")` is in `registry()` — while the chart actually enabled is unregistered and therefore unfiltered. That is exactly the failure mode the module docs promise is impossible ("a *new* interchain chart cannot reach production without passing through here").

Secondary point: `snake_to_camel` is a hand-rolled duplicate of `convert_case`'s `from_case(Snake).to_case(Camel)`, which the workspace already uses for this mapping and which handles cases this version does not (keys with digits, e.g. `new_txns_24h`). Adding `convert_case` to `stats`'s dev-dependencies and reading the `implementation` field when present would make the guard mean what it claims.

stats/stats/src/charts/db_interaction/read/interchain.rs
```diff
@@ -1,37 +1,325 @@
...
+            )
+            .into_model::<MinTimestamp>()
+            .one(interchain)
+            .await?
+            .and_then(|row| row.min_timestamp))
+    }
+
+    let messages_min = min_init_timestamp(filter.messages_query(), interchain).await?;
+    // the joined query, because `init_timestamp` lives on the message
+    let transfers_min = min_init_timestamp(filter.transfers_joined_query(), interchain).await?;
```
EvgenKor: **efficiency** — this floor is now two aggregate queries and it is recomputed per chart, with no memoisation, even though `UpdateContext::cache` already exists for exactly this.

`get_min_date` is called from `batching/mod.rs:74` once per chart doing a full update, plus once per unbounded remote pull via `QueryFullIndexerTimestampRange`. The value is identical for every chart in a cycle (it depends only on `cx.interchain_filter`), yet a full rebuild of the 39 interchain charts runs this pair ~40+ times.

The second query is the expensive one: an inner join across the whole `crosschain_transfers` table. And per the PR's own plan analysis, the horizon disjunction has no index support, so under a horizon-only configuration each call is a sequential scan of `crosschain_messages` plus a scan-and-join of `crosschain_transfers`. At universal-indexer volume that is tens of full scans per rebuild — and the rebuild is precisely what a filter change triggers.

`cx.cache` is keyed by statement string, so wrapping both queries in it would collapse this to two scans per cycle; resolving the floor once in `update_service` next to the horizon would do the same more explicitly.

stats/stats/src/data_source/kinds/local_db/mod.rs
```diff
@@ -196,6 +197,38 @@ where
...
+        // Deliberately NOT gated on `!cx.force_full`: if the fingerprint changed
+        // the delete is wanted either way, and `force_full` skips
+        // `last_accurate_point`'s read but not the staleness problem. Deliberately
+        // NOT extended to other modes: there `min_blockscout_block` means an actual
+        // block number and the existing reindex semantics depend on it.
+        if cx.mode == Mode::Interchain
+            // no stored rows ⇒ nothing to clear. A NULL recorded value counts as a
+            // mismatch: interchain never writes one, so this should be
+            // unreachable, and treating it as a mismatch is the safe direction.
+            && let Some(recorded) = recorded_min_indexer_block(cx.stats_db, chart_id).await?
```
EvgenKor: **efficiency** — this reads the row that `last_accurate_point` is about to read again, five lines below.

`recorded_min_indexer_block` was extracted *out of* `last_accurate_point`, and `last_accurate_point` still calls it. So in interchain mode every chart runs `SELECT min_blockscout_block FROM chart_data WHERE chart_id = ? ORDER BY date DESC LIMIT 1` twice per update cycle — 39 charts × every cycle, for one value.

Passing `recorded` into `last_accurate_point` (or having it return what it read) also delivers the property the helper's doc comment says it exists for — "the two must never drift" — by construction rather than by both calling the same function.

stats/stats/src/charts/db_interaction/filters/interchain.rs
```diff
@@ -0,0 +1,603 @@
...
+pub struct InterchainFilterConfig {
+    /// `only_indexed_by_bridge` is always `None` here.
+    configured: ChainBridgeFilter,
+    include_unindexed_chains: bool,
+    fingerprint: i64,
+}
+
+impl InterchainFilterConfig {
+    pub fn new(configured: ChainBridgeFilter, include_unindexed_chains: bool) -> Self {
+        debug_assert!(
```
EvgenKor: **correctness** — this invariant is compiled out in release, and both of its consequences fail silently there.

If a future caller (or a refactor of `build_interchain_filter_config`) passes a `configured` that already carries `only_indexed_by_bridge`, a debug/test build trips the assertion but the release binary does not. Two things then go wrong with no signal:

1. `with_horizon` overwrites the field, so the caller's horizon is silently dropped and the restriction never applies;
2. `filter_fingerprint(&configured, …)` hashes the horizon's *contents*, so every upstream bridge addition starts wiping and rebuilding all 39 interchain charts — the exact outcome `filter_fingerprint`'s doc comment says is being avoided.

Since the field is unused in this type, making it unreachable is cheap: take the five configured dimensions as arguments and build the `ChainBridgeFilter` here, so `only_indexed_by_bridge: None` is a fact rather than a precondition. Failing that, `assert!` rather than `debug_assert!` — this is a startup-time construction, so the cost is nil.

stats/stats-server/src/interchain_filter.rs
```diff
@@ -0,0 +1,512 @@
...
+    }
+
+    // warning 8
+    if home_chain_id.is_none() {
+        let directional: Vec<String> = setup
+            .update_groups
+            .values()
+            .flat_map(|g| g.group.enabled_members_with_deps(&g.enabled_members))
+            .map(|key| key.name().to_owned())
+            .filter(|name| name.contains("Sent") || name.contains("Received"))
```
EvgenKor, line 195: **altitude** — "is this chart directional?" is being answered by a substring match on the chart id, and the warning fires on the default configuration.

Two costs:

- **Noise on a correct config.** No home chain is the documented default, so every default interchain deployment now emits a `warn!` listing eight chart ids at startup. A warning that fires on the normal case stops distinguishing "look at this" from "this is how it works" — `info!` would fit, or gate it on some other dimension being configured.
- **It silently stops covering.** A future directional chart named `newMessagesOutboundInterchain` or `transfersFromHome` is missed; a non-directional chart whose id happens to contain those words is falsely included.

The property being tested is "this chart's meaning depends on a focal chain", which is a fact about the chart. It belongs on `ChartProperties` next to `indexing_status_requirement()` (or as a const on the `InterchainFiltered` impls, which already exist for all thirteen statements), not as a name match in the settings validator.

stats/justfile
```diff
@@ -95,15 +101,36 @@ run-multichain:
...
+#
+# Serve interchain charts from a local interchain-indexer DB
 run-interchain:
     STATS__MODE=interchain \
     STATS__CHARTS_CONFIG=config/interchain/charts.json \
     STATS__LAYOUT_CONFIG=config/interchain/layout.json \
     STATS__UPDATE_GROUPS_CONFIG=config/interchain/update_groups.json \
     STATS__DB_URL={{DATABASE_URL}} \
-    STATS__INDEXER_DB_URL=postgresql://postgres:admin@localhost:5433/interchain_indexer \
+    STATS__INDEXER_DB_URL=postgresql://{{db-user}}:{{db-password}}@{{db-host}}:{{interchain-db-port}}/{{interchain-db-name}} \
```
EvgenKor, line 124: **correctness (dev recipe)** — after this change the default invocation is the one the comment above says not to use.

`interchain-db-port` defaults to `5432`, which is also `db-port`'s default, so with no overrides `STATS__DB_URL` and `STATS__INDEXER_DB_URL` both resolve to `localhost:5432`. Combined with `STATS__CREATE_DATABASE=true STATS__RUN_MIGRATIONS=true` a few lines down, plain `just run-interchain` creates and migrates the `stats` schema *inside the interchain indexer's* Postgres — and `just start-postgres` cannot be run alongside it because the port is already published.

The comment does spell out `just db-port=5434 run-interchain`, but nothing enforces it. Either default `interchain-db-port` to something that cannot collide, or make the recipe fail fast when `db-port == interchain-db-port`, so the default invocation is the working one.

stats/Cargo.toml
```diff
@@ -13,6 +13,19 @@ members = [
...
+# All three MUST carry the identical rev: cargo keys a git source by
+# `(url, rev)`, so two revs of one URL give two checkouts and two distinct
+# copies of `interchain-indexer-entity` — `ChainBridgeFilter`'s column
+# identifiers would then be a different type from the ones stats builds with,
+# and it will not compile.
+# Pinned to the squash-merge commit of #1727 on `main`, which is where
+# `interchain-indexer-filters` was introduced. The feature branch it came from
+# has been deleted, so any earlier rev of these crates is unreachable.
+interchain-indexer-entity = { git = "https://github.com/blockscout/blockscout-rs/", rev = "58d861876900a900d64d7337d9f9ba095e5f6e11", package = "interchain-indexer-entity" }
+interchain-indexer-filters = { git = "https://github.com/blockscout/blockscout-rs/", rev = "58d861876900a900d64d7337d9f9ba095e5f6e11", package = "interchain-indexer-filters" }
```
EvgenKor, line 27: **altitude** — sharing the crate makes parity *checkable*, not *guaranteed*; the rev pin is the gap.

The README states it as absolute: "stats depends on the indexer's `interchain-indexer-filters` crate and evaluates the same code, so the two cannot drift apart silently." That holds only while the deployed indexer runs this rev. The moment the indexer ships a change to `ChainBridgeFilter` — a new dimension, a different focal truth table, a changed NULL rule — stats keeps rendering the old predicate until someone bumps `58d8618`, and the dashboard and the list view disagree while both look internally correct. Which is the exact failure the sentence promises is structurally impossible.

The design choice is right; the claim needs qualifying. Cheapest real guard: have the shared crate export a predicate-shape version constant, have the indexer report it (its bridges endpoint already exists) and have stats compare it at startup. Failing that, say in the README that parity is per-rev and that bumping this pin is a release step whenever the indexer's filter changes.

stats/stats/src/data_source/kinds/local_db/mod.rs
EvgenKor: Confirmed and fixed in 13267e70 — with one correction to the framing.

I did not make the clear atomic with the rebuild, and I don't think it can be: `BatchUpdate` commits per batch by design, and wrapping a full universal-indexer backfill in a single transaction trades this problem for a worse one. What the fix removes is the *misreporting*. `clear_chart_data_and_updated_at` now clears `charts.last_updated_at` in the same transaction as the delete, so an interrupted rebuild leaves the chart in the state a newly created one has — empty and visibly never updated — instead of empty while still carrying its pre-clear freshness.

Also worth recording, since it bounds the window narrower than the comment implies: `BatchUpdate` calls `update_metadata` after **every** batch (`batching/mod.rs:112`), so for the line charts the exposure runs from the clear until the first batch commits, not across the whole backfill. Counters are single-point, as you noted. The residual behaviour — an interrupted rebuild serves an empty or partial series — is unchanged and now written down in `.memory-bank/gotchas.md`.

Interchain-only, as requested: the new helper has exactly one caller, the fingerprint branch. `ClearAllAndPassVec` and the windowed batch step keep using plain `clear_all_chart_data`, so no other mode's behaviour moves.

stats/stats/src/charts/db_interaction/read/interchain.rs
EvgenKor: Confirmed, including the part that makes it more than transient: `upsert_bridges` (`server.rs:365`) and `upsert_bridge_contracts` (`server.rs:398`) are two separate awaited calls, so a cycle landing between them really does see `(b, vec![])` for a bridge whose contracts are not in yet — and because the fingerprint deliberately excludes the horizon's contents, the points written in that cycle are never recomputed on their own.

I have **not** changed the narrowing, and I'd argue it shouldn't change: "present in `bridges`, observes nothing ⇒ exclude all its rows" is ADR-004 Decision 5 and it is what the indexer's own API does. Flipping it stats-side would break the parity the filter exists for, and would trade a rare transient under-count for a permanent silent over-count relative to the API.

What 13267e70 does is remove the "with no log line to say so" part: the resolver now emits a `warn!` naming every bridge that resolved to an empty chain set, spelling out both readings (configured to observe nothing vs. indexer mid-startup) and what to do in the second case. Detecting the *change* rather than the state is still follow-up 1 in the description; it needs cross-cycle state.

stats/stats/src/tests/mock_interchain.rs
EvgenKor: Agreed, and added in 13267e70 — essentially the test as written, as `fixture_tables_resolve_to_mock_interchain_horizon` in `read/interchain.rs`'s test module (next to the other `resolve_horizon_*` tests, so the resolver's coverage lives in one place).

Your second point is the one I hadn't accounted for: this is now the only test that exercises the production default path at all — `include_unindexed_chains = false` with the horizon read from the database instead of injected. Both facts are in the test's doc comment so the next person editing the fixture's contract rows sees why it exists.

stats/stats/src/charts/interchain_filter_coverage.rs
EvgenKor: Confirmed and fixed in 13267e70. `config/read/charts.rs` camel-cases both the key and the `implementation` value, and enable/disable is keyed on the served id, so layer 2 now resolves each entry to `implementation` when present and asserts on that; the failure message names both ("`x` (via `implementation: y`) serves chart id `Y`"). No interchain entry uses a remap today, so this was latent — but a guard whose whole claim is "a new interchain chart cannot reach production without passing through here" shouldn't have a latent hole in it.

`snake_to_camel` now delegates to `convert_case`'s `from_case(Snake).to_case(Camel)` (added to `stats`'s dev-dependencies at the same version `stats-server` already uses), so the guard converts ids exactly as the server does rather than approximately.

stats/stats/src/charts/db_interaction/read/interchain.rs
EvgenKor: Confirmed and fixed in 13267e70 via `cx.cache`, as suggested — both queries are memoised, keyed on their own built `Statement`, which encodes the whole predicate including the horizon resolved for this cycle. So the key is complete for the cache's lifetime even though the filter *fingerprint* deliberately is not.

One honest qualification on the size of the win: `UpdateCache` is constructed per **group** update and dropped after it, not per cycle, so a full rebuild now pays one pair of queries per group rather than one per chart — roughly 13 pairs instead of 40+, not 2. Getting to 2 needs the floor resolved in `update_service` next to the horizon and threaded through `UpdateParameters`, which is the same shape as the interchain filter itself and is a fair follow-up; I stopped at the local fix because it removes the per-resolution repetition, which is the bulk of it.

stats/stats/src/data_source/kinds/local_db/mod.rs
EvgenKor: Confirmed and fixed in 13267e70. `update_itself_inner` reads it once and passes it to `last_accurate_point`, which now takes `recorded: Option<Option<i64>>` instead of querying again — so the "the two must never drift" property the helper's doc comment claims now holds by construction.

Two consequences worth flagging for review: `chart_id` became unused inside `last_accurate_point` and is gone from its signature (the ten test call sites now pass `recorded_min_indexer_block(&db, id).await.unwrap()` explicitly, so they exercise the same value they did before); and the read is skipped entirely when `force_full` outside interchain mode, which is the one combination where nothing looks at it.

stats/stats/src/charts/db_interaction/filters/interchain.rs
EvgenKor: Confirmed and fixed in 13267e70 — promoted to `assert!`, with the two silent-in-release consequences you listed written out at the call site.

I did not take the "five configured dimensions as arguments" route. It does make the invariant unrepresentable, but it also means this constructor grows a five-argument positional signature over fields that already travel together as one struct everywhere else, and the guard is at startup where an `assert!` costs nothing. If `ChainBridgeFilter` ever gains a second per-cycle field the balance changes and the structural version becomes the right call.

stats/stats-server/src/interchain_filter.rs
EvgenKor: Splitting this one, and consciously not changing the code.

The substring match is a real fragility and your framing is right — it's the same class of defect the coverage guard exists to prevent. But moving the fact onto `ChartProperties` or the `InterchainFiltered` impls means exporting a chart-key → directional mapping from `stats` to `stats-server` for the sake of a startup log line, and that's a poor trade at this size. Recorded instead as a gotcha in `.memory-bank/gotchas.md` (13267e70), including the instruction not to paper over it with a special case: a future directional chart whose id carries neither word should either be renamed or trigger the refactor.

On the level: keeping `warn!`. It does fire on the default, but what it describes is a genuinely ambiguous configuration — those charts mean "source-side observed" rather than a direction relative to any chain — and `info!` would bury exactly the thing an operator should read once. The gotchas entry says so explicitly, so the next reader doesn't file it as a defect.

stats/justfile
EvgenKor: Confirmed — my own commit, and you're right that a comment is not a guard. Fixed in 13267e70 by making the stats DB's default port 5434 instead of adding a check: now `just start-postgres` and every `run-*` recipe work alongside the sibling indexer with no overrides at all, and there is no collision left to fail fast on. `interchain-db-port` stays 5432, which is where that indexer's own `just start-postgres` publishes.

`docker-compose.dev.yml` is unaffected — it uses container-internal 5432 over the compose network and never reads `db-port`.

stats/Cargo.toml
EvgenKor: Agreed on the claim being too strong. Fixed at the documentation level in 13267e70: the README now says the predicate is shared so nobody maintains two copies of the truth table, then states plainly that the guarantee is **per revision** — if the deployed indexer runs a newer rev with a changed filter the two disagree until the pin is bumped, each looking internally correct — and names bumping the pin a release step whenever the indexer's filter changes.

The version-constant handshake is the better answer and I've left it out of this PR deliberately: it needs the shared crate to export the constant and the indexer to report it, so it's a request to that side rather than something stats can do alone. Worth opening separately.

stats/stats/src/charts/db_interaction/read/interchain.rs
EvgenKor: Accepted, and resolving on that basis — recording what remains open so the resolution isn't read as the risk being gone.

I agree the narrowing should stay: ADR-004 Decision 5 plus API parity outrank a rare transient, and the inverse (dropping contract-less bridges) would swap a transient under-count for a permanent silent over-count against the API. The `warn!` closes the part I actually objected to — that nothing said anything.

Still reachable after this commit: a cycle landing between the two upserts writes under-counted points that no later cycle recomputes, because the fingerprint excludes the horizon's contents. The `warn!` makes it diagnosable, not impossible, and the remedy is manual (force a full update once contracts are in). Follow-up 1 — the cross-cycle *change* detection — is what would make it self-announcing; worth keeping visible on that list rather than letting it settle.

stats/stats-server/src/interchain_filter.rs
EvgenKor: Agreed on both halves, resolving.

The export-a-mapping-for-one-log-line trade is the right call at this size, and `warn!` is defensible given the configuration really is ambiguous. What makes this resolvable rather than merely deferred is the gotchas entry — specifically the instruction *not* to add a special case to the substring list, which is the failure mode I was actually worried about: someone hitting the gap and widening the match instead of fixing the design. With that written down, the next directional chart forces the decision rather than sliding past it.

PR #1747, opened 2026-09-09
feat(interchain-indexer)!: deprecate the /stats/common and /stats/daily counters (follow-up to #1746)

EvgenKor:
## Why this arrives after #1746

#1746 was merged with a title that promises this change — *"adaptive narrowing of
failed-range retries, **plus deprecation of the /stats counter endpoints**"* — but
its content does not contain the deprecation half.

The title came from a PR description that covered both changes. The commit that
actually implemented the deprecation was never pushed before the branch was
squash-merged and deleted, so it existed only locally. The result is that `main`
advertises the deprecation in its history while still serving the live counter
queries: `origin/main` has no `total_messages: 0`, no `DEPRECATED` marker in
`stats.proto`, and no `tests/deprecated_stats_counters.rs`.

This PR lands the missing half on top of that merge. It revisits nothing from
#1746 — outside `graphify-out/`, main's tree is byte-identical to the retry
branch's final commit, so this is purely additive.

## Summary

- `GET /api/v1/stats/common` and `GET /api/v1/stats/daily` are deprecated. Both
  ran an unbounded `COUNT(*)` over `crosschain_messages` plus a joined `COUNT(*)`
  over `crosschain_transfers` on **every** request, uncached — the heaviest
  queries this service served, growing with the canonical tables.
- The same numbers are now precomputed by the stats service in interchain mode,
  so this service no longer needs to compute them at request time.
- Removing the endpoints is premature, so they keep responding with their full
  response shape and the count fields hard-zeroed. No query is issued.
- Request parsing and filter validation are untouched, so a malformed query
  still fails with `400` rather than silently answering zeros.

## Implementation Plan

- `stats.proto`: `option deprecated = true` on `GetCommonStatistics` and
  `GetDailyStatistics`, plus `DEPRECATED:` comments on both RPCs, their response
  messages and every count field. The comments reach the regenerated swagger as
  each operation's `summary`/`description`. Note that this `protoc-gen-openapiv2`
  build does **not** emit an OpenAPI `deprecated: true` flag, so the text is what
  marks them — the swagger in this PR is the output of the proto crate's
  `build.rs`, verified idempotent.
- Both handlers build their response from the request alone: `timestamp` echoed
  back, `date` = the UTC day of that timestamp, counts `0`. The chain/bridge
  filters are still built and the result dropped, which is what preserves the
  `400` behavior.
- `InterchainDatabase::get_total_counters` / `get_daily_counters` are kept but
  are now unreachable from the API — they had no other production callers. They
  stay only so the endpoints can be removed in one step, and the existing parity
  tests still exercise them.
- Every touched site carries a `TODO(next API iteration)` naming what to delete:
  the handlers, the RPCs and messages, the rules in `api_config_http.yaml`, the
  two DB methods with their tests, and the new test file itself.
- Docs corrected where they described these endpoints as live request-time DB
  reads: `docker/README.md` and `.memory-bank/research/stats-subsystem.md`.

## API Changes

**Deprecated, behavior changed — not removed.** Breaking for anyone still
reading those two numbers from this service, hence the `!` in the commit subject.

- `GET /api/v1/stats/common` — still `200`, same response shape;
  `total_messages` and `total_transfers` are now always `0`.
- `GET /api/v1/stats/daily` — still `200`, same response shape; `daily_messages`
  and `daily_transfers` are now always `0`. `date` is still the UTC day of the
  request's `timestamp`.
- Request schemas, query parameters and `400` validation are unchanged. The gRPC
  methods are marked deprecated in the descriptor, and the swagger documents both
  operations as DEPRECATED.
- Clients needing real counters move to the stats service in interchain mode.

## ENV / Config Changes

None.

## Database / Migration Impact

None. No migration, no entity regeneration, no backfill, no schema change. The
only persistence-side effect is negative: two unbounded `COUNT(*)` queries per
request disappear from the read path, which is the point of the change.

## Testing

All verification below was run on this branch with the **exact command CI uses**
for its DB tests step, `cargo test --locked --workspace -- --nocapture --ignored
--skip avalanche_e2e`:

- Three consecutive full runs — no failures, no `AddrInUse`.
- `cargo clippy --all --all-targets --all-features -- -D warnings` — clean (only
  the pre-existing `proc-macro-error2` future-incompat dependency warning).
- New `tests/deprecated_stats_counters.rs` — passes. It seeds one message on
  bridge 1 between chains `{1, 100}` with `init_timestamp` strictly inside the
  requested day, so the removed queries *would* have answered `1`, then asserts
  both endpoints answer `0`, that `timestamp` / `date` are still returned, and
  that a malformed `bridge_ids` still yields `400`.
- **Two negative controls**, run and reverted. Restoring the queries in both
  handlers fails the test with `total_messages: "1"`; restoring only the daily
  handler fails it with `daily_messages: "1"`. Both halves of the assertions are
  therefore reachable and non-vacuous — they pin the deprecation, not a filter
  that happens to exclude the fixture.

### On the flake in #1746's own CI

Run 34323504182 failed its DB tests step on an unrelated test,
`get_indexing_progress_pair_with_no_checkpoint_reports_zero_and_absent_updated_at`.
That is **not** caused by this change — the run was on `f5f44463`, which does not
contain these tests at all — but it is worth recording, because the same trap
shaped how the new test is written.

`blockscout-service-launcher`'s `test_server::get_free_port()` binds `127.0.0.1:0`,
reads the port and then **drops the listener**, releasing it before `run()` binds
it for real. Two tests of one binary racing there get the same port; the loser
panics `AddrInUse` inside a spawned task that nobody awaits, so the failure is
swallowed, and `init_server`'s health check then passes against the *winner's*
server. The victim test proceeds to query someone else's database and dies as an
unexplained `500 {"code":13,"message":"internal server error"}` — exactly the two
panics in that log, in that order.

The new test therefore does all three of its read-only checks against **one**
server and **one** database instead of booting three. That is prophylaxis, not a
fix: the race is in the shared harness and still applies to the six servers
`indexing_progress.rs` boots, so the failed job should simply be re-run. Worth
fixing upstream — a swallowed bind error whose health check passes against a
foreign server is a bad failure mode for every consumer of that harness.

## Follow-ups

- Delete the endpoints once clients have moved: the RPCs and their messages, the
  `api_config_http.yaml` rules, the handlers, the two DB methods with their
  parity tests, and `tests/deprecated_stats_counters.rs`. Each site points at the
  others, so it is one mechanical pass.
- `main`'s `f5f44463` commit message will remain inaccurate about what it
  contained; this PR fixes the substance, not that history.
- Report the `get_free_port` TOCTOU and the swallowed bind panic upstream in
  `blockscout-service-launcher`.

---

## Summary by CodeRabbit

- **API Changes**
  - Deprecated the common and daily statistics endpoints.
  - These endpoints now return zero counters while preserving request validation and timestamp/date fields.
  - Statistics have moved to a separate service; the deprecated endpoints are planned for removal.

- **Documentation**
  - Updated API documentation and service guidance to describe the deprecation and zero-value responses.

- **Tests**
  - Added coverage confirming zero counters, preserved response fields, and validation of malformed filters.


EvgenKor:
Thanks — one of the two findings was right and is fixed in 4fc7b977. The other cannot be applied on this repo's toolchain; details below, since the reasoning matters more than the verdict.

### ✅ Applied: stale Open Question in `stats-subsystem.md`

Correct, and it was worse than reported — the same assumption appeared twice:

- **Open Questions** asked whether the endpoints should stay request-time queries *or* be reprojected. The answer is neither: counts are hard-zeroed, no query runs, and the stats service in interchain mode owns the real numbers. The entry now records that and names exactly what to delete (endpoints, RPCs, messages, `get_total_counters` / `get_daily_counters`).
- The note's own **"update this note when"** list carried the same replacement as a *future* event. It now triggers on the eventual removal instead.

(For the record, the inline anchor pointed at lines 200-201; the actual text was at 1010-1013.)

### ❌ Rejected: `enable_rpc_deprecation=true` — not available in CI's generator

The option is real, but not in the version this monorepo pins. `.github/actions/deps/action.yml` installs **protoc-gen-openapiv2 v2.18.1**:

```
wget https://github.com/grpc-ecosystem/grpc-gateway/releases/download/v2.18.1/protoc-gen-openapiv2-v2.18.1-linux-x86_64
```

`enable_rpc_deprecation` is not among the flags registered in `protoc-gen-openapiv2/main.go` at tag `v2.18.1` — it was added later (my local v2.26.1 does have it, which is presumably where the suggestion comes from; the comment cites v2.27.1, which is neither). Go's `flag` package treats an unknown option as a hard error, so adding it to `build.rs` would fail the **Build** step for every CI run, not just this PR.

Bumping the pinned generator is a shared-action change affecting every service in the monorepo, so it does not belong in a deprecation PR. Happy to open a separate one if you want the OpenAPI `deprecated: true` flag — the proto side already carries `option deprecated = true`, so it would start emitting with no further proto changes.

I have corrected the PR description, which previously claimed this build simply *cannot* emit the flag. It can; this repo's pinned version cannot.

### ❌ Rejected: `[deprecated = true]` on the four response fields — breaks the build

Tried it. `prost` emits `#[deprecated]`, and `-D warnings` (which CI's clippy uses) turns every use into an error — including inside the generated crate itself, which we do not author:

```
error: use of deprecated field `...GetCommonStatisticsResponse::total_messages`
     = note: `-D deprecated` implied by `-D warnings`
error: could not compile `interchain-indexer-proto` (lib) due to 4 previous errors
```

`.type_attribute("GetCommonStatisticsResponse", "#[allow(deprecated)]")` does **not** rescue it: the errors come from the serde impls that `prost_wkt_build::add_serde` generates into a separate file, so suppressing them needs a blanket allow over generated code plus another in the handlers that set `total_messages: 0`. That would silence exactly the signal being added, at which point the annotation buys nothing the existing `DEPRECATED` comments on both messages and all four fields do not already convey.

Both experiments were reverted; the working tree is unchanged apart from the doc fix.

### Verification

`cargo test --locked --workspace -- --nocapture --ignored --skip avalanche_e2e` (the exact DB-tests command CI runs) and `cargo clippy --all --all-targets --all-features -- -D warnings` — both clean after the doc change.

